Issue #58, opened 2026-05-28
SnapKV: CUDA hybrid GDN port (#51 follow-up)


pekkah:
Follow-up to #51 / PR #57.

The CPU SnapKV foundation is in master once #57 merges. The CUDA hybrid GDN path (`CudaHybridGdnForwardPass` — used by qwen35moe and qwen35 27B-MTP) is the primary 12 GB VRAM target for SnapKV but isn't covered by v1.

## Scope

1. **Last-W per-layer query capture during prefill.** `CudaHybridGdnForwardPass.Prefill` is a sequential token-by-token loop (`Forward(tokens[i], …)`), so there's no batched Q buffer to harvest. Add a GPU scratch ring `[numAttnLayers × W × qDim]` that captures `_gpuQ` per layer per token for the last `W` tokens. Reuse the trailing slots across the loop — only the last `W` need to survive into the eviction pass.

2. **Post-prefill scoring.** Run per-layer scoring against the GPU K cache (`_gpuKCache[layer]`). Two options:
   - Pure GPU: dispatch `Attention`-style kernel that emits softmaxed weights summed across heads/queries into a `[promptLen]` device float buffer.
   - GPU+CPU split: download captured Q's (small, ~6 MiB total for Qwen3.6-27B-MTP at W=64) and use existing `SimdKernels.DotF32` on a host-side K download. Simpler but pays the KV download cost; only worth it if the GPU scoring kernel is non-trivial to fuse.

3. **GPU-side compaction.** New NVRTC kernel `llm_kv_compact` that takes a `[K]` keep-position list and copies survivors into slots `[0..K)` within the flat ring. Two passes (K then V) or fused.

4. **Wire up bookkeeping.** `CudaHybridGdnForwardPass` keeps its own `PagedKvCache` purely for length tracking — switch it to call `.Compact(keep)` after the GPU compaction. Decode-side: `AttnLayer` already uses `kvPosition + 1` for the kernel `seqLen` — change to `_kvCache.Length + 1` so it sees the post-compaction count.

5. **Compose with #56 (bf16 KV).** The kernel needs a bf16-store variant too (i.e. `llm_kv_compact_bf16`), or generic byte-stride copy.

## Acceptance

- `SHARPI_SNAPKV_BUDGET=2048` on a 16 K prompt: GPU KV cache shrinks ~8× (matching the issue's projection), decode stays well-formed.
- No decode regression vs no-eviction (prefill-only cost).
[...]

PR #64, opened 2026-05-28
feat: SnapKV port to Vulkan (closes #59)

### LINKED ISSUE #59: SnapKV: dense CUDA + Vulkan ports (#51 follow-up)
Follow-up to #51 / PR #57.

Once the hybrid GDN port lands, the remaining backends are the **dense** CUDA forward (`CudaForwardPass` / `CudaHybridForwardPass`) and the Vulkan forward (`GpuForwardPass`).

### Dense CUDA
- Same approach as the hybrid GDN port (capture last-W Q's per layer during prefill, score per layer, compact each layer's GPU K/V ring).
- `CudaForwardPass` has a batched prefill (one GEMM per weight matrix across N tokens), so the per-layer Q capture is just a slice of the batchQ tensor — easier than the hybrid GDN sequential loop.

### Vulkan
- `GpuForwardPass` uses Vulkan compute shaders against per-layer GPU K/V tensors.
- New GLSL shader for the compaction (gather-scatter over a uniform position list) — keep it as a single dispatch that consumes the `[K]` keep list + source ring + writes to the same ring's prefix.
- Scoring: easiest to keep the existing attention shader pattern and accumulate softmax weights into a buffer — same approach as the CUDA NVRTC kernel.

Could be split into two issues if you'd rather one PR per backend; bundled here because they share the algorithm.

## Acceptance

- `SHARPI_SNAPKV_BUDGET` works on every supported model on both backends.
- No decode regression with eviction disabled.
- Existing per-backend smoke tests stay green.

pekkah:
## Summary

Plumbs SnapKV prefill-time KV eviction into `GpuForwardPass`. Together with the already-merged #58 (CUDA hybrid GDN) and #63 (dense CUDA), this **closes #59**.

Two new GLSL compute shaders in `src/SharpInference.Vulkan/Shaders.cs`:

- **`SnapKvScore`** — direct port of the CUDA `llm_snapkv_score` kernel. Same workgroup-per-head × 256-thread structure, same shared-mem ≤ 4096 fast path / scores_scratch spill, same `-1.0/0.0` neg-inf convention as the existing `Attention` shader. Vulkan core has no native float atomicAdd, so the per-position score accumulator is mutated through a **CAS loop on a uint reinterpretation** of the f32 buffer (`atomicCompSwap` over `coherent buffer ScoreAccumAtomic { uint accum_uint[]; }`). Slower than CUDA's native float atomicAdd but functionally equivalent and works on every Vulkan core device without requiring `VK_EXT_shader_atomic_float`.
- **`KvCompact`** — direct port of `llm_kv_compact`. 2D dispatch `(ceil(kvDim/256), K, 1)` matches the CUDA reference grid.

`VulkanBackend` wrappers `SnapKvScore` / `KvCompact` / `ClearRegion` mirror the `CudaBackend` API so `GpuForwardPass.ApplySnapKvEviction` reads identically to `CudaForwardPass.ApplySnapKvEviction` (PR #63).

## Bundled fixes

- **`_kvLength` was never updated in master's `GpuForwardPass.Forward`** — confirmed via `git show master`; `KvLength` always returned 0 from a clean prefill. Added `_kvLength = Math.Max(_kvLength, position + 1)` at end of `Forward` matching `CudaForwardPass.cs:625`. SnapKV's correctness tests need it.
- `EnsureSnapKvScoreBuffers`: when `_maxSeqLen ≤ 4096` the score kernel never touches scratch (shared-mem fast path), so alias to `_attnScoresScratch` (1-float placeholder) and skip the dedicated `[numHeads × maxSeqLen]` allocation — saves ~9 MiB on small-context Vulkan models. (From code review.)

## Test plan

- [x] `dotnet build -c Release` clean under `TreatWarningsAsErrors=true`.
- [x] `dotnet test -c Release` — **409/409 pass**, 3 new `GpuForwardPassSnapKvTests` actually run on this Vulkan-capable box.
- [x] Hardware sanity on RTX 4070 Ti (Vulkan backend):
  - SnapKV active (budget=64, window=32, recency=32) on 118-token prompt: coherent decode, **47.9 t/s**, identifies the question correctly.
  - SnapKV explicitly disabled (`SHARPI_SNAPKV_BUDGET=0`): **50.0 t/s**, identical output to pre-PR master — no regression from the new capture / barrier in `Forward`.

## Known follow-ups (out of scope, not blocking)

- **Submission batching in score loop**: currently 2304 submissions for Qwen3-8B at W=64 (~23-115 ms tail). The CAS atomicAdd serialises ordering inside a single command buffer too, so batching is safe — defer to a perf follow-up.
- **`CudaHybridForwardPass`**: split CPU+GPU dense; needs new `KvCache.Compact` infra. Will queue as a follow-up issue.

### DIFF
--- a/src/SharpInference.Engine/GpuForwardPass.cs
+++ b/src/SharpInference.Engine/GpuForwardPass.cs
@@ -452,16 +452,20 @@ public sealed unsafe class GpuForwardPass : IForwardPass
         // doesn't participate in scoring and will not be downloaded.
         _gpu.ClearRegion(_snapKvScoreAccum!, 0, N);
 
+        // Record ALL (layer, w) score dispatches into one command buffer and
+        // submit once. The CAS-based atomicAdd on _snapKvScoreAccum (binding 2
+        // of SnapKvScore) serialises writes globally, so dispatch order doesn't
+        // affect the final accumulator. WAR/WAW hazards on the reused _q buffer
+        // are handled by RecordBarrier between each (CopyRegion → SnapKvScore)
+        // pair and after each SnapKvScore (before the next iter's CopyRegion).
+        // For Qwen3-8B (36 × 64 = 2304 dispatches) this collapses ~50 ms of
+        // per-submit overhead into a single submission. (Closes #66.)
         int qDim = _numHeads * _headDim;
+        _gpu.BeginRecord();
         for (int layer = 0; layer < _hp.NumLayers; layer++)
         {
             for (int w = 0; w < W; w++)
             {
-                // Stage the captured Q into _q so the scoring kernel can read a
-                // contiguous [numHeads × headDim] vector at the same device
-                // pointer it does during Forward. One command buffer per pair
-                // so the host stays in lockstep with the score atomicAdds.
-                _gpu.BeginRecord();
                 long srcOffsetElems = ((long)layer * _snapKvQCaptureW + w) * qDim;
                 _gpu.RecordComputeCopyRegion(_q, 0,
                     _snapKvQCapture!, srcOffsetElems * sizeof(float),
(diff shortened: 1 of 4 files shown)


pekkah:
Addressed gemini-code-assist's submission-batching feedback in 13ef904 — collapsed the 2304 per-(layer,w) submissions into a single command buffer with per-pair RecordBarriers. Hardware sanity: prefill 33.5 → 35.1 t/s on Qwen3-8B Q4_K_M / 118-token prompt with budget=64. 3/3 GpuForwardPassSnapKv tests still pass. Closes #66 (the follow-up issue I'd queued before realising the bot's suggestion fit cleanly into this PR's scope).

Issue #70, opened 2026-05-28
SnapKV+TQ (#60) Phase 3: Vulkan TQ scoring + compaction


pekkah:
Sub-issue from #60 (SnapKV + TurboQuant composition). Depends on Phase 1 (CPU) and Phase 2 (CUDA) landing first.

## Scope

Mirror Phase 2's CUDA design in Vulkan GLSL.

### Part A — Scoring against TQ-compressed K

Add \`SnapKvScoreTq\` GLSL shader mirroring \`TqAttention.comp\`'s per-block dequant-dot pattern. Same CAS-based float atomicAdd as \`SnapKvScore\` (#64) for the per-position accumulator.

### Part B — Compaction

Add \`KvCompactTq\` GLSL shader for the gather-with-decompress step + a re-quantize companion shader.

### Part C — bf16 stacking

If bf16 + TQ + SnapKV is meaningful on Vulkan, add a bf16-K variant; otherwise scope-narrow to fp32 K only and document.

## Acceptance

Same as #60 acceptance, limited to Vulkan backend.

## References

- Phase 1 (CPU design — re-quantization algorithm).
- Phase 2 (CUDA implementation — kernel structure to mirror in GLSL).
- Existing Vulkan TQ shaders: \`Shaders.cs\` \`TqAttention\` / \`TqRotateQuery\` / \`TqKvAppend\` blocks.

Issue #69, opened 2026-05-28
SnapKV+TQ (#60) Phase 2: CUDA TQ scoring + compaction


pekkah:
Sub-issue from #60 (SnapKV + TurboQuant composition). Depends on Phase 1 landing first so the CPU re-quantization design is concrete.

## Scope

Port the CPU TQ-aware SnapKV path from Phase 1 to the CUDA backend.

### Part A — Scoring against the TQ-compressed K cache

Existing CUDA SnapKV kernels (\`SnapKvScore\` / \`SnapKvScoreBf16\` in \`CudaTextKernels.cs\`) read a contiguous \`[max_seq_len, kv_dim]\` K cache. The TQ path stores K in \`_gpuTqKCache\` (per-block compressed) + \`_gpuKCache\` (FP32 ring), with a separate \`TqAttention\` kernel that interleaves them at attention time.

Options:
1. Add a \`SnapKvScoreTq\` NVRTC kernel that mirrors the \`TqAttention\` pattern (per-block dequant-dot using the codebook + sign patterns), atomicAdd's the same per-position scores.
2. Materialise a temporary fp32 K cache on GPU at SnapKV time only, score against it, throw it away.

Option 1 is the right answer for VRAM efficiency on 12 GB cards (the temporary materialise costs \`maxSeqLen × kvDim × 4 B\` per layer = ~9 MiB × 36 layers = ~324 MiB transient for Qwen3-8B). Pick this.

### Part B — Compaction with re-quantization

Add \`CudaBackend.KvCompactTq(...)\` and \`KvCompactTqStaging(...)\` kernels that:
1. Dequantize each kept position's block on the fly.
2. Reseat into freshly-allocated TQ tiles, re-using the layer/head sign patterns + codebook from Phase 1.

\`CudaHybridGdnForwardPass.ApplySnapKvEviction\` and \`CudaForwardPass.ApplySnapKvEviction\` get the new dispatch path conditional on \`_tqEnabled\`.

### Part C — bf16 K cache + TQ stacking

Verify the bf16 K cache (\`_kvDType == BFloat16\`) path composes correctly with TQ + SnapKV. Likely needs \`SnapKvScoreTqBf16\` kernel variant.

## Acceptance

Same as #60 acceptance (16K prompt + \`--tq\` + SnapKV budget=2048 → 16× total KV reduction vs fp16 baseline) but limited to the CUDA backend.

## Out of scope

- Vulkan path (Phase 3).
[...]

Issue #68, opened 2026-05-28
SnapKV+TQ (#60) Phase 1: CPU TurboQuantKvCache.Compact + scoring


pekkah:
First sub-issue from #60 (SnapKV + TurboQuant composition).

## Scope

CPU-only — defer CUDA and Vulkan to follow-up phases.

### Part A — Scoring with a TQ K cache

\`ForwardPass.Prefill\` already gates SnapKV out when \`--tq\` is on (per the gate at the bottom of the SnapKV-active condition, mirroring CudaForwardPass after #63 / GpuForwardPass after #64). Phase 1 lifts that gate for the CPU path:

1. In \`SnapKvSelector.AccumulateLayer\`, when the K cache is a \`TurboQuantKvCache\`, decompress each position's K row on the fly via \`TurboQuantOps.Dequantize\` against the layer/head compressor before the dot product.
2. Add a TQ-aware overload of \`AccumulateLayer(float* batchQ, int N, TurboQuantKvCache cache, ...)\` to avoid the abstract-virtual cost. Or extend with an \`IKvKeyAccessor\` shim.

### Part B — Compaction with re-quantization

Add \`TurboQuantKvCache.Compact(int[] keep, int N)\`:

1. Split \`keep\` into \`keepTq\` (positions in \`[0, layer.TqLength)\`) and \`keepFp32\` (positions in \`[layer.TqLength, N)\`).
2. For each layer × kv_head:
   - Decompress every position in \`keepTq\` into a temporary fp32 buffer.
   - Concatenate with the FP32 positions from \`keepFp32\` (already fp32).
   - Re-quantize the survivors using the existing layer/head compressor's codebook — the codebook is per-(layer, head) Lloyd-Max trained from initial samples, and reusing it lets us avoid re-training. Layout the result as fresh FastScan tiles + staging.
3. Update \`_layerTqLengths[layer]\` to the count of TQ-region survivors and \`_totalLength = K\`.

The kept-position decompress + re-quantize round trip introduces a small quality loss (compressing through the codebook twice). Measure it via the \`benchmarks/SnapKvEval\` harness from #61 — pass criterion: needle recovery at \`budget=2048 + --tq\` should match \`budget=2048\` (no TQ) within 1 sample on the SmolLM2 + Qwen3-8B harness.

### Part C — Lift the gate

[...]

Issue #60, opened 2026-05-28
SnapKV + TurboQuant composition (#51 follow-up)


pekkah:
Follow-up to #51 / PR #57.

`TurboQuantKvCache` stores K/V in 3-/4-bit Lloyd-Max codebooks with per-block scale staging and a fixed-size FP32 attention window at the tail. Compositing SnapKV onto it gets ~16× total KV reduction (4× TQ × 4× position pruning) — the issue's headline use case for 12 GB cards at long context.

## Scope

1. **Score the TQ cache.** TQ K is stored compressed; to score it against the last-W queries we either decompress on the fly per (head, position) or hoist the existing per-block staging path. The TQ attention kernel (CPU + CUDA + Vulkan) already does on-the-fly decompress at attention time — extend that same path to emit per-position softmax weights as a side output instead of (or in addition to) the attention output.

2. **Compact the TQ cache.** TQ blocks are fixed-size groups (32 or 64 tokens per block, see `TurboQuantKvCache.cs`). Compaction needs to either:
   - Re-quantize after picking survivors (cleanest, preserves block invariants).
   - Carry a per-block bitmap of kept positions and skip-evicted-positions in attention (simpler but no actual VRAM win for unevicted blocks).

   Prefer re-quantization, since the issue's value is VRAM reduction.

3. **FastScan compatibility.** The CPU `--tq` path uses FastScan tiles (`#34`); confirm the compaction doesn't invalidate the per-block scale tables.

## Acceptance

- `SHARPI_SNAPKV_BUDGET=2048` + `--tq` on a 16 K prompt: total KV footprint drops ~16× vs fp16 baseline (Issue #51 motivation).
- Coherent decode + no decode-throughput regression vs `--tq` alone.
- CPU + CUDA + Vulkan all functional.

## Notes

- The TQ layout doc lives in `docs/SharpInference-Design.md` ("TurboQuant" section) — read before touching block boundaries.

pekkah:
Broken into a phased roadmap so this can land incrementally rather than as one huge multi-backend PR:

- **Phase 1**: #68 — CPU-only. \`TurboQuantKvCache.Compact\` with re-quantization + TQ-aware \`SnapKvSelector.AccumulateLayer\` + lift the \`!_tqEnabled\` gate. Smallest of the three; the CPU codebook + decompress path are already exposed via \`TurboQuantOps\`, so this is mostly engine plumbing.
[...]

PR #71, opened 2026-05-28
feat: SnapKV + TurboQuant composition on CPU (#68 — Phase 1 of #60)

pekkah:
## Summary

Phase 1 of GitHub issue #60 (SnapKV + TurboQuant composition). **CPU path only.** CUDA (#69) and Vulkan (#70) are tracked separately — same algorithm, different dispatch paths.

Lifts the previously-implicit gate where SnapKV silently no-ops under \`--tq\`, and plumbs eviction through the TQ cache's FastScan tile + FP32 ring layout.

## Changes

- **\`SnapKvSelector\`**: new \`AccumulateLayer(TurboQuantKvCache cache, ...)\` overload. Rotates Q once per (head, w) via \`TurboQuantOps.RotateQuery\`, scores against the TQ-compressed K via \`cache.ComputeKScores\` (the same path \`TqAttention\` uses), muxes with FP32-ring scores in position order under the same causal mask + softmax.

- **\`TurboQuantKvCache.Compact(int[] keep, int N)\`**: per-layer partition of \`keep\` into TQ-source and FP32-source survivors. TQ survivors are decompressed via per-(layer, head) Lloyd-Max codebook + sign pattern, then re-quantized through the **same** compressor into fresh FastScan tiles + staging. FP32 survivors copy directly into the new ring. Atomic pointer swap. The FP32-overflow → TQ-promotion branch is defense-in-depth (today's \`Append\` enforces FP32 count ≤ window, so it's unreachable) but kept for the GPU phases where async append staging may transiently overflow.

- **\`ForwardPass.PrefillCoreTq\`**: mirrors \`PrefillCore\`'s SnapKV plumbing — \`AccumulateLayer\` per layer after attention writes, \`Compact\` once at end. Same env-driven gate (\`N > budget && N > window && startPos == 0\`).

- **\`ForwardPass.TqAttention\`**: clamps \`seqLen = min(position + 1, tq.Length + 1)\` so post-Compact decode doesn't read past the shrunk TQ region (mirrors how the non-TQ \`Attention\` path handles \`PagedKvCache.Compact\`).

## Test plan

- [x] \`dotnet build -c Release\` clean under \`TreatWarningsAsErrors=true\`.
- [x] \`dotnet test -c Release --filter \"FullyQualifiedName~SnapKv\"\` — **17/17 pass** (existing CPU SnapKV + 4 new TQ tests):
  - \`Compact_RoundTrip_PreservesKeptValues\` — striped 256→128 keep set, asserts ≥ 0.95 K/V cosine on every survivor (codec's steady-state direction-preservation bound).
  - \`SnapKvTqPath_LongPrompt_CacheShrinksToBudget\` — Qwen3-8B, budget=256, 384-token prompt → \`TqCache.Length == 256\`.
  - \`SnapKvTqPath_DecodeStaysCoherent\` — same setup, 4 decode steps, finite logits + ≥ 2 distinct argmaxes.
  - \`SnapKvTqPath_EnvUnset_NoEviction\` — env unset → cache length stays at prompt length.
- [x] \`dotnet test -c Release --filter \"FullyQualifiedName~TurboQuant\"\` — **45/45 pass** (no regression).
- [x] Hardware sanity on RTX 4070 Ti (CPU backend):
  \`SHARPI_SNAPKV_BUDGET=64 SHARPI_SNAPKV_WINDOW=32 SHARPI_SNAPKV_RECENCY=32\` + \`--tq\` on Qwen3-8B Q4_K_M @ 118-token prompt: eviction triggers, decode produces coherent output (recognises the embedded question correctly).

## Out of scope (tracked as follow-ups)

- **#69 — Phase 2**: CUDA port. New \`SnapKvScoreTq\` / \`KvCompactTq\` NVRTC kernels.
- **#70 — Phase 3**: Vulkan port. New GLSL shaders.

### DIFF
--- a/src/SharpInference.Engine/ForwardPass.cs
+++ b/src/SharpInference.Engine/ForwardPass.cs
@@ -609,6 +609,22 @@ public sealed unsafe class ForwardPass : IForwardPass
     {
         var cache = _tqKvCache!;
         int N = tokens.Count;
+
+        // SnapKV (issue #60) gating: fresh prefill, explicit budget, prompt
+        // long enough to drop something. TQ is fine to compose with — the
+        // selector reads from the same cache the per-token TqAttention writes
+        // and the compaction promotes the oldest FP32-window survivors into
+        // the TQ region as needed.
+        bool snapKvActive = _snapKvCfg.Enabled
+                         && startPos == 0
+                         && N > _snapKvCfg.Budget
+                         && N > _snapKvCfg.Window;
+        if (snapKvActive)
+        {
+            _snapKv ??= new SnapKvSelector(_numHeads, _numKvHeads, _headDim);
+            _snapKv.Reset(N);
+        }
+
         var batchHidden = (float*)NativeMemory.AllocZeroed((nuint)((long)N * _embDim * sizeof(float)));
         var batchResidual = (float*)NativeMemory.AllocZeroed((nuint)((long)N * _embDim * sizeof(float)));
         try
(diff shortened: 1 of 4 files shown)


pekkah:
Addressed both gemini-code-assist medium findings in aa09721 — hoisted the per-layer scratch arrays in Compact (decompKeyArr/decompValArr/unpackBlockArr) out to a single allocation, and switched tqScoreScratch in the SnapKvSelector TQ overload to ArrayPool<float>.Shared.Rent / Return. 17/17 SnapKV tests still pass.

Issue #123, opened 2026-06-04
perf(cuda): batched-trunk prefill for the pure-attention layer-split hybrid (CudaHybridForwardPass / Qwen3-Coder-30B)


pekkah:
Follow-up to **#119** (PR #120). #119 extended the batched-prefill trunk to the dense and GPU-SLRU configs of **`CudaHybridGdnForwardPass`** (the all-GPU-trunk GDN-hybrid class). But **Qwen3-Coder-30B-A3B** — named in #119 — is a *pure-attention* MoE (no GDN layers) routed through a different class, **`CudaHybridForwardPass`** (the `-g N` layer-split hybrid: N attention layers + MoE on GPU, the rest on CPU). That class has **no batched-prefill path at all** — it prefills token-by-token via its own `Forward`.

So the Coder-30B CUDA row (and any other model on `CudaHybridForwardPass`) still pays per-token prefill: per-position attention launches + per-token MoE.

## Scope
- Give `CudaHybridForwardPass` a batched prompt-prefill analogous to `CudaHybridGdnForwardPass`'s: GEMM-batched attention projections (q/k/v/o, norms, RoPE) + batched KV-append + batched-query SDPA (shared-scores ≤4096 + wave >4096, reusing the #114-B/#118 kernels) over the N prompt tokens, with the MoE/FFN stage swapped per the layer-split placement (GPU layers batched; CPU-offloaded layers per-token or CPU-batched).
- The attention kernels (`llm_full_seq_attention`(_global)(_bf16), `llm_kv_append_batched`, `MatMulBatched` GEMM-N, batched norms/RoPE) already exist and are backend-shared, so this is mostly host wiring + a parity oracle.
- Keep the `_faulted` non-transactional latch + bit-exactness contract.

## Tests
- `BatchedPrefill_BitwiseMatchesSequential` for a `CudaHybridForwardPass` model (Qwen3-Coder-30B-A3B CUDA, `-g -1`), single + multi-chunk + >4096.

## Related / out of scope
- Vulkan and CPU-only GDN-hybrid backends still prefill per-token (no batched kernels) — would need their own batched implementations; track separately if wanted.

pekkah:
Closing as completed — the in-scope deliverables shipped and are green on master:

[...]

Issue #122, opened 2026-06-04
perf(cuda): batched-query SDPA + Q-capture for SnapKV-active GDN-hybrid prefill (#118 follow-up)


pekkah:
Follow-up to **#118** (PR #120). The wave-based >4096 batched-query SDPA (and the ≤4096 shared-scores `AttentionBatched`) is gated `BatchedAttnEnabled && !snapKvActive` in `AttnBlockBatched`. When SnapKV (issue #58) is active the prefill falls back to the **per-position** KV-append + `Attention` loop, because that loop is also where the trailing-window query vectors are captured into `_snapKvQCapture` (for the post-prefill eviction scoring).

So a SnapKV-active prefill — fully reachable on any backend once `SHARPI_SNAPKV_BUDGET` is set and the prompt exceeds budget+window — gets neither the batched-query SDPA nor the #118 wave path; it pays the per-position attention-launch cost the rest of #114-B/#118 removed, and that cost still grows with context.

## Goal
Let the batched / wave SDPA run with SnapKV active by capturing Q in the batched path.

## Approach (sketch)
- The batched projections already produce `_gpuBtQ` ([N × qDim]) before the SDPA. For the trailing window `[wStart, N)`, copy those rows into `_snapKvQCapture` in one batched `CopyDeviceRegion` (or a small strided-copy kernel) keyed by the attention-layer index — instead of the per-position copies in the fallback loop — then run `AttentionBatched` / `AttentionBatchedWave` as normal.
- Must stay bit-identical to the per-position path for both the attention output AND the captured Q (the SnapKV eviction oracle + `BatchedTrunkGpuFfn_SnapKvActive_BitwiseMatchesSequential_Dense27BMtp` must hold).

## Tests
- Extend the SnapKV-active parity oracle to assert batched == sequential with `BatchedAttnEnabled` on (today it passes because both arms take the per-position path; the new code must keep it bit-identical while actually using the batched SDPA).
- A >4096 SnapKV-active case (wave path + Q-capture).

Issue #121, opened 2026-06-04
perf(cuda): batch the FFN/MoE stage of the dense + GPU-SLRU GDN-hybrid prefill (#119 follow-up)


pekkah:
Follow-up to **#119** (PR #120). `PrefillBatchedTrunkGpuFfn` batches the **trunk** (GDN/attn launches collapsed via #111/#114-B) for the dense GDN-hybrid and GPU-SLRU MoE configs, but the **FFN/MoE stage still runs per token**:

```csharp
// PrefillBatchedTrunkGpuFfn
TrunkBlockBatched(layer, N, ...);          // batched
for (int i = 0; i < N; i++) {              // per token
    CopyDeviceRegion(_gpuNormBuf, ..., moeNorm, i*embDim, ...);
    if (!isMoe)  GpuDenseFfn(layer) / CpuDenseFfn(layer);
    else         GpuMoeFfn(layer);          // GPU-SLRU routed experts, per token
    AddInPlace + write stream slice
}
```

So the trunk win lands but the FFN is N separate launches per layer. Two batchings remain:

## Scope
1. **Dense FFN batched over N** — gate/up/down as single GEMM-N launches over the N post-attn-norm rows (à la the existing `MatMulN2` / `CpuDenseFfn2` two-token path, generalized to N), instead of the per-token `GpuDenseFfn`/`CpuDenseFfn` loop. The GEMM-N kernels already exist for Q4_K/Q5_K/Q6_K/F32 (`MatMulBatched`); the dense FFN gate/up/down just need to feed them. Must stay bit-identical to the per-token path.
2. **GPU-SLRU routed experts grouped-by-expert** — analogous to the CPU-MoE `BatchedRoutedExperts` but on-GPU: load each cached expert's gate/up/down once per layer and matmul against all N tokens routing to it, instead of `GpuMoeFfn`'s per-token SLRU `GetOrLoad` + matmul. Keeps the routed-expert amortization the CPU-MoE path already gets.

## Related
- `MatMulBatched` currently throws `NotSupportedException` for Q8_0 (and other non-{Q4_K,Q5_K,Q6_K,F32}) weights. A GDN-hybrid model with Q8_0 trunk/FFN projection weights would hit this; add `llm_matvec_q8_0_gemm_n` if such a model needs the batched path.

## Tests
[...]

Issue #119, opened 2026-06-03
perf(cuda,engine): extend batched-prefill trunk to dense GDN-hybrid + GPU-SLRU MoE paths (#114 follow-up)


pekkah:
Follow-up to **#110/#111/#114-B**. The batched-prefill harness (`PrefillBatchedCpuMoe` → `TrunkLayerBatched` → `GdnBlockBatched`/`AttnBlockBatched`) is gated on `_cpuMoe`:

```csharp
// CudaHybridGdnForwardPass.Prefill
if (BatchedPrefillEnabled && _cpuMoe && batchedSafe) { PrefillBatchedCpuMoe(...); }
else { for (i..N) Forward(tokens[i], startPos+i); }   // sequential per-token prefill
```

So two CUDA GDN-hybrid configurations still prefill **sequentially per token** and get neither the #111 GEMM-batched projections nor the #114-B fused GDN scan / batched-query SDPA:

1. **Dense GDN-hybrid** (`!hp.IsMoE`, e.g. Qwen3.6-27B-MTP) — `_cpuMoe = false` (constructor line ~809). Prefill is per-token `Forward` → `GpuGdnBlockAt` / `GpuAttnBlockAt` + per-layer CPU/GPU dense FFN.
2. **GPU-SLRU MoE** (`_cpuMoe = false` when SLRU capacity ≥ ~50% of experts, e.g. Qwen3-Coder-30B on CUDA) — same sequential prefill loop.

The trunk batching (#111 + #114-B) is independent of how the FFN/MoE runs — it only needs the per-token GDN/attn launches collapsed — so it should extend to both paths. The routed-MoE amortization (#110/#112/#114-A) is CPU-MoE-specific and would *not* apply to the SLRU path, but the **trunk** half would.

## Scope
- Factor the trunk-batching path (`TrunkLayerBatched` + the host residual/norm/shared-expert plumbing) so it can drive a batched prefill for the dense and SLRU configs, with the FFN/MoE stage swapped per config (dense FFN batched à la `CpuDenseFfn2` / on-GPU FFN; SLRU routed experts).
- Keep the `_faulted` non-transactional latch and the same bit-exactness contract.

## Tests
- Extend `BatchedPrefill_BitwiseMatchesSequential` coverage to a dense GDN-hybrid model (27B-MTP) and a forced GPU-SLRU MoE config (`SHARPI_CPU_MOE=0` on a model that fits SLRU), asserting bit-identical prefill logits vs the sequential loop.

[...]

PR #128, opened 2026-06-04
perf(cuda): batched FFN/MoE prefill (#121), SnapKV batched SDPA (#122), Coder-30B batched trunk (#123)

pekkah:
Three follow-ups to #119 (PR #120), each **bit-identical** to its sequential reference and validated on an RTX 4070 Ti.

## #121 — batch the FFN/MoE stage of the dense + GPU-SLRU GDN-hybrid prefill
`PrefillBatchedTrunkGpuFfn` batched the trunk but ran FFN/MoE per token. Now:
- **Dense FFN** gate/up/down as single GEMM-N launches over N (`BatchedGpuDenseFfn`).
- **GPU-SLRU routed experts grouped-by-expert** (`BatchedGpuMoeFfn`): each cached expert loaded once and matmul'd against all its tokens, then a **per-token top-k-ordered reduce** that preserves the exact `AddScaledInPlace` accumulation order of the sequential `GpuMoeFfn` (shared expert added last) — so it stays byte-identical.
- `SHARPI_BATCHED_FFN` gate; unsupported dtypes / CPU-dense-fallback layers stay on the per-token path.

## #122 — batched-query SDPA + Q-capture for SnapKV-active GDN-hybrid prefill
SnapKV used to force the per-position loop (the only place Q was captured). Now `AttnBlockBatched` captures the trailing window `[wStart,N)` from `qAll` in a single batched `CopyDeviceRegion`, then runs the `AttentionBatched` / `AttentionBatchedWave` (#118) SDPA — byte-identical capture *and* attention output vs the per-position path.

## #123 — batched-trunk prefill for `CudaHybridForwardPass` (Qwen3-Coder-30B)
The `-g N` layer-split pure-attention MoE hybrid had no batched prefill. New `PrefillBatchedTrunk` / `GpuLayerBatchedTrunk` batch the attention trunk of the GPU layers over N (norm/proj/RoPE/QK-norm/KV-append/SDPA shared+wave), run FFN/MoE per token, transfer the N hidden rows GPU→CPU, run the CPU layers per token, then output. Gated to supported configs (non-Gemma4, non-TQ, NEOX RoPE, no L2-QK-norm, no attn-bias, GPU-resident embedding, batchable dtypes) with per-token fallback; `_faulted` latch added.

## Validation
All new/changed oracles assert **bit-identical** final-token logits vs the deterministic sequential `Forward` reference:
- `BatchedFfn_BitwiseMatchesPerTokenFfn_Dense27BMtp` — `SHARPI_BATCHED_FFN=0` equivalence (#121).
- `SnapKvActive_BatchedWave_BitwiseMatchesSequential_Over4096_Carnice` — >4096 wave + SnapKV capture (#122).
- new `CudaHybridBatchedPrefillTests` — Coder-30B single / multi-chunk / >4096, each asserting `LastPrefillWasBatched` so a gated-out config can't pass vacuously (#123). Coder split: 29 GPU + 19 CPU layers.
- GPU-SLRU MoE oracle pins the batched flags on; existing GDN oracles stay green (11/11).

Reviewed per-issue (pr-review-toolkit code-reviewer) + a final pass (silent-failure-hunter / pr-test-analyzer / comment-analyzer); findings addressed (view-handle leak, anti-vacuity assertions, flag pinning, the missing `SHARPI_BATCHED_FFN=0` oracle, comment accuracy).

## Measured perf (warm-cache ~1K-ctx A/B, RTX 4070 Ti) + README
- **#123 Coder-30B CUDA hybrid: prefill 25.2 → 30.1 t/s (+19%)** vs per-token `Forward` (`SHARPI_BATCHED_PREFILL=0`). README row updated to **30.1**; decode unchanged (FFN/MoE + CPU layers stay per token).
- **#121 dense FFN (27B-MTP Q4/Q5 CUDA): prefill-neutral at ~1K** (8.5↔8.5, 5.6↔5.5) — 44/64 (Q4) / 51/64 (Q5) FFN layers are CPU-mmap and dominate prefill; the launch-collapse pays off when more FFN is GPU-resident. README numbers unchanged; notes credit #121.
- **#122 SnapKV: not in the perf table** — no benchmarked row uses SnapKV (long-context-only win). (Aside: the ~1K SnapKV probe on 27B-MTP surfaced a *pre-existing* SnapKV+MTP-decode incompatibility — `BatchForward2: _kvCache.Length=128 != startPos`, the evicted cache vs MTP verify — unrelated to this PR; worth a separate issue.)


pekkah:
Thanks @gemini-code-assist — adjudicated each finding against the codebase (independently cross-checked with the repo's review toolkit). Summary:

**1. Host-loop top-k reduce in `BatchedGpuMoeFfn` (high) — valid, deferred to #129.**
Correct that the Phase-3 per-token reduce issues ~N·(na+2) small stream ops; for large-N prefill the launch overhead undercuts the grouped-GEMM win. It is **bit-identical** to the sequential `GpuMoeFfn` accumulation (load-bearing FP order), so this is a throughput optimization, not a correctness fix, and it only gates the **non-default GPU-SLRU path** (`SHARPI_CPU_MOE=0`) — the benchmarked 35B-A3B default routes MoE to CPU (`BatchedRoutedExperts`, already batched) and never enters this loop. A proper fix needs a new batched weighted-scatter-reduce NVRTC kernel (must preserve the k-order), so I've tracked it as **#129** and added a code note pointing there rather than rushing a kernel into this PR.

**2. Per-call `Download(_gpuWGateInpShexp[layer], …)` (high) — false positive.**
The sequential reference this path must stay bit-identical to (`GpuMoeFfn`) downloads this same constant weight **once per token**; the batched version hoists it **out of the per-token loop** (once per layer-call, reused for all N). So #121 already improves this by ~N×. Caching at construction would help the sequential path equally and is out of scope here.

**3. `CopyDevice` for host↔pinned transfers ×3 (medium) — false positive (established convention).**
`AllocatePinned` registers the `cudaMallocHost` pointer as a UVA device pointer, and the existing per-token `Forward`/`GpuLayer` path already stages every `_gpuHidden`↔`_pinnedHidden` move via `CopyGpuBuffer` (== `RecordComputeCopy` == `CopyDevice`, D2D). The #123 code follows that exact UVA-dependent convention — it introduces no new dependency or risk, and the bit-parity oracles (`CudaHybridBatchedPrefillTests`, single/multi-chunk/>4096) pass. Switching #123 alone to `Download`/`UploadInto` would make it inconsistent with the rest of the class; a uniform migration would be a separate codebase-wide change.

No correctness/bit-parity issues — all parity oracles assert byte-identical logits vs the sequential reference and are green.

PR #133, opened 2026-06-04
perf(cuda,engine): batched GPU-SLRU MoE prefill (#129) + SnapKV×MTP gate (#130)

### LINKED ISSUE #129: perf(cuda): batched weighted scatter-reduce kernel for the GPU-SLRU MoE prefill (#121 follow-up)
Follow-up to #121 (PR #128). `BatchedGpuMoeFfn` (the GPU-SLRU routed-MoE batched prefill, reached only with `SHARPI_CPU_MOE=0`) batches the routed-expert gate/up/down GEMMs grouped-by-expert, but its **Phase 3 per-token reduce** and **Phase 1 per-token shared-expert** still loop on the host:

```csharp
for (int i = 0; i < N; i++) {            // per token
    _gpu.Clear(_gpuHidden);
    for (int k = 0; k < na; k++) {        // per active expert
        var partV = _gpu.View(downPartial, ...);
        _gpu.AddScaledInPlace(_gpuHidden, partV, weights[...]);  // host-issued
    }
    _gpu.AddInPlace(_gpuHidden, sharedV);
    _gpu.CopyDeviceRegion(hiddenAll, ...);
}
```

At `N = 4096`, `na = 8` this is ~120k sequential host-issued stream ops, whose P/Invoke + driver-launch overhead undercuts the grouped-GEMM batching win on this stage (flagged by Gemini review on PR #128).

## Scope
- A single NVRTC **batched weighted scatter-reduce** kernel: for each token row, sum the `na` unweighted down-partials in top-k slot order with the per-(token,slot) weights, then add the shared-expert row — in one launch over `N × embDim`.
- Must stay **bit-identical** to the current host reduce (which mirrors the sequential `GpuMoeFfn` `Clear` → per-k `AddScaledInPlace` (top-k order) → `AddInPlace(shared)` order). The per-token FP accumulation order is load-bearing (cf. the K/V-MatVecDual MTP-parity note) — the kernel must reduce in the same k=0..na-1 order per token.
- Optionally batch the Phase 1 shared-expert (gate/up/down) over N as well.

## Notes
- Only affects the GPU-SLRU config (`SHARPI_CPU_MOE=0`); the benchmarked 35B-A3B default routes MoE to CPU (`BatchedRoutedExperts`, already batched). So this is a correctness-neutral throughput optimization for the on-GPU-experts path, not a default-path regression.

## Tests
- The existing `BatchedTrunkGpuFfn_BitwiseMatchesSequential_GpuSlruMoe` oracle (with the batched flags pinned on) must stay bit-identical; add an A/B prefill measurement on the GPU-SLRU config to quantify the win.

### LINKED ISSUE #130: MTP batched-verify (BatchForward2) crashes when SnapKV eviction is active: _kvCache.Length != startPos
## Symptom

When SnapKV prefill eviction is active on an MTP model (e.g. `Qwen3.6-27B-MTP`), the first decode iteration throws:

```
BatchForward2: _kvCache.Length=128 != startPos=<promptLen>. Caches must be at startPos before the batched verify call.
```

(`128` is the SnapKV budget; `<promptLen>` is the original, pre-eviction prompt length.)

## Root cause

SnapKV eviction is a prefill-only compaction. After it runs, `PagedKvCache` deliberately splits its two length notions (see `PagedKvCache.cs`):

- `_length` (physical) shrinks to the kept-slot count `K` (= budget, e.g. 128) — the slot index of the next append.
- `_logicalLength` stays at the original prompt length `N` — the absolute position the next decode token sits at, used for RoPE so cached (already-RoPE'd) keys and the incoming query share a reference frame.

The single-token `Forward` decode path already handles this `Length != LogicalLength` split correctly (covered by `CudaHybridGdnSnapKv_LongPrompt_CacheShrinksToBudget_DecodeStaysWellFormed`).

`BatchForward2` (the MTP N=2 batched-verify primitive in both `HybridGdnForwardPass` and `CudaHybridGdnForwardPass`) opens with a precondition `if (_kvCache.Length != startPos) throw ...`. `MtpDecoder.DecodeBatched` passes `startPos = _nextPos` (= logical position). Post-eviction `_kvCache.Length == K` but `startPos == N`, so the precondition fails on the very first batched-verify call. There is no reconciliation between eviction and batched verify.

## Fix (this PR)

Cleanly **gate batched-verify off when the cache has been evicted** — detect `_kvCache.Length != _kvCache.LogicalLength` and fall back to the sequential MTP decode path (which uses `Forward`, already eviction-safe). SnapKV + MTP then coexist correctly, just without the N=2 verify speedup while eviction is active. Add a decode-after-eviction coherence test.

## Follow-up (not this PR)

Teach `BatchForward2` to operate on an evicted cache directly: change the precondition to key off `LogicalLength` (the RoPE position) and make the two-token append/attention use physical `Length` for storage slots and `LogicalLength(+0/+1)` for RoPE — mirroring what single-token `Forward` already does. That restores the verify speedup under eviction. Requires re-running the MTP byte-parity oracle (`MtpDecoder_GreedyParity_LlamaCpp`) since MTP FP parity is fragile.

pekkah:
Batched-prefill follow-ups from #128. Two independent changes + a README condensation, split into three commits.

## #129 — batched GPU-SLRU MoE prefill reduce (`perf(cuda)`)
Replaces `BatchedGpuMoeFfn`'s two host-side per-token loops with batched ops:
- **Phase 1 (shared expert):** batched GEMM-N gate/up/down + one `llm_scale_rows_inplace` for the per-token sigmoid gate.
- **Phase 3 (reduce):** one new `llm_moe_weighted_reduce` kernel does the top-k weighted scatter-reduce + shared add over all N tokens, replacing ~N·(na+2) tiny stream ops.

**Bit-identical** to the prior host reduce — each routed term contracts to the same `fmaf` (NVRTC `fmad=true`) as `add_scaled_inplace`, the shexp scale is a separate rounded pass (never folded into an fma), and the plain shared-add matches `add_inplace`, routed-first/shared-last. Only the non-default `SHARPI_CPU_MOE=0` GPU-SLRU path is affected.

**Verified:** new per-kernel oracle (`MoeWeightedReduce`/`ScaleRowsInPlace`, no model needed) + the 22 GB `BatchedTrunkGpuFfn_BitwiseMatchesSequential_GpuSlruMoe` end-to-end oracle.
**Benchmark** (Qwen3.6-35B-A3B, RTX 4070 Ti, `SHARPI_CPU_MOE=0`, ~1K ctx, warm, backend confirmed CUDA-hybrid GPU-SLRU): prefill **45.3 → 54.3 t/s (+20%)**, decode unchanged. Reproduce with `scripts/bench-129-ab.ps1`.

Closes #129.

## #130 — gate MTP batched-verify off on a SnapKV-evicted cache (`fix(engine)`)
`BatchForward2` requires `_kvCache.Length == startPos`, but SnapKV eviction leaves `Length` at budget K while `LogicalLength` (RoPE position) stays at prompt length N — so MTP decode threw `_kvCache.Length=128 != startPos=N`. `SupportsBatchVerify` now also requires `!KvCacheCompacted` (`Length != LogicalLength`), so `MtpDecoder` falls back to the eviction-safe sequential path; defensive preconditions retained. The signal is false in all normal operation, so the working non-evicted batched path is unaffected. Coexisting batched-verify with eviction is the follow-up.

**Tests:** non-vacuous decode-after-eviction regression test (proven to fail without the gate), a no-eviction gate-false-positive guard, a `PagedKvCache` Length/LogicalLength invariant test, and a `SHARPI_TRACE_MTP` fallback trace line.

Closes #130.

## README
Condensed the perf-table notes and prose **340 → 199 lines** — kept every perf number, repo/size, the FastScan kernel table, and usage/flags; dropped the issue-by-issue implementation narrative.

## Process
Implemented via the subagent workflow with a code-review gate between tasks and a 4-agent toolkit review at the end (silent-failure-hunter, comment-analyzer, pr-test-analyzer, code-reviewer). No defects; high-value findings applied (incl. the toolkit catching a factual error in a comment, now fixed). Full solution builds clean (0 warnings); the #129-only commit also builds standalone. All SnapKV/MTP/kernel tests + the GPU-SLRU oracle pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,124 +12,84 @@ OpenBLAS in `tools/openblas/` for faster batched GEMM. Build with `dotnet build 
 ## Text generation
 
 Supported architectures: `llama`, `llama4`, `olmoe`, `qwen3`, `qwen3moe`, `qwen35moe`
-(hybrid Gated-DeltaNet + attention + MoE), `gemma4` (per-layer head_dim, SWA + global, PLE). Benchmarked on
-AMD Zen 4 (12c/24t, DDR4-3200) + RTX 4070 Ti (12 GB), Q4_K_M, `--temp 0`,
-`-n 80`, prompt `"Write a Python function that sorts a list using the quicksort algorithm:"`.
-Decode rate is **forward-pass iterations / decode time**, so it counts
-thinking-mode tokens too. All outputs verified coherent
-(`scripts/bench-all.ps1`). Cross-engine top-1 parity vs llama.cpp b8585
-verified on Qwen3-8B (byte-identical 60-token greedy decode with
-matching chat template).
+(hybrid Gated-DeltaNet + attention + MoE), `gemma4` (per-layer head_dim, SWA + global, PLE).
+Benchmarked on AMD Zen 4 (12c/24t, DDR4-3200) + RTX 4070 Ti (12 GB), Q4_K_M, `--temp 0`.
+**Prefill t/s** is the warm-cache rate at a ~1K-token prompt; **decode t/s** is the near-zero-ctx
+generation rate (forward-pass iterations / time, so thinking tokens count). All outputs verified
+coherent (`scripts/bench-all.ps1`); top-1 parity vs llama.cpp b8585 verified on Qwen3-8B (byte-identical
+60-token greedy decode).
 
 | Model | Repo | Size | Backend | Prefill t/s | Decode t/s | Notes |
 |---|---|---:|---|---:|---:|---|
 | SmolLM2 1.7B Instruct | [HuggingFaceTB](https://huggingface.co/HuggingFaceTB/SmolLM2-1.7B-Instruct-GGUF) | 1 GB | CPU | 40.4 | 38.9 | AVX2 fused dequant-matvec |
 | SmolLM2 1.7B Instruct | (same) | 1 GB | Vulkan `-g -1` | 123.2 | **139.7** | GLSL `subgroupAdd` reduce |
 | SmolLM2 1.7B Instruct | (same) | 1 GB | **CUDA** `-g -1` | **163.1** | **158.1** | NVRTC `__dp4a` + Q8_1 |
-| Qwen3 8B | [Qwen](https://huggingface.co/Qwen/Qwen3-8B-GGUF) | 5 GB | CPU | 9.9 | 11.7 | dense forward pass, no KV compression |
-| Qwen3 8B | (same) | 5 GB | CPU `--tq` | 9.5 | **11.9** | 3-bit KV → up to 40 960 ctx. FastScan K + V kernels (issue #34) deliver ~20× over the per-block AVX2 path on the attention K+V hot loop; visible end-to-end at long context: **10.2 t/s @ 3K, 9.4 t/s @ 6K** (vs ~5 t/s baseline at 6K). Batched prefill is TQ-aware (issue #34 follow-up) so prompt processing matches the no-`--tq` rate |
+| Qwen3 8B | [Qwen](https://huggingface.co/Qwen/Qwen3-8B-GGUF) | 5 GB | CPU | 9.9 | 11.7 | dense, no KV compression |
+| Qwen3 8B | (same) | 5 GB | CPU `--tq` | 9.5 | **11.9** | 3-bit KV → 40 960 ctx; FastScan K+V (#34) keeps long-ctx decode ~flat (10.2 @ 3K, 9.4 @ 6K) |
 | Qwen3 8B | (same) | 5 GB | Vulkan `-g -1` | 45.4 | 45.8 | 11.4K auto-ctx |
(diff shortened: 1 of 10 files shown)


pekkah:
Thanks — applied the 2D-grid suggestion to both kernels in 75d219a. `llm_scale_rows_inplace` and `llm_moe_weighted_reduce` now launch with a 2D grid (x = element/column, y = row/token) and recover `(i, e)` directly from block/thread indices, dropping the per-thread integer divide/modulo. This also removes the `checked((int)total)` host cast; I added a loud `gridDim.y <= 65535` guard so an oversized N/rows fails fast rather than silently skipping rows.

Bit-identical (only the index→(i,e) mapping changed; the per-term `fmaf` accumulation, k-order, and plain shared-add are untouched) — re-verified by the per-kernel oracle (`MoeWeightedReduce_BitwiseMatchesSequentialReduce` / `ScaleRowsInPlace_BitwiseMatchesPerRowScaleInPlace`) and the 22 GB `BatchedTrunkGpuFfn_BitwiseMatchesSequential_GpuSlruMoe` end-to-end oracle.

Issue #136, opened 2026-06-04
Gemma 4 CUDA perf: collapse per-token kernel launches + enable batched prefill


pekkah:
> ## ✅ RESOLVED (2026-06-14)
> The headline launch-collapse + batched-prefill work shipped: PLE projection fused (`llm_ple_proj_norm`), batched-trunk prefill enabled for Gemma 4 (#137), and CUDA-graph decode capture/replay made real and **default-on** (#139/#142 — `SHARPI_CUDA_GRAPH=0` reverts), plus the attnScale and dual-HeadNorm fusions. Prefill/decode then moved to cuBLAS GEMM / dp4a (#141/#142). Decode is now HBM-bandwidth-bound (matvecs ~90% HBM), so the two remaining P1 micro-fusions (`layer_output_scale`, `EmbeddingScale` bake) are moot — there is no launch-count lever left. Remaining prefill **compute** lever lives in #152. Closing.
>
> ---

Performance review of the Gemma 4 CUDA execution path (`CudaForwardPass.ForwardGemma4`, `CudaHybridForwardPass.GpuLayerGemma4`/`Prefill`).

**Key finding:** Gemma 4 decode is **launch-bound**, not compute-bound. The inner loop issues ~28 kernel launches/layer (~950/token over ~34 layers) plus ~200 upfront for PLE projection prep — **~1,100–1,200 kernel launches per token**. At ~4–6 µs host launch cost that's ~5–7 ms/token of pure host overhead (effective ceiling ~150–200 t/s before any compute). PLE alone is ~half the launches. Numerics (SWA bounding, dual-RoPE, KV-share, per-layer head_dim) are already correct — the wins are in launch count and prefill batching.

The recent GDN/MoE/MTP perf work is architecturally irrelevant to Gemma 4 (dense, no SSM, no experts), but two pieces transfer: the non-owning `View` slice type (#111) and the batched-trunk GEMM-N machinery (#123, currently gated off for Gemma 4).

### P0 — Collapse PLE launches (biggest decode win)
- [x] Replace PLE staging `CopyDeviceRegion` round-trips with `CudaBackend.View(parent, elemOffset, elemCount)` (added in #111). The "Tensor can't encode a device-pointer offset" comments at `CudaForwardPass.cs:1122-1126` and `1156-1158` are stale. Zero numeric change.
[...]

Issue #111, opened 2026-06-03
perf(engine,cuda): batch the GDN-hybrid prefill trunk (GEMM-N attn/GDN projections) — trunk is now ~62% of prefill (#110 follow-up)


pekkah:
## Summary

Follow-up to #110. The batched prefill landed in #110 cut the **routed MoE** to ~2.7× (≈25→9.2 ms/tok) and overall prefill to **1.34×** (30→40.5 tok/s on Carnice). The win is now Amdahl-capped: with the routed MoE batched, the **trunk (attention + GDN blocks) is ~62% of prefill wall** and still runs **one token at a time** on the GPU.

Per-512-token-chunk split after #110 (`SHARPI_PREFILL_PROFILE=1`):

```
[batched-prefill] N=512 total=13312ms (26.0ms/tok) trunk=8282ms router=299ms routedMoE=4707ms combine=15ms
```

## Root cause

`PrefillBatchedCpuMoe` runs the trunk as a sequential per-token loop (correct — the GDN recurrence and KV append are positional). But the trunk is **host launch-overhead bound**: ~15 tiny single-token GPU launches per token per layer (`CopyDeviceRegion`, `RmsNorm`×2, the attn/GDN block's matmuls, `AddInPlace`, `DownloadAsync`×2/3, shared-expert matmuls). At 512 tokens × 40 layers that's ~300k `cuLaunchKernel` calls/chunk; the single-token matmuls are latency-bound, not throughput-bound, so the GPU is badly underutilized.

## Proposed work

Batch the **non-recurrent** parts of the trunk into GEMMs over the N prompt tokens, keeping only the positional parts per-token:

- **GDN block:** batch the joint QKV projection, z/gate projection, alpha/beta projections, and `ssm_out` projection as GEMM-N (one weight read amortized over N tokens). The conv1d + delta-net recurrence stay sequential.
- **Attention block:** batch Q/K/V/O projections + per-head Q/K RMSNorm + RoPE as GEMM-N; KV-append + scaled-dot-product attention stay per-position (or use a batched prefill attention kernel).
- Batch `RmsNorm` over N tokens (one launch instead of N).

## Blockers / prerequisites

- No general **GEMM-N** kernel exists on `CudaBackend` — only `MatMul` (GEMV) and `MatMulN2` (2-input). Needs a quantized weight × N-token-input GEMM for Q4_K/Q5_K/Q6_K/Q8_0/F32.
[...]

PR #137, opened 2026-06-05
perf(gemma4-cuda): collapse per-token launches + all-GPU batched-trunk prefill (#136)

pekkah:
## Gemma 4 CUDA performance: collapse per-token launches + batched-trunk prefill

Closes the work tracked in #136. Gemma 4 decode on CUDA is **launch-bound** (~1,000+ kernel launches/token), and prefill ran a per-token loop. This PR collapses launches on the decode path and adds an all-GPU batched-trunk prefill, plus two launch fusions.

### Measured impact (E4B-it Q8, RTX 4070 Ti, `-g -1 -c 2048`, ~1K warm)
| | Before | After | Δ |
|---|---:|---:|---|
| Prefill t/s | 47.7 | **109.4** | ~2.05× (`SHARPI_BATCHED_PREFILL=0` A/B: 53.4 → 109.4) |
| Decode t/s (near-zero-ctx) | 44.1 | **49.1** | +11% |

### What's in it (8 perf commits + 1 docs)
**P0 — decode launch collapse**
- PLE staging copies → non-owning `View` slices (the "Tensor can't encode a device offset" comment was stale; `View` from #111 does exactly that).
- PLE per-slice projection loop → `RmsNormBatched` + 2 full-buffer ops (~204 → 3 launches/token).

**P0 — all-GPU batched-trunk prefill** (`CudaForwardPass`)
- Embeds N tokens, builds PLE projections batched, runs every layer batched across N, final norm/output on the last token. Per-layer head_dim via max-sized scratch + exact-N views.
- Four new batched kernels, each with a bit-exact oracle: `RoPEWithFactorsBatched`, `AttentionSwaBatched`, `llm_matvec_q8_0_gemm_n` (unblocks the Q8_0 model), `llm_gelu_tanh_mul_strided` (gather-free batched PLE).
- Gated by `IsGemma4BatchedPrefillSupported` (dense, NEOX, non-L2 QK-norm, no attn-bias, TQ off, batchable dtypes, ctx ≤ 4096); SnapKV-active / unsupported configs fall back to the bit-exact per-token loop. `SHARPI_BATCHED_PREFILL=0` forces sequential.

**P1/P2 — launch fusions**
- Drop the per-layer Q-prescale via an `attn_scale` kernel param (sentinel `-1f` keeps every non-Gemma caller bit-identical; Gemma passes `1.0` and now matches the CPU `scale=1.0` reference exactly).
- Fuse Q+K HeadNorm into one launch (`llm_head_norm_qk` + batched).

### Verification
- Every new kernel has a `*_BitwiseMatchesSingle/Sequential/Separate` oracle.
- End-to-end `Gemma4_E4B_BatchedPrefill_MatchesSequential` confirms batched ≡ per-token post-prompt logits on the real Q8_0 GGUF.
- **Full CUDA regression green: 111/111** (all non-Gemma model paths — Qwen/GDN/MoE/Coder — confirm the `attn_scale` sentinel + shared kernel-array additions are inert). 0 warnings (TreatWarningsAsErrors).

### Deferred (tracked as open boxes on #136)
- CUDA graphs (the structural fix for the launch-bound ceiling; large, its own effort — no-op `BeginRecord`/`RecordBarrier` scaffolding already present).
- `layer_output_scale` fold / `EmbeddingScale` bake (~1 launch/layer each; marginal vs. residual-path risk).


pekkah:
## Review addressed (toolkit + gemini)

Ran the pr-review-toolkit (code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer) plus the gemini-code-assist comments. Fixes pushed in `5b1a859` (toolkit) and `9ae0329` (gemini). 28/28 Gemma 4 CUDA + batched-kernel oracles green after the changes.

**Code / safety**
- **Fail-safe prefill gate** — new `BatchableWeight()` treats an *unregistered* weight handle as not-batchable (was `WDType → Q4_K` default, which biased the gate toward "supported"); also gate on `_gpuPerLayerModelProj`'s dtype. An untracked weight now forces the per-token fallback instead of being misdispatched into GEMM-N.

**Tests** (the top finding — the central change was under-covered)
- Added `AttentionExplicitScale_BitwiseMatchesSingle_AndDiffersFromSentinel`: exercises the `attn_scale=1f` path (Gemma's production path) bit-exact for both batched SWA and full attention, and asserts `1f` ≠ the `rsqrt` sentinel so the param is provably plumbed. Previously only end-to-end coverage hit it.
- Tightened the end-to-end parity test to assert a >95% bit-exact fraction (the `exact` count was computed but unused), not just the loose `maxAbs<1e-2` envelope.

**Comments / docs**
- `<param name="attnScale">` added to all 4 attention dispatches (the `≤0 → 1/sqrt(headDim)` sentinel contract).
- Fixed the PLE-collapse note (≈6 ops×L, not 3×L) and the batched-scratch note (Q/K/V/attn buffers are max-head_dim-sized + viewed per layer).

**Bench script (gemini)**
- `$ModelDir` param instead of a hardcoded path; `try/finally` around the `SHARPI_BATCHED_PREFILL` toggle.

**Deferred to a separate issue (#138), not a blocker**
- silent-failure-hunter flagged the Gemma SWA KV cache (window-sized buffer + absolute indexing → potential OOB when `ctx > sliding_window`). It's **pre-existing** and identical in the per-token and batched paths (so parity holds), and the benched/tested configs don't hit it. Filed as #138 for a proper ring-buffer-vs-full-length fix; out of scope for this parity-preserving perf PR.

Issue #146, opened 2026-06-06
perf(cuda): full tensor-core (mma.sync) flash-attention prefill for d=512 — beat the half2 kernel (#141 follow-up)


pekkah:
> ## ✅ RESOLVED by PR #148 (2026-06-06)
> Tensor-core (`mma.sync`) flash-attention prefill shipped for the Gemma 4 head dims: #146 single-warp (+5%, occupancy-limited at d=512) and the #147 multi-warp / d-split register-resident-O variant (+27–40%, **default-on**) which beats the half2 kernel — this issue's goal. Parity 3e-4 vs the sequential `AttentionBatched` oracle. TC-flash test-coverage hardening is tracked in #151; the deeper MMQ-GEMM prefill lever in #152.
>
> ---

Follow-up to #141. PR #143 landed a **half2 fp16x2** flash-attention prefill kernel (`llm_flash_attn_prefill_f32`): one-warp-per-query, shared K/V tile reuse, online softmax, half2 QK dot, scalar-fp32 PV. That took prefill 1564 → 2853 t/s and made attention (previously 67% of prefill, BW-bound) far cheaper.

Remaining gap to llama.cpp (~8475 t/s) is now attention **compute**. The next lever is a **full int8/fp16 tensor-core (mma.sync) flash kernel** — QK *and* PV on tensor cores — analogous to llama.cpp's `fattn-mma-f16.cuh`.

**Why it was deferred from the #143 session (the half2 path was chosen instead):**
- A study of llama.cpp's `fattn-mma-f16.cuh` showed that at Gemma 4's global `head_dim = 512`, the full-TC flash kernel runs **occupancy-1 even in llama.cpp** — the register/shared pressure of a 512-wide head dim is brutal. So the expected win over half2 was uncertain and the kernel is high-risk.
- half2 captured most of the QK speedup at a fraction of the implementation risk.

**Scope (not an estimate, just the pieces):**
- mma.sync m16n8k32 s8/s8 (or f16) fragments for QK^T and for P·V, online-softmax rescale between tiles.
- Handle the two Gemma 4 head dims (512 global / 256 SWA) and GQA 4:1.
- d=512 occupancy/register-pressure mitigation (split-D accumulation or 2-pass).
[...]

Issue #147, opened 2026-06-06
perf(cuda): multi-warp / d-split TC flash prefill — fix the single-warp occupancy limit (#146 follow-up)


pekkah:
Follow-up to #146. The single-warp TC flash (`llm_flash_attn_prefill_tc`, PR on `perf/gemma4-cuda-flash-tc-146`) is correct and +5% over half2, but **occupancy-limited**: 1 warp/block + 48 KB shared (O[16×512] fp32 + K/V fp16) → only ~2 warps/SM on the 4070 Ti, so the tensor cores stall on memory/sync latency with no other warps to hide it. That caps the win far below the tensor cores' arithmetic ceiling and is why prefill is still ~2849 vs llama.cpp's ~8475.

**The unlock is more warps per query-tile**, matching llama.cpp's `fattn-mma-f16` structure:
- **d-split across warps**: W warps each own head_dim/W output columns → O stays *register-resident* (16×128 = 64 regs/lane at W=4) instead of in shared, killing the per-key-tile shared-O rescale traffic. Warps compute partial QK^T over their d-slice and reduce S across warps (shared/shfl) before the softmax, then each does P·V for its d-slice.
- frees the 32 KB shared currently held by O → room for **double-buffered K/V tiles** (`cp.async`) to overlap staging with compute.
- target: many more resident warps/SM → tensor cores saturated.

**Scope (pieces, not estimates):**
- restructure to W=4 (or 8) warps/block; cross-warp S reduction for softmax; register-resident O per d-slice.
- `cp.async` double-buffered K/V staging (note: K/V are fp32 in the cache and converted to fp16 on stage — the async copy is fp32→shared, convert on read).
- reuse the validated `mma.sync m16n8k16` primitive and the no-transpose score→P trick from #146.
- parity vs the scalar oracle (argmax-stable), same harness as `CudaFlashAttnTcTests`; A/B vs the +5% single-warp baseline.

Once it beats the single-warp kernel, flip `SHARPI_PREFILL_FLASH_TC` default-on (and pin it off in the Gemma4 bit-exact oracles).

pekkah:
**Done — the d-split occupancy fix landed and it is a big win.** Folded into PR #148.

[...]

PR #148, opened 2026-06-06
perf(gemma4-cuda): TC flash-attention prefill (#146/#147) + SoA Q8_0 weight layout (#149) — prefill 2853→4240

pekkah:
## Summary
Two stacked, default-on Gemma 4 CUDA prefill wins, each built bottom-up and validated. **Prefill 2853 → 4240 t/s (1.8K ctx) = 34% → 50% of llama.cpp's ~8475.** All fast paths argmax-stable; the SoA repack is bit-identical. Full ForwardPass suite **382/382** green.

## 1. Tensor-core flash-attention prefill (#146/#147)
Both QK^T and P·V on the mma cores (`mma.sync.m16n8k16.f32.f16.f16.f32`), replacing the scalar O(n²) per-query attention.
- **Validated the mma fragment layouts in isolation first** (`llm_mma_test_m16n8k16_f32`, maxAbs 4.77e-7) — a wrong lane→register map silently produces garbage.
- **#146 single-warp**: the QK^T score C-fragment is reused *directly* as the P·V A-fragment (no transpose — the key index is QK^T's N-col and P·V's contraction, and the C/A layouts coincide on `(row, 2·tig)`); O[16×512] lives in shared (256 regs/lane otherwise). +5%, occupancy-limited.
- **#147 multi-warp / d-split**: W=4 warps split the head dim so O is *register-resident* — occupancy ~2 → ~16-20 warps/SM. **+27% @1K / +40% @1.8K** over the half2 kernel. Default-on (`SHARPI_PREFILL_FLASH_TC=0` reverts).
- Parity ~3-4e-4 vs the scalar kernels across GQA / 256-SWA & 512-global / windowing / partial tiles.

## 2. SoA Q8_0 weight layout (#149)
Repacks 2-D Q8_0 weights at upload into `[quants rows*cols B][scales rows*nb fp16]` so all readers load the quants 16-byte aligned instead of the interleaved 34-byte block's 2-byte-misalignment `__funnelshift`.
- One-time GPU repack (`llm_q8_0_repack_soa`); all five Q8_0 readers (MMQ, dp4a, fp32 matvec, GEMM-N, dequant) get bit-identical SoA variants and **auto-route per repacked handle**.
- **Bit-identical end-to-end** (`CudaMmqSoaE2ETests`: prefill + 4 decode, maxAbs 0) — the bit-exact oracles exercise the SoA path (converting *every* reader is what keeps them meaningful; a half-migration would false-pass on garbage).
- **+10-12% prefill by default** (N=965 3297→3698, N=2054 3849→4240). `SHARPI_MMQ_SOA=0` reverts.

## Dead-ends ruled out (documented, reverted)
- **split-K MMQ**: occupancy isn't the lever at real prefill batch sizes (probe mismeasured at nTok=1024).
- **cp.async-pipelined MMQ**: bit-identical but slower than register-prefetch on Ada (barriers add overhead; `.ca` beats `.cg` but both lose).

## Diagnostics kept
`CudaMmqRooflineProbe` (int8 TOPS vs peak at FFN shapes) — but probes must use the *real* prefill nTok.

## Test plan
- Focused: `dotnet test tests/SharpInference.Tests.ForwardPass --filter "Gemma4Cuda|CudaFlashAttn|CudaMmq"` (26 green).
- Full ForwardPass suite: **382/382, 0 failed**.


pekkah:
---
## Update: multi-warp / d-split (#147) landed — default-on, +27-40%

The single-warp kernel above was occupancy-limited (+5%). The follow-up `llm_flash_attn_prefill_tc2` fixes it: **W=4 warps cooperate on each 16-query tile, head dim split across them → O is register-resident** (no shared-O rescale), occupancy ~2 → ~16-20 warps/SM. Partial QK^T per warp reduced through a small shared S buffer; same no-transpose score→P trick downstream.

| N | half2 | tc1 (#146) | tc2 (#147) | tc2 vs half2 |
|---|---|---|---|---|
| 965 | 2582 | 2691 | **3269** | **+26.6%** |
| 2054 | 2732 | 2851 | **3818** | **+39.8%** |

Parity identical to scalar (~3-4e-4). **Now default-on** (`SHARPI_PREFILL_FLASH_TC=0` → half2). Pinned off in the bit-exact/half2 Gemma4 oracles; 23 focused tests green. README updated 2853→3269. Closes #146 and #147.

pekkah:
## pr-review-toolkit cycle (4 agents) — results

Ran code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer over `git diff master...HEAD`.

**No critical/high defects.** Verified by the agents: all 5 Q8_0 readers are SoA-routed; the one non-SoA reader (`MatMulN2`) throws on Q8_0 rather than reading garbage; the embedding/MoE weights are never repacked (uploaded outside `UploadWeight`); NVRTC failures surface as load-time exceptions, not inference garbage; repack byte-math verified; kernel comments (fragment layouts, shared-mem formulas, guard conditions) are accurate and internally consistent.

**Addressed (commit `e028ea8`):**
- **Test gap (sev 8)** — the GPU repack kernel + the fp32-matvec / GEMM-N / dequant SoA readers had no GGUF-free coverage (only bench-machine model oracles). Added `GpuRepack_AllSoaReaders_BitIdenticalToInterleaved`: repacks on the GPU (production path) and asserts all 5 readers bit-identical to interleaved across 3 shapes — runs on any CUDA box. ✅ all green.
- **`_soaHandles` lifecycle** — the whole SoA-correctness invariant rested on this set. Switched `HashSet`→`ConcurrentDictionary` (matches the sibling handle tables, removes the latent thread-safety footgun) and prune it on `Free`/`Dispose` (was an unbounded leak across model load/free cycles).
- **README** — balanced a stray parenthesis.

**Noted, not addressed (lower-value test hardening, optional follow-ups):**
- TC flash kernels are only tested with `startPos==0`; continued-prefill (chat cache) can pass `startPos>0` — worth one parity config.
- No `nTok < 16` (single partial tile) or genuinely TC1-only head_dim (`%16 && !%64`, e.g. 48/80) config; Gemma 4's 256/512 are both `%64` so TC1 is only force-exercised via `RunParity(tc2:false)`.
- Comment perf/occupancy figures ("~2 warps/SM", "23-34% of TC peak") lack a "measured on RTX 4070 Ti" anchor and will rot.

These are hardening suggestions, not bugs — deferring.

Issue #162, opened 2026-06-06
perf(cuda): close remaining Qwen3-8B Q4_K DECODE gap to llama.cpp — non-matvec cost (prefill handled by #167; kernel-efficiency in #149/#152)


pekkah:
Follow-up to #156/#158/#160 (PR #161). Qwen3-8B Q4_K_M on RTX 4070 Ti is now **prefill 432 / decode 74.7 t/s**; llama.cpp b8585 is **pp1008 ~5764 / tg128 ~78–91**. Two distinct gaps remain, each needing its own approach.

## Decode (74.7 → ~78–91) — the lever is NOT the matvec

The Q4_K decode matvec is now ~**89% of HBM peak** after the #160 SoA repack (was 74%), vs Q8_0's ~90%. Per `CudaDecodeMatvecQ4KRooflineProbe`, that's near the realistic ceiling — squeezing the last ~1pp (the min-correction's 2 extra dp4a + `0x01010101` sum, the d8/scale float mults, or occupancy) is low-yield. **Confirm with the probe before investing.**

The real signal: the matvec gained +13–15% BW but e2e decode only moved +7%, so **~half the per-token cost is elsewhere**. Mirror the Gemma win (#142, which turned out to be the sampler, not kernels):

- [ ] Run `SHARPI_CUDA_PROFILE=1` per-phase decode breakdown on Qwen3-8B and identify the dominant non-matvec cost (attention / RoPE / norms / KV append / sampler).
- [ ] Check the sampler path specifically — Qwen3-8B greedy `--temp 0` may still hit a full-vocab artifact like Gemma's `bench --verbose-prompt` LINQ-sort (see `project_gemma4_decode_sampler_142`).
- [ ] Fix the top item; argmax-stable; same-session A/B; README.

## Prefill (432 → ~5764) — cp.async-pipelined MMQ

The ~13× gap is llama.cpp's `mul_mat_q` **cp.async double-buffered** weight load, which hides the per-token-tile weight re-read that our `llm_mmq_q4k`/`llm_mmq_q4k_soa` (and cuBLAS, via L2) pay. The SoA layout (#160, quants 16B-aligned) is the right substrate for cp.async.

- [ ] Prototype `cp.async.cg.shared.global` (sm_80+) double-buffered tile load in `llm_mmq_q4k_soa`; overlap the next tile's global→shared copy with the current tile's mma.
[...]

Issue #149, opened 2026-06-06
perf(cuda): split-K + better-tiled int8 MMQ GEMM — 23-34% → 60-80% of TC peak (#141 follow-up)


pekkah:
> ## ✅ RESOLVED / superseded by #152 (2026-06-14)
> Of the approaches here: **split-K was ruled out** (occupancy isn't the bottleneck at real prefill batch sizes — end-to-end wash, reverted), **cp.async staging was slower on Ada** (reverted), and the **SoA Q8_0 repack shipped** (+10–12% prefill, bit-identical). The int8 MMQ still sits at ~40–43% of TC peak; that remaining "different tiling regime" lever — the only one not ruled out — is consolidated in **#152**. Closing in favor of #152 to avoid duplicate tracking.
>
> ---

Follow-up to #141. After the TC flash work (#146/#147) moved the prefill bottleneck to the matmul/FFN GEMMs, a roofline probe (`CudaMmqRooflineProbe`) shows the int8 MMQ kernel runs at only **23-34% of the 4070 Ti's ~160 TOPS dense int8 TC peak** — ~3× headroom, matching the 2.2× end-to-end prefill gap to llama.cpp (~8475 vs our 3818 t/s).

**Worst shape: `ffn-down` [out=2048 × in=8192], 23% peak.** Few output rows → few CTA tiles → the SMs are occupancy-starved over the long (8192) contraction, exactly the class of problem the single-warp flash kernel had before #147's d-split.

**Scope (pieces, not estimates):**
- **Split-K** the contraction across CTAs for tall/thin-output GEMMs (ffn-down) so all SMs get work; reduce the partials (atomic or a second pass). This alone should lift ffn-down's 23%.
- Better **register tiling** / larger per-warp output tiles to raise arithmetic intensity (more mma per weight load).
- **cp.async** double-buffered weight/activation staging — complicated by the Q8_0 `qs` 2-byte misalignment (the same funnelshift issue noted for decode); may need a re-aligned Q8_0 staging or a 16-byte-aligned shared layout.
- Cross-reference llama.cpp `ggml/src/ggml-cuda/mmq.cuh` (on disk at `C:\p\llama.cpp`) for their tiling/pipelining constants.
- Parity: the existing `CudaMmqQ8_0Tests` oracle (argmax-stable vs CPU `DotQ8_0`); A/B with the roofline probe (target: push the three FFN/qkv shapes toward 60-80% peak).

[...]

Issue #142, opened 2026-06-05
Gemma 4 GPU decode is ~1.8-2.9x slower than llama.cpp (kernel efficiency + low-ctx clock cliff)


pekkah:
> ## ✅ RESOLVED (2026-06-14)
> Decode reached ~70.5 t/s (E4B Q8, RTX 4070 Ti) — greedy **71.5 vs llama.cpp 76.6 = 1.07×**, within the "≤1.3× and flat" acceptance — via the top-k-first penalty-aware sampler (#154, +44%), dp4a/Q8_1 decode matvec (#143), and default-on CUDA graphs (#139). Problem 1 (the "low-ctx clock cliff", 27 t/s at depth 0) was diagnosed as a `--verbose-prompt` full-vocab-LINQ-sort **bench artifact**, not a GPU stall — real matvecs run at ~90% HBM. Steady-state decode is bandwidth-bound at the ~504 GB/s ÷ 8 GB-per-token ceiling, so there is no structural lever left. Closing.
>
> ---

## Finding

Gemma 4 GPU **decode is ~1.8–2.9× slower than llama.cpp**, and unlike llama.cpp it *degrades*
at low context.

Benchmark (RTX 4070 Ti, `gemma-4-E4B-it-Q8_0`, all layers on GPU):

| Decode | llama.cpp b9529 | SharpInference | gap |
|---|---|---|---|
| near-zero ctx (tg128 @ d0) | **78.5 t/s** | ~27 t/s | ~2.9× |
| ~1K ctx (tg128 @ d1024) | **77.7 t/s** | ~42–45 t/s | ~1.8× |

llama.cpp holds **~78 t/s flat** from depth 0 → 1024. Ours is *faster at 1K (42) than at
near-zero (27)* — backwards.

## Two distinct problems

1. **Low-context under-utilization / clock cliff.** Our near-zero-ctx decode (27) is slower
   than our 1K decode (42) because the GPU never reaches boost clock — the per-token work is a
   long chain of tiny, serialized kernels with idle gaps, so power draw stays low. llama.cpp
   keeps the GPU busy enough to boost even at depth 0 (78 flat). This is a *symptom of an
   inefficient/launch-gappy decode*, not an inherent GPU limit. (The merged CUDA-graph path
   targets exactly these launch gaps but the per-token `SetParams` overhead makes it −9.7% at
   short ctx; the device-side-`position` follow-up #140 is the only graph variant that might
[...]

PR #163, opened 2026-06-07
perf(cuda): chunked batched prefill for >4096-token prompts (#162)

### LINKED ISSUE #159: perf(cuda): compute-bound Q4_K prefill matmul — dequant->cuBLAS or Q4_K MMQ (#156 Item C)
## Goal

Give dense Q4_K models a **compute-bound prefill matmul** so the batched-trunk prefill (#156 Item A) closes the trunk-matmul gap, not just the attention gap. Split out of the #156 umbrella as Work Item C. **Optional** — only pursue if #156 Item A's prefill profiling shows trunk matmuls (not attention) dominating after flash attention lands.

## Background

#156 opened the batched-trunk prefill to dense Q4_K, but `GpuMatMulBatched` (`CudaForwardPass.cs:~1802`) only routes **Q8_0** to the compute-bound cuBLAS GEMM / int8 MMQ. Q4_K falls to the memory-bound matvec GEMM-N (`MatMulBatched`), which collapses launches but still re-streams each weight per token. So Item A's Qwen3-8B win (1.96×) is flash-attention + launch-collapse; the trunk matmuls are still HBM-bound.

## Options

- [ ] **C1 (cheaper):** extend `MatMulBatchedGemm` to dequant Q4_K → fp16 → cuBLAS GEMM (mirrors the existing Q8_0 path). Reuses cuBLAS; gets weight-read-once amortization; argmax-stable like the Q8_0 GEMM path.
- [ ] **C2 (maximal):** a Q4_K int8 MMQ tensor-core kernel (llama.cpp `mul_mat_q` Q4_K as reference). Largest effort; this is the genuinely-new-kernel item.

## Decision gate

Run #156 Item A's `SHARPI_PREFILL_PROFILE`/region profiler on Qwen3-8B at ~1-2K ctx. If attention is still the dominant prefill cost, defer. If trunk matmuls dominate, do C1 first and re-profile before considering C2.

## Verification

Argmax-stable parity vs the per-token loop (extend `Qwen3CudaBatchedPrefillTests`); before/after prefill t/s on Qwen3-8B Q4_K.

Refs: #156 (umbrella, Item A context), branch `perf/cuda-dense-q4k-batched-prefill-156`.

pekkah:
## Summary

This PR started as the #159/#162 profiling decision gates and produced one concrete shippable win plus a clear redirect of the remaining work.

### Shipped: chunked batched prefill for long prompts (#162)

Prompts longer than **4096 tokens silently fell back to the per-token prefill loop** — memory-bound, weight re-streamed per token. The batched-trunk gate hard-capped at `startPos + N <= 4096`, but that 4096 is a limit of the **non-flash** shared-scores `AttentionBatched` kernel (it throws above 4096). The flash prefill kernels (`llm_flash_attn_prefill_tc2/tc/f32`) stream KV in 16-position tiles with online softmax and register-resident O — **no 4096 cap**.

When flash is enabled (default) and the model has no SWA layers, we now run the batched-trunk path for prompts of any length up to the context size, **chunking into 4096-token windows** so the N-sized trunk scratch stays bounded. Each window is batched at its own `startPos`; flash attends to all prior KV.

| Qwen3-8B Q4_K, 4711-token prompt | prefill t/s |
|---|---|
| before (per-token fallback) | 49.9 |
| after (chunked batched) | **415.5 (8.3×)** |

Greedy `--temp 0` output is **argmax-identical** to the per-token reference. SWA models (Gemma 4) keep the proven 4096 cap — their window-ring KV semantics across chunk boundaries aren't yet validated (follow-up tracked in #162).

### Investigation findings (redirect, no code change)

**Decode (#162)** — Profiled Qwen3-8B: matvecs are **~89% of forward time** (ffn 58%, qkv 14%, lm_head 9%, o-proj 8%); non-matvec work (rope/qknorm/kv/attn) is only ~11%. The greedy sampler is a trivial O(vocab) argmax, **not** a Gemma-#142-style full-vocab artifact. Decode is genuinely **BW-bound on already-SoA-optimized matvecs** → no deletable hotspot. The only decode lever left is matvec kernel efficiency, which the issue itself rates low-yield. **Decode item: no easy win.**

**Prefill (#159)** — Decision gate confirmed: **trunk matmuls dominate (95.5%** at N=1844; attention 2.9%). But C1 (dequant→cuBLAS GEMM) **and** C2 (int8 MMQ) are **already shipped on master** (default-on), giving 2.8× over the memory-bound GEMM-N (153→433 t/s). So #159's stated deliverable is **done**.

The residual prefill gap to llama.cpp (433 vs 5764 t/s) is a **cold-L2 ~5× kernel-efficiency penalty**, reproduced precisely by the new `CudaQ4KPrefillMatmulProbe`: the isolated MMQ/GEMM kernels do the whole trunk in ~0.5s (looping one warm weight) but the real prefill — cycling the full 5GB model working set — spends ~4s in the same kernels at 100% SM / low DRAM util. That's exactly the **MMQ tiling-rewrite** territory of #149/#152 (large, risky; split-K and cp.async already ruled out there).

## Tests
- `Qwen3_8B_ChunkedBatchedPrefill_Over4096_MatchesSequential` — 5040-token prompt forces the two-window chunked branch; asserts argmax equality + top-5 overlap vs the bit-exact per-token loop.
- `CudaQ4KPrefillMatmulProbe` — real-shape (nTok=1844) roofline probe for the Q4_K MMQ and cuBLAS-GEMM trunk paths.
- All **25 batched-prefill tests green** (Qwen3 ×5, Gemma4, both Hybrid suites) — Gemma4's ≤4096 path confirmed intact by the SWA guard.

Closes #159. Refs #162 (decode no-win + cold-L2 prefill lever → #149/#152; Gemma SWA chunking follow-up).

### DIFF
--- a/src/SharpInference.Engine/CudaForwardPass.cs
+++ b/src/SharpInference.Engine/CudaForwardPass.cs
@@ -176,6 +176,14 @@ public sealed unsafe class CudaForwardPass : IForwardPass
     /// </summary>
     public bool BatchedPrefillEnabled { get; set; }
     /// <summary>
+    /// Issue #162: window size (tokens) for chunked batched prefill of prompts longer
+    /// than the non-flash 4096 cap. Each window is batched at its own startPos with flash
+    /// attention streaming the prior KV, so the N-sized trunk scratch stays bounded to
+    /// this many tokens regardless of prompt length. 4096 matches the well-tested
+    /// single-shot batch size.
+    /// </summary>
+    private const int PrefillBatchChunk = 4096;
+    /// <summary>
     /// Issue #141: route Q8_0 trunk matmuls in the batched prefill through the
     /// compute-bound cuBLAS GEMM (<see cref="CudaBackend.MatMulBatchedGemm"/>)
     /// instead of the memory-bound matvec GEMM-N. Default on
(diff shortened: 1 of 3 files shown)


pekkah:
Addressed the review (commit 4fe26c4):

- **SWA guard fail-closed** (silent-failure-hunter + pr-test-analyzer): hardened to `_hp.IsSwaLayer is null && _hp.SlidingWindowSize <= 0`. Today `IsSwaLayer` is only populated for Gemma 4, so the explicit window check fails closed if SWA parsing is later extended to a uniform-window arch that sets the window without a per-layer pattern array.
- **Exact-boundary test** (pr-test-analyzer): the chunked parity test is now a `[Theory]` covering both N=5040 (full+partial window) and **N=8192 (exact 4096-multiple, two full windows)** to catch chunk-loop off-by-one. Both pass.

@gemini-code-assist re the `ArraySegment` → `ToArray` allocation: that `tokens as int[] ?? ToArray` path is inside the **Q8_0-embed** branch of `PrefillBatchedTrunk`, which is Gemma-4-only — and Gemma 4 (SWA) is excluded from the chunked path by the guard above. Dense non-SWA models (the only ones that chunk) use the non-Q8_0 embed loop that indexes the `ArraySegment` directly, so no per-chunk heap allocation occurs on the chunked path. Prefill setup is also not the per-token hot path. Left as-is.

PR #167, opened 2026-06-07
perf(cuda): compute-bound Q6_K/Q5_K prefill closes the Qwen3-8B gap (#162)

pekkah:
## Summary

Closes the dominant Qwen3-8B Q4_K_M prefill gap to llama.cpp — **432 → 2431 t/s @1008 (5.6×)** — and the root cause was **not** the cold-L2 MMQ kernel penalty previously theorized.

A cold-regime probe (`CudaMmqColdProbe`: round-robin distinct weights, working set ≫ L2) reproduced the trunk in ~0.6 s — only ~1.2× the warm probe, not the 8× of real prefill. The real cause: **mixed `_M` quants**. `Qwen3-8B-Q4_K_M` keeps 37 tensors in **Q6_K** (half of `ffn_down` + `attn_v` + `output`), and Q6_K/Q5_K had no compute-bound prefill path — they fell back to the per-token GEMM-N matvec that re-streams the whole weight **once per token** (~94 GB of HBM traffic for one N=2290 `ffn_down`). That memory-bound fallback was the dominant large-N prefill cost.

## What changed

- New NVRTC kernels `llm_dequant_q6k_to_f16` / `llm_dequant_q5k_to_f16` (element decode mirrors `llm_matvec_q6k`/`q5k`; thread `e` owns weight column `e`).
- `MatMulBatchedGemm` accepts Q6_K/Q5_K (colAlign 256, per-dtype dequant kernel).
- Routed Q6_K/Q5_K → GEMM in **all three** CUDA forward passes:
  - all-GPU dense (`CudaForwardPass.GpuMatMulBatchedCore`)
  - GDN-hybrid (`CudaHybridGdnForwardPass`) — also completed Q8_0/Q4_K → MMQ there (they were on the per-token matvec too)
  - dense-hybrid (`CudaHybridForwardPass`) — attention projections (FFN is per-token by #123 design)
- All **default-on** so users benefit automatically (`SHARPI_GDN_PREFILL_COMPUTE` / `SHARPI_HYBRID_PREFILL_COMPUTE`, `!= "0"`). Argmax-stable not byte-exact → the two GDN/dense-hybrid bit-parity oracles pin the flag **off** class-wide to keep validating the byte-exact matvec batching.

## Results (RTX 4070 Ti)

| model | prefill | note |
|---|---|---|
| Qwen3-8B Q4_K_M | **432 → 2431 t/s @1008 (5.6×)** | VRAM-resident → GPU matmuls are the whole prefill |
| Gemma 4 E4B Q8_0 | 3882 t/s (unchanged) | Q8_0, no regression |
| 27B-MTP (hybrid) | 9.0 → 9.8 t/s (+9%) | doesn't fit 12 GB → CPU-offload-bound; win bounded by VRAM-residency, not dtype coverage |

- matmul time 4972 → 627 ms at N=2290
- decode unchanged (BW-bound, path untouched)
- **MTP draft acceptance unchanged: 92% == 92%** with full routing default-on (and 95%==95% on the Q6/Q5-only check); decode/MTP-verify stay byte-exact

## Validation

- `CudaGemmQ6KTests` / `CudaGemmQ5KTests` track the fp32 CPU reference (`SimdKernels.DotQ6K`/`DotQ5K`) with **0 mismatches** across shapes.
- GDN + dense-hybrid bit-parity oracles green (gating keeps the byte-exact matvec path).
- Qwen3/Gemma4 batched-prefill oracles + MMQ/SoA suites green; full build clean (TreatWarningsAsErrors).
- Output coherent.


pekkah:
Review-cycle resolution:
- **BitConverter.GetBytes(Half)** — addressed in 917dc78: switched the test helpers to `BitConverter.HalfToUInt16Bits(h)`. (Note it was not actually a compile error — the code built and the Q6_K/Q5_K parity tests ran and produced byte-identical results before/after, so .NET 10's `GetBytes(Half)` was already returning the correct fp16 bytes — but the explicit API is clearer.)
- **Static-flag race / test flakiness** — non-issue for this repo: `xunit.runner.json` sets `parallelizeTestCollections: false` and `parallelizeAssembly: false`, so test classes never execute concurrently. The save/restore-in-Dispose pattern matches every existing test in these files that toggles `BatchedPrefillEnabled`.
- **Coverage** — added `BatchedPrefill_ComputeRouting_ArgmaxStableVsMatvec_Coder` so the default-ON routing is asserted argmax-stable at model scale (greedy=362, maxAbs=1.23), complementing the kernel-isolation parity tests.

Issue #156, opened 2026-06-06
perf(cuda): generalize Gemma 4 batched prefill + flash attn + decode graphs to dense Q4_K models


pekkah:
## Goal

Bring the Gemma 4 CUDA performance work to other **dense models that fully offload** via `CudaForwardPass` — primarily **Qwen3-8B Q4_K_M**. The batched-prefill orchestration and CUDA-graph machinery are already largely generic; they are just structurally gated behind `_isGemma4Like`. This issue generalizes them.

Out of scope: MoE/GDN models (separate `CudaHybridForwardPass` / `CudaHybridGdnForwardPass`), and Vulkan.

## Background

What the Gemma 4 work (#136/#141/#142/#146/#147/#149/#154) splits into:

- **Already general** — Q4_K decode already uses dp4a (the #142 work brought Q8_0 *up to* Q4_K); the #154 sampler fix is model-agnostic and already benefits everyone.
- **Q8_0-specific** — int8 MMQ prefill (#141) + SoA repack (#149) are tied to the Q8_0 block layout.
- **Quant-agnostic but Gemma-gated** — batched-trunk prefill (#136), flash-attention prefill (#141/#146/#147), decode CUDA graphs (#136). **This is the opportunity.**

Two config knobs that looked like blockers are already first-class: `attnScale` (kernels do `<=0 => 1/sqrt(headDim)`, Gemma hardcodes `1f`) and `hp.FfnActivation` (`Silu` vs `GeluApprox`, batched body hardcodes `GeluTanhMul`).

## Work Item A — Generalize batched prefill + flash attention (biggest win, low risk)

Three hard Gemma assumptions in `GpuLayerBatchedTrunkGemma4` (`CudaForwardPass.cs:2011`):

- [ ] A1. `int layerHd = _hp.LayerHeadDim![layer]` (`:2013`) NPEs on non-Gemma → `_hp.LayerHeadDim?[layer] ?? _headDim`
- [ ] A2. `_gpu.GeluTanhMul(...)` (`:2103`) hardcoded → dispatch on `_hp.FfnActivation` (`SiLuMul` for `Silu`)
- [ ] A3. `attnScale: 1f` on all 4 attention calls (`:2077-2090`) → cached field (`1f` if `_isGemma4Like` else `-1f`)
[...]

PR #171, opened 2026-06-07
fix(cuda,vulkan): apply weighted QK-norm before RoPE on dense paths (#157)

### LINKED ISSUE #157: investigate: dense Forward applies RoPE before QK-norm (Qwen3 expects norm before RoPE)
## Summary

While generalizing the batched-trunk prefill to dense models (#156), I found that the **dense per-token `CudaForwardPass.Forward` applies RoPE _before_ per-head QK-norm**, whereas the Gemma path (`RunGemma4DeviceRegion`) applies **QK-norm before RoPE**.

```
// Forward (dense, ~CudaForwardPass.cs:940-958)
RoPE(q); RoPE(k);                       // RoPE first
if (_hasQkNorm && !UseL2QkNorm) { HeadNorm(q); HeadNorm(k); }   // then QK-norm

// RunGemma4DeviceRegion (~:1199-1223)
HeadNormQk(q,k);                        // QK-norm first
RoPE(q); RoPE(k);                       // then RoPE
```

RoPE does **not** commute with per-channel-weighted RMSNorm (RoPE preserves the per-pair L2 norm so the RMS scale is unchanged, but it mixes channels `i` and `i+d/2`, which carry different learned weights `w_i ≠ w_{i+d/2}`). The two orders produce materially different Q/K — in #156 the order mismatch alone caused a ~9-logit divergence on Qwen3-8B.

## Why this matters

The HF Qwen3 reference (and llama.cpp `build_qwen3`) apply **q_norm/k_norm before RoPE**. The dense per-token path here does the opposite. If that's a genuine deviation, dense Qwen3 (and any other weighted-QK-norm model: Qwen3-MoE, etc.) decode/prefill output may be subtly off vs the reference — even though it produces coherent text.

In #156 the batched path was deliberately made to **match the existing per-token oracle** (rope→norm for dense), so #156 introduces no behavior change — it faithfully reproduces whatever `Forward` already does. This issue is about whether `Forward` itself is correct.

## Investigation

- [ ] Confirm the actual order in HF `Qwen3Attention.forward` and llama.cpp `build_qwen3` (norm-then-rope expected).
- [ ] Check whether sharpi's top-1 parity claim vs llama.cpp b8585 on Qwen3-8B (README line ~19) was measured against this path — if it passed, reconcile how (e.g. weights near-uniform? parity tolerance? different code path?).
- [ ] Decide: if `Forward` is wrong, swap to norm→rope and re-validate parity; then update the #156 batched body to match (drop the per-family ordering branch — both would become norm→rope).
- [ ] Check the Vulkan `GpuForwardPass` and CPU `ForwardPass` for the same ordering, so a fix stays consistent across backends.

## Scope

Affects weighted (non-L2) QK-norm models on the dense paths. Gemma (norm→rope) and L2-QK-norm models are already correct. Not introduced by #156 — pre-existing.

Refs: #156 (where this was surfaced), branch `perf/cuda-dense-q4k-batched-prefill-156`.

pekkah:
Fixes #157.

## The bug
The dense GPU forward passes applied per-head **QK-norm _after_ RoPE** for weighted-QK-norm models (Qwen3, OLMoE) — the inverse of HF Qwen3, llama.cpp `build_qwen3`, and sharpi's own CPU `ForwardPass` (already correct). RoPE does **not** commute with per-channel-weighted RMSNorm: NEOX RoPE preserves each pair's L2 (RMS scale unchanged) but rotates the `(i, i+d/2)` pairs that `q_norm`/`k_norm` scale per-channel, so `RoPE(diag(w)·x) ≠ diag(w)·RoPE(x)` unless `w[i]==w[i+d/2]`. Qwen3 = `HasQkNorm && !UseL2QkNorm && IsNeoxRope` → exactly the affected branch. The flipped order silently degraded Qwen3 output on every GPU path.

## Why it stayed hidden
The CUDA-vs-CUDA batched-prefill oracles can't catch it: #156 deliberately built the batched path to **match the per-token loop**, so both were equally wrong and agreed. The CPU path was the only independent reference. The README's "byte-identical 60-token greedy decode vs llama.cpp" is a **top-1 token-sequence** claim — robust to sub-argmax logit perturbation on clear-margin prompts — so it never tripped on the order bug.

## Numeric proof (new cross-backend oracle)
`Qwen3_8B_CudaForward_MatchesCpu_QkNormBeforeRope` (CPU↔CUDA, Qwen3-8B-Q4_K_M, identical prompt), demonstrated by stashing the fix:
- **Pre-fix:** `maxAbs = 9.21` logits, top-5 overlap **3/5** → FAIL (matches the ~9-logit divergence #156 noted).
- **Post-fix:** argmax match, top-5 ≥ 4/5, `maxAbs < 4` → PASS.

## Fix (all dense paths; GDN + Gemma were already norm→rope, untouched)
- `CudaForwardPass`: `RunDeviceRegion`, `ForwardProfiled`, `GpuLayerBatchedTrunk` (dropped the per-family `_isGemma4Like` branch — both families are weighted → norm-first).
- `CudaHybridForwardPass`: per-token + `GpuLayerBatchedTrunk`.
- `GpuForwardPass` (Vulkan dense), `HybridForwardPass` (Vulkan dense-hybrid).
- **L2 QK-norm (Llama-4) unchanged** — stays norm-AFTER-RoPE (RoPE layers only); the batched trunks already gate L2 out via `IsBatchedPrefillSupported`.

## Validation
- New cross-backend oracle (above).
- Qwen3 batched-prefill **7/7** (incl. the bit-exact FlashOff one — batched/sequential stayed consistent since both were reordered together).
- Gemma4 CUDA **21/21**, dense-hybrid **5/5**, CUDA + Vulkan SnapKV forward **6/6**.
- One dense-hybrid oracle (`BatchedPrefill_ComputeRouting_ArgmaxStableVsMatvec_Coder`) relaxed from strict greedy-argmax to top-5-overlap + `maxAbs`: the _correct_ output now lands on a **0.27-logit final-token tie** (362 @ 16.67 vs 4220 @ 16.40) that the argmax-stable fp16 GEMM tie-breaks differently than the byte-exact matvec — the same top-5 contract the sibling Qwen3 compute-routing oracles use.
- Qwen3-8B CUDA generation coherent end-to-end; decode **75.3 t/s** (no regression). Full solution builds with **0 warnings**.

### DIFF
--- a/src/SharpInference.Engine/CudaForwardPass.cs
+++ b/src/SharpInference.Engine/CudaForwardPass.cs
@@ -1029,24 +1029,29 @@ public sealed unsafe class CudaForwardPass : IForwardPass
 
             bool useRoPE = _hp.NoRopeLayerStep == 0
                 || (layer + 1) % _hp.NoRopeLayerStep != 0;
+
+            // Order matters: RoPE does NOT commute with per-channel-weighted QK-norm
+            // (NEOX RoPE mixes channels i and i+d/2, which carry different learned
+            // weights), so we mirror the CPU ForwardPass / HF Qwen3 / llama.cpp
+            // build_qwen3 ordering exactly (issue #157):
+            //   • weighted QK-norm (Qwen3, OLMoE, …): norm BEFORE RoPE
+            //   • L2 QK-norm (Llama-4):               norm AFTER  RoPE (RoPE layers only)
+            if (_hasQkNorm && !_hp.UseL2QkNorm)
+            {
+                _gpu.HeadNorm(_q, _wqNorm![layer], _numHeads, _headDim, _hp.RmsNormEps, _hp.IsPerChannelQkNorm);
+                _gpu.HeadNorm(_k, _wkNorm![layer], _numKvHeads, _headDim, _hp.RmsNormEps, _hp.IsPerChannelQkNorm);
+            }
+
             if (useRoPE)
             {
                 _gpu.RoPE(_q, position, _headDim, _hp.RopeTheta, _hp.IsNeoxRope);
                 _gpu.RoPE(_k, position, _headDim, _hp.RopeTheta, _hp.IsNeoxRope);
             }
 
-            if (_hasQkNorm && (_hp.UseL2QkNorm ? useRoPE : true))
+            if (_hasQkNorm && _hp.UseL2QkNorm && useRoPE)
             {
-                if (_hp.UseL2QkNorm)
-                {
-                    _gpu.HeadNormPure(_q, _numHeads, _headDim, _hp.RmsNormEps);
-                    _gpu.HeadNormPure(_k, _numKvHeads, _headDim, _hp.RmsNormEps);
-                }
-                else
-                {
-                    _gpu.HeadNorm(_q, _wqNorm![layer], _numHeads, _headDim, _hp.RmsNormEps, _hp.IsPerChannelQkNorm);
-                    _gpu.HeadNorm(_k, _wkNorm![layer], _numKvHeads, _headDim, _hp.RmsNormEps, _hp.IsPerChannelQkNorm);
-                }
+                _gpu.HeadNormPure(_q, _numHeads, _headDim, _hp.RmsNormEps);
+                _gpu.HeadNormPure(_k, _numKvHeads, _headDim, _hp.RmsNormEps);
             }
 
             // SnapKV (issue #59): capture the post-RoPE / post-Q-norm query for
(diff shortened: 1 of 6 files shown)


pekkah:
Addressed the Gemini review: added the explicit `!_hp.UseL2QkNorm` guard at `CudaHybridForwardPass.cs:1255` (commit 379d7d5), matching `CudaForwardPass`'s `ApplyQkNormBatched` pattern so `_gpuQNorm`/`_gpuKNorm` (null for L2 models) can't dereference even if the upstream `IsBatchedPrefillSupported` gate ever changes. Engine builds with 0 warnings; no behavior change (L2 models never reach the batched trunk today).

Issue #179, opened 2026-06-09
Quantized (fp16/q8_0) KV cache for the dense CUDA path → unlock long context (Gemma 4 12B first)


pekkah:
## Problem

The dense CUDA path stores the KV cache in **fp32** (`CudaForwardPass.cs:594-595` allocates `_gpuKCache`/`_gpuVCache` as fp32; `GpuForwardPass.cs:199` — "K + V, fp32"). `KvAppend`, `Attention`, and `AttentionSwa` all read/write fp32 buffers with `sizeof(float)` strides. This applies to **every dense model on the CUDA path**, not just Gemma — fp32 KV is the single biggest barrier to long context on a 12 GB card.

`SHARPI_KV_DTYPE` already exists but only on the GDN hybrid path (`CudaHybridGdnForwardPass.cs:672`) and only supports `fp32 | bf16` — it is not wired to the dense path and has no q8_0 mode. llama.cpp serves these models at 128K on the same class of card only because `--cache-type-k/v q8_0` shrinks KV ~4×.

## This is a general backend enhancement

Quantized KV storage + quant-aware append/attention kernels are a **dense-CUDA-path capability**, not a model feature. Every dense model benefits:

- **Qwen3-8B Q4_K_M** — KV-bound at long context just like Gemma; gets the same ~4× context headroom.
- **Gemma 4 12B QAT** — the *driving use case* (see VRAM math) and the *validation target* (its SWA ring was never exercised past `-c 2048`).
- Any future dense model on the CUDA path inherits the knob for free.

Two Gemma-specific behaviors must be **preserved** (constraints, not new work):
- **SWA ring** (`SwaRingSize`/`SwaRingHeadroom`) — quantized kernels must keep honoring the modulo wrap. Non-SWA models (Qwen3) use the full-context branch unchanged.
- **`attention_k_eq_v`** (12B global layers, K=V, store K only) — V's dequant-on-read reads the same quantized K blocks.

## VRAM math — the driving case (Gemma 4 12B QAT, 12 GB card, 128K ctx)

48 layers, head_dim 256, ~40 SWA + 8 global; global layers MQA with `attention_k_eq_v` (K=V, store K only). SWA ring = window + `SwaRingHeadroom` (4096), capped at ctx.

| Component | fp32 (today) | q8_0 |
|---|---:|---:|
| Model (Q4_0 bulk + Q6_K tied embd) | ~7.0 GB | ~7.0 GB |
[...]

Issue #178, opened 2026-06-09
GPU draft-MTP speculative decoding for Gemma 4 12B (decode 54 → ~70 t/s)


pekkah:
> ## Status (2026-06-14): unblocked
> The hard blocker — **#179 (fp16/q8_0 KV for the dense Gemma CUDA path)** — **shipped**, so a draft model now fits alongside a quantized KV cache at usable context. CUDA single-user batched-verify infrastructure also landed since (#207/#208: `BatchVerify` on `BatchForwardMulti`, n-gram `--draft-lookup`, ~1.33× at k=4). Remaining for this issue: drive an **external** GPU draft-MTP model for Gemma 4 12B and wire **N>1** sampled (temp>0) accept. Still a good-case-only ~1.3× win, so deprioritized — but no longer blocked.
>
> ---

## Motivation

llama.cpp users report **~70 decode t/s** on Gemma 4 12B-it QAT on 12 GB cards using MTP speculative decoding with an external "assistant-MTP" draft model:

```
llama-server -m gemma-4-12B-it-qat-UD-Q4_K_XL.gguf \
  -md gemma-4-12B-it-qat-assistant-MTP-Q8_0.gguf \
  -ngl 99 --spec-draft-ngl 99 --spec-type draft-mtp --spec-draft-n-max 4 \
  --cache-type-k q8_0 --cache-type-v q8_0 ... --temp 0.2
```

Our recorded raw decode is **54.1 t/s** (README text-gen table, `CUDA -g -1`), already within ~6% of llama.cpp's *non-speculative* 57 t/s. The 54 → 70 gap is **entirely** the MTP draft speculation (draft proposes up to 4 tokens, the 12B verifies them in one batched pass; their temp 0.2 gives high acceptance).

## Why this is a follow-on, not a first move

- The 54 → 70 win is ~1.3× and **good-case only** (low temp / high acceptance). Both rates are far above reading speed, so the benefit is real only for long generations / agentic loops, not interactive chat.
- **It does not fit in VRAM without KV quantization first.** Our dense Gemma CUDA KV cache is fp32 (`GpuForwardPass.cs:199`, "K + V, fp32"), so the 12B already nearly fills a 12 GB card at long context with *no* draft model (that's why the README row is measured at `-c 2048`). A draft model only fits once KV is quantized. **Blocked on #179 (fp16/q8_0 KV for the dense Gemma CUDA path).**

So: land KV quantization first, then this becomes a purely additive throughput feature.

[...]

PR #184, opened 2026-06-09
bf16 + q8_0 KV cache for the dense CUDA path → 128K on Gemma 4 12B in 12 GB (#179)

pekkah:
Adds opt-in narrowed KV-cache dtypes (`--kv-type bf16|q8_0` / `SHARPI_KV_DTYPE`) to the dense CUDA forward pass, plus a host-allocation fix that was the real context ceiling. Takes the Gemma 4 12B QAT (`-g -1`) from a ~32K ceiling to **`-c 131072` (128K) within 12 GB**, coherent. Default KV stays fp32; both narrowed dtypes are opt-in.

## Result (Gemma 4 12B QAT Q4_0, RTX 4070 Ti 12 GB, `-g -1`)

| KV dtype | Max ctx in 12 GB | Decode @128K |
|---|---|---|
| fp32 (default) | ~32K (`cudaMalloc`-fails at 64K) | — |
| bf16 | **128K** | 18.8 t/s (embed-table cliff) |
| **q8_0** | **128K** | **53.3 t/s — full speed (2.8× over bf16)** |

At 128K the tied ~818 MB Q6_K embed/LM-head table can't stay GPU-resident under bf16's KV (~350 MB free) and spills to host per token → decode cliffs. q8_0's block-quantized cache (~¼ fp32, ~½ bf16) frees ~1.1 GB more, keeps the table resident, and holds full speed. Same coherent output across all three dtypes.

## What's in here (squashes #179 increments 1 → 2)

- **Increment 1** — bf16 KV storage + SWA/global bf16 kernels, per-token decode/prefill dispatch, `--kv-type` CLI flag (rejects bf16+TQ / explicit-SnapKV).
- **Increment 1.5a/b** — bf16 scalar batched prefill (≤4096) + bf16 Tc2 tensor-core flash prefill incl. chunked >4096.
- **Host-cache fix** — `KvCache.CreateBookkeepingOnly`: the GPU passes kept the real KV in VRAM but still allocated a full *host* KV buffer used only for the position counter (~50 GB at 64K on the 12B → managed `OutOfMemoryException`). This was the hidden ceiling, masking the bf16 win.
- **Increment 2 — q8_0 KV** (this PR's headline). ggml `block_q8_0` layout (32 int8 + fp16 scale = 34 B/block). One `sharpi_kvload(const block_q8_0*)` overload is the single variation point; the 6 bf16 read kernels were **templated on `<KV>`** with bf16/q8_0 thunks (bf16 thunk byte-identical, since `sharpi_kvload(unsigned short) == sharpi_bf16_to_fp32`). New store kernels quantize per 32-lane warp (`amax → d=amax/127 → rintf` clamp ±127, mirroring `llm_quantize_q8_1`). `CudaBackend.Allocate` now sizes via `DTypeInfo.ByteSize` (identical for scalar, adds block-quantized).

## Validation

- **Parity oracles** (logit-level top-5 argmax-stable vs an independent fp32 trajectory, teacher-forced so KV dtype is the only variable, per `feedback_cross_backend_parity_test`):
  - bf16: per-token + batched (Qwen3-8B, E4B) + chunked (E4B) — all green, preserved through the templating.
  - q8_0: per-token + batched (Qwen3-8B, E4B, 12B) + chunked >4096 (E4B **and 12B/k_eq_v**) — all green.
  - `DTypeInfo.ByteSize(.., Q8_0)==34` CPU unit assert.
- **CLI A/B** at `-c 131072` on the 12B: q8_0 53.3 t/s vs bf16 18.8 t/s, both coherent.

## Review (pr-review-toolkit)

code-reviewer: no critical/important issues. Addressed: a fail-loud guard for the q8_0 `kvDim % 32` invariant (was a latent silent under-allocation; all current models satisfy it), the 12B chunked test above, the ByteSize unit test, and 6 stale kernel-header comments. Knowingly deferred: the `_ => fp32` dtype-selector fall-through (future-proofing only — `_kvDType` is centrally validated to {fp32,bf16,q8_0}) and a pre-existing `EnsureImageKernels` logging nit.

## Notes / follow-ups

- Non-SWA scalar batched q8 kernels + the wave (>4096 global) q8 kernel are wired but only reachable by a non-`%64` head_dim model (none on disk) — same latent-but-unexercised status as their bf16 twins.
- Open follow-ups: auto-narrow default when ctx won't fit fp32; Tc/half2-flash q8 thunks; q8_0 draft-model headroom (#178).


pekkah:
Added the narrowed-KV option to the **API server** too (`4bdbd83`), so it isn't CLI-only:

- `SharpInferenceServerOptions.KvType` (`fp32`/`bf16`/`q8_0`) — bindable from `appsettings.json` (`SharpInference:KvType`), inline `Configure`, or the `SHARPI_KV_DTYPE` env override (mirrors the `SHARPI_MODEL`/`SHARPI_BACKEND` pattern).
- `InferenceEngineLoader` forwards it to `SHARPI_KV_DTYPE` before building the forward pass (only when explicitly set, so an externally-set env var still works); validated at model load.
- Has effect only on the CUDA dense path; ignored by CPU/Vulkan and rejected with TurboQuant / explicit SnapKV — same semantics as the CLI `--kv-type`.
- Tests: `KvType` default + appsettings binding (`q8_0`).

pekkah:
Good catch on the subnormal-scale edge — applied in `003bacf`. A sub-block whose `amax` is subnormal makes `d=amax/127` subnormal, `1/d` overflows to +inf, and a zero lane's `0*inf = NaN → (int)NaN` clamps to -127. Switched to the `d < 1e-30f` threshold as suggested. It's far below any real KV scale (`d ≈ amax/127 ≈ 1e-3..1e-1`), so real blocks are unaffected and the parity oracles are unchanged; an all-near-zero block now correctly quantizes to all-zeros.

Issue #185, opened 2026-06-09
q8_0/bf16 KV follow-ups: auto-narrow default + Tc/half2-flash q8 thunks (#179)


pekkah:
Follow-ups deferred from #179 / #184 (bf16 + q8_0 KV cache for the dense CUDA path). Neither blocks anything today; both are opt-in/perf niceties.

## 1. Auto-narrow KV default when the requested context won't fit at fp32
Today `--kv-type` defaults to `fp32`; an operator must opt into `bf16`/`q8_0` to reach long context, and an oversized `-c` at fp32 fails with a `cudaMalloc` error instead of degrading gracefully. Mirror the existing **auto-SnapKV VRAM heuristic**: when the requested context's fp32 KV won't fit the VRAM budget, automatically select a narrowed dtype (prefer bf16, then q8_0) and log the choice, rather than erroring. Safe now that narrowed-KV prefill is full-speed.

- Hook near the `CudaForwardPass` ctx/VRAM sizing (where SnapKV auto-budget is computed).
- Keep `--kv-type` an explicit override (an explicit `fp32` should still error loudly rather than silently narrow).
- Decide precedence vs auto-SnapKV (both chase the same VRAM win).

## 2. Tc / half2-flash q8_0 (and bf16) thunks
The narrowed-KV batched/chunked prefill path uses the **Tc2** tensor-core flash kernel, which requires `head_dim % 64 == 0`. Models with a non-`%64` head_dim past 4096 fall back to the per-token loop (correct, but slower) because the single-warp **Tc** (`llm_flash_attn_prefill_tc`, #146) and **half2** (`llm_flash_attn_prefill_f32`, #141) flash kernels have no bf16/q8_0 thunk yet — only Tc2 was templated on `<KV>`.

- Template those two kernels on `<KV>` (same `sharpi_kvload` pattern as Tc2) and add bf16/q8_0 thunks; wire them into the narrowed dispatch's non-`%64` branch + `NarrowedFlashTc2CoversAllLayers` analogue so such models can chunk past 4096.
- **Blocker:** no non-`%64` head_dim dense model is on disk to validate against — needs a test model before this is worth doing. Until then the per-token fallback covers correctness.

## Also worth a look
[...]

PR #186, opened 2026-06-09
feat(cuda): auto-narrow KV dtype when fp32 won't fit VRAM (#185)

pekkah:
Closes the **auto-narrow** half of #185 (item 1). Follow-up to #179/#184 (bf16 + q8_0 KV cache for the dense CUDA path).

## Problem
`--kv-type` defaulted to `fp32`; an oversized `-c` died at construction with `cudaMalloc failed: 2` — capping context at fp32 even though a narrowed store would fit. Operators had to know to pass `--kv-type bf16`/`q8_0` to reach long context.

## Change
`CudaForwardPass` now mirrors the auto-SnapKV VRAM heuristic. When the operator set **neither** an explicit `--kv-type`/`SHARPI_KV_DTYPE` **nor** an explicit SnapKV budget, TQ is off, and the resolved context's fp32 KV cache won't fit the VRAM budget, it auto-selects a narrowed dtype (**bf16 preferred; q8_0** if bf16 still won't fit and the geometry supports it) and logs the choice instead of erroring.

### Precedence vs auto-SnapKV (documented in code)
Auto-narrow runs **first and wins**. SnapKV does *not* shrink the construction-time allocation (the cache is allocated full-`maxSeqLen` up front; eviction only bounds the *logical* length at runtime), so it cannot avert the `cudaMalloc` failure — only narrowing the element width can. Narrowing is also exact-context, argmax-stable, and full-speed-prefill, whereas SnapKV evicts tokens (lossy).

Explicit choices are **never overridden**: an explicit `fp32` still errors loudly at allocation rather than silently narrowing; an explicit SnapKV budget is honoured (auto-narrow stands down) even though it can't avert the alloc failure — the operator's call.

## Implementation
- Decision factored into pure helpers — `ResolveKvDType` / `EstimateKvCacheBytes` / `Q8KvGeometrySupported` / `EstimateAvailableKvVram` (extracted from `EstimateMaxContext`, single-sourcing the VRAM budget so estimator and heuristic stay in agreement). All unit-testable without a GPU or model — **9 new tests**.
- `EstimateKvCacheBytes` rounds each K/V buffer up to the next power of two to match the GPU buffer pool (`gpu.Allocate`, `exact:false`). A raw byte sum undercounts by up to ~2× per buffer (q8_0's 34-byte blocks rarely land on a power of two) and could wrongly conclude fp32 fits, defeating the narrow. (Caught in code review.)
- **Server:** `SharpInferenceServerOptions.KvType → SHARPI_KV_DTYPE` is already forwarded; no `InferenceEngineLoader` change needed.

## Validation
- Filter `Gemma4CudaBatchedPrefill|Qwen3CudaBatchedPrefill|KvDtype`: **39/39 green** (40m12s) — 9 new pure tests + all bf16/q8_0 parity, batched-prefill, and >4096 chunked-prefill cases.
- **E2E:** `gemma-4-12b -g -1 -c 131072` with no `--kv-type` (previously `cudaMalloc`-fails at fp32) now logs `KV auto-narrowed to Q8_0 for context 131072` and decodes coherently — *"The capital of France is **Paris**."* — at full speed (48.5 t/s, no cliff).
- pr-review-toolkit: code-reviewer (1 Important fixed: pool-rounding undercount; + Minor/Nit) and code-simplifier (no changes) cycles run.

## Not in this PR — #185 item 2 (Tc/half2 q8 flash thunks)
Remains blocked: every dense GGUF on disk has head_dim divisible by 64 (Qwen3=128, OLMoE=128, SmolLM=64, Qwen3.6=256, Gemma4=512). With no non-`%64` head_dim model to validate against, templating the single-warp Tc / half2 flash kernels on `<KV>` would ship untested kernels. The per-token fallback covers correctness until a validation model exists.

### DIFF
--- a/src/SharpInference.Engine/CudaForwardPass.cs
+++ b/src/SharpInference.Engine/CudaForwardPass.cs
@@ -529,14 +529,17 @@ public sealed unsafe class CudaForwardPass : IForwardPass
         // prevent the cudaMalloc failure — only narrowing the element width can. Narrowing
         // is also exact-context, argmax-stable, and full-speed-prefill, whereas SnapKV
         // evicts tokens (lossy). We therefore auto-narrow only when the operator set
-        // NEITHER an explicit --kv-type NOR an explicit SnapKV budget; either explicit
-        // choice is respected and never overridden (an explicit fp32 still errors loudly at
-        // allocation rather than silently narrowing; an explicit SnapKV budget is honoured
-        // even though it can't avert the alloc failure — the operator's call). When
-        // auto-narrow fires it sets a narrowed _kvDType, which flips kvNarrowed below and
-        // disables auto-SnapKV.
+        // NEITHER an explicit --kv-type NOR an explicit *positive* SnapKV budget; either
+        // explicit choice is respected and never overridden (an explicit fp32 still errors
+        // loudly at allocation rather than silently narrowing; an explicit positive SnapKV
+        // budget is honoured even though it can't avert the alloc failure — the operator's
+        // call). Note SHARPI_SNAPKV_BUDGET=0 means "disable SnapKV" (IsBudgetExplicit=true,
+        // Budget=0), the same disable knob the banners advertise — it must NOT suppress
+        // auto-narrow, so we gate on Budget > 0 to match the disable semantics used by the
+        // SnapKV throws/auto-enable below. When auto-narrow fires it sets a narrowed
+        // _kvDType, which flips kvNarrowed below and disables auto-SnapKV.
         if (!_tqEnabled && !kvDTypeExplicit && _kvDType == DType.Float32
-            && !_snapKvCfg.IsBudgetExplicit)
+            && !(_snapKvCfg.IsBudgetExplicit && _snapKvCfg.Budget > 0))
         {
             long availKvBytes = EstimateAvailableKvVram(model, gpu, hp);
             long fp32KvBytes  = EstimateKvCacheBytes(hp, _maxSeqLen, DType.Float32);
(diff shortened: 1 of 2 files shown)


pekkah:
Addressed the review round in e7dbf05:

**Gemini Code Assist (critical):** confirmed and fixed — `EstimateGpuTensorBytes` was missing `DType.Q4_0` from its raw-upload set, counting Q4_0 weights (the Gemma 4 12B QAT dtype) as fp32. That ~7× over-count drove the KV budget to its 64 MiB floor, so auto-narrow was deciding on a degenerate budget. The set now matches `UploadWeight`'s raw branch exactly (`{Float32, Q4_0, Q4_K, Q6_K, Q8_0}`); Q5_K and others correctly stay fp32-counted since `UploadWeight` CPU-dequantizes them. E2E confirms the 12B budget went 64 MiB → ~1549 MiB (realistic).

**silent-failure-hunter (important):** `SHARPI_SNAPKV_BUDGET=0` (the "disable SnapKV" knob the banners advertise) was suppressing auto-narrow because it sets `IsBudgetExplicit=true`. Now gated on `(IsBudgetExplicit && Budget > 0)`, matching the disable semantics used by the SnapKV throws/auto-enable — so disabling SnapKV no longer silently reintroduces the `cudaMalloc` failure.

**pr-test-analyzer (coverage):** added the gemma4-shaped `EstimateKvCacheBytes` test (KvSourceLayer alias-skip + SWA-ring cap + per-layer head_dim/kv-heads, hand-computed independently of the implementation), a mixed-layer `Q8KvGeometrySupported` test (any single %32 violator fails; an aliased violator is skipped), and a `ResolveKvDType` inclusive-fit boundary test.

Validation: 12/12 pure tests; Qwen3-8B bf16/q8 parity 3/3; E2E 12B `-c 131072` → q8_0 @ 47.4 t/s, `-c 8192` → q8_0 (fp32 5376 MiB still exceeds the 1549 budget — correct: the 4096 SWA-ring headroom keeps 12B KV large even at moderate context). The full `Gemma4CudaBatchedPrefill|Qwen3CudaBatchedPrefill|KvDtype` filter (39/39) was green before this round; these fixes don't touch the parity path (explicit ctx + explicit dtype bypass both `EstimateMaxContext` and the auto-narrow block).

Issue #166, opened 2026-06-07
SWA KV ring: bf16-append wrap + CudaHybridForwardPass batched path are latent-only / untested


pekkah:
Tracking the two latent traps the #164 review surfaced (commit 48a5622). Both are correct/inert TODAY; this is so they aren't forgotten if those paths ever activate.

**1. bf16 append ring wrap is untested (identity-only today).** `llm_kv_append_bf16` / `llm_kv_append_batched_bf16` got the `pos % max_seq_len` modulo for symmetry with the f32 kernels, but the only consumer of the bf16 KV cache is the full-context GDN-hybrid path, where `pos < max_seq_len` makes the modulo the identity. If a future **windowed** model ever uses the bf16 KV cache, the wrap path would be exercised for the first time with no test. Add a synthetic bf16 ring-wrap bit-wise test (mirror the f32 `CudaGdnBatchedTrunkTests` append tests) when that happens.

**2. CudaHybridForwardPass SWA ring is bare-window (decode-only).** It deliberately sizes SWA caches at `min(ctx, window)` (not `SwaRingSize` = window+headroom), which is correct because Gemma 4 there is decode-only (`IsBatchedPrefillSupported` requires `!_isGemma4Like`) and per-token decode only needs ring ≥ window. If batched/chunked prefill is ever enabled for Gemma 4 in the hybrid path, the bare-window ring would overwrite a still-needed window — switch it to `SwaRingSize` at that point (there's a code comment flagging this).

**3. Pure-decode multi-wrap is untested** — generating >4608 tokens to wrap the decode ring is impractically slow via the per-token oracle; the prefill path now covers wrapped reads observably (#164), so this is low value.

No action needed now; reference if touching SWA / bf16-KV / hybrid-Gemma paths.

PR #188, opened 2026-06-10
GPU saturation gaps: chunked+packed prefill admission, KV budget, PCIe probe (#183)

### LINKED ISSUE #183: GPU saturation gaps: non-blocking prefill admission, packed multi-seq prefill, VRAM token-budget autotune
## Context

Spun out of reviewing the article ["I Built a C++ Backend So My GPU Would Stop Eating Air"](https://towardsdatascience.com/i-built-a-c-backend-so-my-gpu-would-stop-eating-air/) (write-up of [WarpGroup-Backend](https://github.com/AnubhabBanerjee/WarpGroup-backend)). Its thesis: standard batching pads short sequences with zeros up to the longest in the batch, wasting GPU FLOPs; fix it with padding-free packed batches, FFD bin-packing under a VRAM token budget, pinned-memory async DMA, and a Phase-0 autotune.

**Most of the headline techniques are already in our design and need no work:**

- **Pinned memory + async DMA** — already implemented: `CudaBackend` `cudaMallocHost`s a staging buffer plus a double-buffered async pinned pool with completion events (`src/SharpInference.Cuda/CudaBackend.cs:39-60`) and direct-pinned `Download/UploadAsync` overloads (`:708+`).
- **No padding waste in decode** — `ForwardPass.BatchForwardMulti` (`src/SharpInference.Engine/ForwardPass.cs:2208`) is already ragged: one token per sequence, per-sequence KV cache, weights amortized N×. No rectangle to pad.
- **No padding waste in prefill** — `ContinuousBatchingEngine.AdmitRequest` prefills each prompt individually (`ContinuousBatchingEngine.cs:332`) and `PrefillCore` runs the real token count. We never build the padded matrix the article fights.
- **16-token Tensor-Core micro-pad / FFD packing** — low value for us: cuBLAS pads GEMMs internally and quant block sizes already enforce alignment; FFD only matters once packed prefill exists.

What the article's *deeper* lesson (keep the GPU saturated, never let scheduling idle it) does surface are three real gaps worth tracking:

---

### Gap 1 — Blocking, serialized prefill admission (highest value)

In `ContinuousBatchingEngine.BatcherLoop`, a newly admitted request's prefill runs synchronously and **every already-active sequence stalls** — no decode advances during that prefill (`ContinuousBatchingEngine.cs:171-210`, admission at `:332`). A long prompt freezes the whole decode batch. This is our analog of "the GPU eating air."

**Proposed fix:** chunked / interleaved prefill (Sarathi / vLLM-style) — split a prompt's prefill into chunks and interleave them with decode steps so active sequences keep progressing.

**Scope:** moderate; engine-level scheduling change, no new attention kernel.

### Gap 2 — Packed multi-sequence prefill (direct port of the article)

A varlen `PrefillCore` that prefills several pending prompts in **one** packed forward pass, amortizing weight reads across prompts exactly like `BatchForwardMulti` does for decode.

**Scope:** largest lift — needs a `cu_seqlens`-style packed causal-attention path. Builds naturally on Gap 1.

### Gap 3 — VRAM token-budget autotune + admission backpressure (smallest/safest)

`HardwareProfile` is crude — PCIe bandwidth bucketed by VRAM size, no token budget (`HardwareProfile.cs:39-45`) — and `ContinuousBatchingEngine` admits up to `maxBatchSize` with **no KV-memory backpressure**, so a burst of long prompts can OOM. Add an empirical Phase-0 probe to derive a real token budget and gate admission on it.

**Scope:** small; standalone, no kernel work. Good first step and de-risks Gaps 1–2.

---

## Suggested order

1. Gap 3 (cheap, prevents OOM, unblocks measurement)
2. Gap 1 (biggest throughput win under concurrent load)
3. Gap 2 (largest lift; layers on top of 1)

Each is independently shippable.

pekkah:
Closes #183.

## Gap 1 — chunked / interleaved prefill admission
Admitted prompts prefill in `prefillChunkTokens` slices (default 256; `SHARPI_PREFILL_CHUNK` / `PrefillChunkTokens`; `0` = legacy blocking) with one `BatchForwardMulti` decode step between chunks. Cancellation honoured between chunks; SnapKV-enabled engines fall back to blocking (eviction needs the whole prompt at `startPos==0`).

**A/B** (`SharpInference.Bench -- --cb`, SmolLM2 CPU, 3 active decoders + 1040-token prompt injected):

| | worst decode stall | aggregate |
|---|---|---|
| blocking (pre-#183) | **25.1 s** | 8.1 tok/s |
| chunked+packed (256) | **7.9 s** | 7.1 tok/s |

Smaller chunks are a trap on CPU — every `MatMulBatched` call re-pays weight dequant, so prefill collapses 42→20→13 t/s at chunk 256/64/32. 256 is the default compromise, operator-tunable.

## Gap 2 — packed multi-sequence prefill
`ForwardPass.PrefillPackedMulti(chunks, startPos, caches, wantLogits)`: in-flight prompts' chunks run as ONE packed varlen pass — batched GEMMs over the combined token count, per-token RoPE/append/attention against each sequence's own `PagedKvCache`, no padding, no cross-sequence attention. Logits only for prompt-completing chunks. Same guards as `BatchForwardMulti` (no MoE/TQ/gemma4).

**A/B** (`--cb --packed`, 4×256-token prompts): whole-prompt serial 35.8 tok/s · chunked serial 20.4 · **chunked packed 33.7 (+65% vs serial chunks, −6% vs blocking)**.

## Gap 3 — KV token-budget admission backpressure
Each request reserves `min(prompt + MaxNewTokens, MaxSeqLen)` KV tokens at admission (`ForwardPass.KvBytesPerToken`); the head request waits when the budget would be exceeded instead of risking OOM. A lone oversized request is always admitted (no deadlock). `SHARPI_KV_BUDGET_MB` / `KvBudgetMb`: 0 = auto (half RAM), negative = unlimited.

## HardwareProfile: measured PCIe bandwidth
`CudaBackend.MeasurePcieBandwidthGBps()` (pinned 64 MiB probe, min(H2D, D2H), heuristic fallback) replaces the VRAM-bucket guess in `Detect(CudaBackend)`. 4070 Ti: ~26 GB/s measured vs 20 guessed.

## CUDA 13.3 evaluation (no default change)
All pinned entry points exist in 13.3. Qwen3-8B `-g -1`: prefill 2354→2390 t/s (+1.5%), decode 58.0→57.9 (noise) — decode is HBM-bound in our NVRTC kernels. `CudaLibraryResolver` (assembly's single `DllImportResolver`, NVRTC probing folded in) keeps CUDA 12 default, `SHARPI_CUDA13=1` opts into the 13 pair, and 13-only installs now load instead of hard-failing.

## Test-suite fix found along the way
`Gemma4_12B_CudaForward_ProducesCoherentDecode` has been failing on master at default settings since #186: #185's auto-narrow makes the 12B tests run with bf16 KV on a 12 GB card, and their synthetic OOD prompt's tiny logit margin lets bf16 rounding tip greedy into a one-token attractor (forced bf16 failed identically at #184 — nothing regressed; CLI bf16 output is coherent). The 12B tests now pin `SHARPI_KV_DTYPE=fp32` (they guard trunk math, not dtype noise).

## Tests
5 new in `ContinuousBatchingTests` (chunked parity, packed parity incl. `startPos>0` continuation, engine chunked-vs-unchunked greedy equality, tiny-budget completion). Full suite green: Core 134, ForwardPass 481→ all passing after the 12B pin, Pipeline 37, TurboQuant 41, Server 103.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -73,7 +73,7 @@ Supporting libraries:
 - `PagedKvCache` (in Engine) — lazily allocated paged KV cache used by `ForwardPass`. Pages (16 positions) allocated on first write; `TruncateTo` is a soft operation (enables prefix reuse); `Reset` returns pages to warm pool.
 - `IInferenceEngine` (in Engine) — top-level generation interface used by the server: `GenerateAsync(prompt, sp, ct) → IAsyncEnumerable<string>`. Implemented by `InferenceEngine` (single-user, prefix caching) and `ContinuousBatchingEngine` (multi-user batching, activated via `SHARPI_MAX_BATCH`).
 - `ForwardPass.BatchForwardMulti(tokens[], positions[], caches[])` — batched multi-sequence decode; amortizes weight reads N× across concurrent users. Each sequence has its own `PagedKvCache`. Not supported for MoE or TurboQuant.
-- `ForwardPass.PrefillWithCache(tokens, cache, startPos)` — prefills a per-sequence cache (used by `ContinuousBatchingEngine` during request admission).
+- `ForwardPass.PrefillWithCache(tokens, cache, startPos)` — prefills a per-sequence cache (used by `ContinuousBatchingEngine` during request admission). Admission is chunked (`SHARPI_PREFILL_CHUNK`, default 256 tokens) and interleaved with decode steps; multiple in-flight prompts prefill as one packed pass via `ForwardPass.PrefillPackedMulti` and admission is gated by a KV token budget (`SHARPI_KV_BUDGET_MB`) — issue #183.
 - `ForwardPass` / `GpuForwardPass` in Engine dispatch the transformer layer sequence.
 - Hot paths use `NativeMemory`, `Span<T>`, and Vulkan buffers — no managed heap allocations.
 - Unsafe code is used throughout for performance. `AllowUnsafeBlocks` is enabled globally.
--- a/README.md
+++ b/README.md
@@ -157,6 +157,10 @@ dotnet run --project src/SharpInference.Cli -c Release -- \
   -p "Write a binary search in Rust" --temp 0
 
 # API server (OpenAI /v1/chat/completions + Anthropic /v1/messages, port 5000)
+# Multi-user: SHARPI_MAX_BATCH=8 enables continuous batching (CPU backend). Long-prompt
+# admission prefills in SHARPI_PREFILL_CHUNK-token slices (default 256, 0 = blocking)
+# interleaved with decode, packed across prompts; SHARPI_KV_BUDGET_MB caps total KV
+# memory for admission backpressure (default: half of RAM). See issue #183.
 SHARPI_MODEL=models/SmolLM2-1.7B-Instruct-Q4_K_M.gguf \
   dotnet run --project src/SharpInference.Server.Host -c Release
 ```
--- a/benchmarks/SharpInference.Bench/ContinuousBatchingHarness.cs
+++ b/benchmarks/SharpInference.Bench/ContinuousBatchingHarness.cs
@@ -1,0 +1,220 @@
+using System.Diagnostics;
+using SharpInference.Core;
+using SharpInference.Cpu;
+using SharpInference.Engine;
+
+namespace SharpInference.Bench;
+
+/// <summary>
+/// Manual A/B harness for issue #183 (GPU/CPU saturation under concurrent load):
+/// measures how much a long-prompt admission stalls already-decoding sequences,
+/// with legacy blocking prefill (chunk=0) vs chunked+packed interleaved prefill.
+///
+/// Scenario: 3 interactive requests decode continuously; once they are all streaming,
+/// a long-prompt request is injected. Reported per mode:
+///   - per-interactive-sequence max / p95 inter-token gap (the "stall")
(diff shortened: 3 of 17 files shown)


pekkah:
Addressed all three review findings in 2bf8f98:

1. **Gemma 4 + continuous batching** — gated in `BuildForwardPass` (`batchOk` now requires `hp.LayerHeadDim is null`), falling back to the single-user `InferenceEngine`. Note the failure mode was a clean per-request `NotSupportedException` (prefill exceptions are caught and completed onto the request channel), not a batcher hang — but every request failing made the engine useless for gemma4, so the gate is the right call.
2. **Batcher loop resilience** — `BatcherLoop` is now try/catch/finally: an exception the per-request handlers didn't isolate fails every in-flight request with that exception via the finally drain (no hung callers, no leaked caches).
3. **Eager dual-version DLL loading** — `DecideRuntimeMajor` probes the preferred pair first, only falls back when absent, and frees probe handles, so the unused CUDA major is never left mapped.

Follow-ups filed from this work: #189 (dequant-once MatMulBatched cache → smaller prefill chunks), #190 (continuous batching on the CUDA backend), #191 (narrowed-KV decode-coherence coverage, refs #166).

Issue #191, opened 2026-06-10
test(cuda): narrowed-KV (bf16/q8_0) decode-coherence coverage with template-correct prompts (refs #166)


pekkah:
## Context

While landing PR #188, `Gemma4_12B_CudaForward_ProducesCoherentDecode` turned out to be failing on master at default settings since #186: the #185 KV auto-narrow switched the 12B tests to bf16 KV on a 12 GB card, and their synthetic OOD token prompt has a tiny top-logit margin where bf16 rounding tips greedy decode into a single-token repetition attractor ([236771 ×6]). Forced `SHARPI_KV_DTYPE=bf16` failed identically back at #184 — nothing regressed; the kernels are fine (CLI bf16 output on a real prompt is coherent, q8_0 and fp32 pass). PR #188 pins those tests to fp32, restoring determinism.

## Gap

That pin means there is now **zero automated decode-coherence coverage for the narrowed KV dtypes** — the dtypes that #185 auto-selects by default whenever fp32 doesn't fit (i.e. exactly the configs real 12 GB-card users run at long context). Known-latent areas with no test (issue #166): the bf16 SWA-ring append wrap and the CudaHybridForwardPass batched narrowed-KV path.

## Proposal

- Add bf16 and q8_0 variants of the Gemma4 12B (and Qwen3-8B dense) CUDA coherence tests using a **template-correct prompt** (per the `prompt must match chat template` lesson — extract `tokenizer.chat_template`, verify token IDs) so the top-logit margin is healthy and dtype rounding cannot flip the assertion. Assert ≥2 distinct tokens + non-EOS, optionally argmax agreement with fp32 on the first decode step.
- Exercise the SWA-ring wrap under bf16: prefill past `sliding_window` (512) so the ring actually wraps, then assert decode coherence (closes the latent-only half of #166).
- Keep the synthetic-prompt fp32 tests as-is — they are the trunk-math guards.

Issue #190, opened 2026-06-10
feat(cuda): continuous batching on the CUDA backend (per-sequence GPU KV caches + batched decode + packed prefill)


pekkah:
## Context

Issue #183 / PR #188 closed the scheduling gaps in `ContinuousBatchingEngine` — but the engine remains CPU-only: `InferenceEngineLoader` enables it only when the forward pass is the CPU `ForwardPass` (`batchingSupported` is false on every GPU/hybrid path). `CudaForwardPass` has no `CreateCache` / `PrefillWithCache` / `BatchForwardMulti` / `PrefillPackedMulti` equivalents — multi-user serving on the GPU falls back to the serialized single-user `InferenceEngine`.

This is the real remaining "GPU eating air" gap behind #183's theme: a 4070 Ti decoding one Qwen3-8B stream at ~75 t/s is using a fraction of its matvec bandwidth; batched decode amortizes weight reads N× exactly as the CPU `BatchForwardMulti` does.

## Scope (large — on the order of #124/#136)

1. **Per-sequence GPU KV caches** — today `_gpuKCache[layer]` is a single `[maxSeqLen, kvDim]` allocation owned by the forward pass. Needs a per-sequence cache object (paged or slab) + VRAM-budgeted allocation, which dovetails with the #185 KV-budget machinery and PR #188's admission backpressure (`KvBytesPerToken` analogue for the GPU dtypes fp32/bf16/q8_0).
2. **Batched decode kernels** — N tokens, N cache pointers per layer: batched matvec is the easy half (cuBLAS GEMM at N>1); attention needs per-sequence (cache, position) indirection (block-table or pointer-array kernel signature).
3. **Packed multi-prompt prefill** — the chunked batched-prefill trunk (#136, PrefillBatchChunk=4096) already exists for one sequence; a cu_seqlens-style packed variant mirrors `ForwardPass.PrefillPackedMulti` from PR #188.
4. **Engine generalization** — `ContinuousBatchingEngine` takes the concrete `ForwardPass`; needs an interface (e.g. `IBatchedForwardPass`) so the CUDA implementation can slot in, plus the loader gate.

Suggested order: 1 → 2 (decode batching alone is most of the throughput win) → 4 → 3.

## Constraints

- CUDA graphs (#136) assume a single cache layout — batched decode either bypasses graphs initially or captures per batch-size.
[...]

Issue #189, opened 2026-06-10
perf(cpu): dequant-once weight reuse for MatMulBatched — small-chunk prefill pays full dequant per call (#183 follow-up)


pekkah:
## Context

PR #188 (issue #183 Gap 1) chunks prompt admission so decode keeps flowing. The chunk-size sweep exposed a CPU cost cliff: every `SimdKernels.MatMulBatched` call re-pays the full weight dequantization regardless of batch size N, so chunked prefill throughput collapses as chunks shrink.

Measured (SmolLM2-1.7B Q4_K_M, Zen 4, OpenBLAS, engine-shaped chunked prefill of a 1040-token prompt):

| chunk tokens | effective prefill t/s |
|---:|---:|
| whole prompt (1040) | ~42 |
| 256 (default) | ~38 |
| 64 | ~20 |
| 32 | ~13 |

This is why the default chunk stays at 256 (≈8 s worst-case decode stall on this box) instead of something interactive like 32–64.

## Proposal

Amortize dequant across calls within a layer: a dequant-once fp32 (or bf16) weight cache keyed by (layer, tensor) with reuse when successive MatMulBatched calls hit the same weights — exactly the chunked-admission access pattern (`PrefillWithCache`/`PrefillPackedMulti` walk the same layer sequence every chunk). Alternatives: a persistent dequant buffer owned by ForwardPass for the BLAS path only, or chunk-aware fused dequant+GEMM.

## Win

Makes small chunks nearly free → default chunk can drop to 32–64 → worst-case decode stall under concurrent load drops from ~8 s toward decode-step scale, without the current −43% serial-chunk throughput penalty (packed prefill already recovers most of it when several prompts are in flight: 33.7 vs 20.4 t/s — see PR #188 `--cb --packed` A/B; this issue is about the single-prefilling-prompt case).

Bench harness: `dotnet run --project benchmarks/SharpInference.Bench -c Release -- --cb [--chunk N]` and `--cb --packed`.

pekkah:
Shipped in PR #192 (merged as commit `ab6a398` on master).

[...]

PR #192, opened 2026-06-10
feat(cuda): #190 dense continuous batching (+ #183 follow-ups #189/#191)

pekkah:
## Summary

Completes the `feat/183-followups-189-191-190` branch. The headline change is **#190 CUDA dense continuous batching**: `CudaForwardPass` now implements `IBatchedForwardPass`, so `ContinuousBatchingEngine` can drive **CUDA** dense serving (per-sequence GPU KV caches + a true batched decode that amortizes weight reads across the batch), not just the CPU path. Also folds in the remaining #183 follow-ups already on the branch (#189 dequant cache, #191 narrowed-KV coherence tests, the #190 batched-forward abstraction).

## #190 — CUDA dense continuous batching (new)

`CudaForwardPass : IForwardPass, IBatchedForwardPass`, scoped to the **dense** path.

- **`CreateCache`** → `CudaSequenceKvCache`: NumLayers full-context K/V pairs at the active KV dtype (with OOM cleanup). Disables CUDA graphs for the instance — the per-token decode graph bakes owned-cache device pointers, which a per-sequence rebind would invalidate.
- **Cache rebind**: `_gpuKCache`/`_gpuVCache` made mutable; `_ownedKCache`/`_ownedVCache` are the real allocation home (and the only arrays `Dispose` frees). `PrefillWithCache` = `BindCache` → existing batched-trunk `Prefill` → writeback `cache.Length` → `RestoreOwned` in a finally.
- **`BatchForwardMulti`** — *true* batched decode (not a rebind+`Forward` loop): batched embed → per layer {batched RmsNorm + batched QKV/O/FFN **GEMM-N**, with **per-sequence** RoPE / QK-norm / KV-append / single-query attention against each sequence's own cache} → batched final norm + one batched output GEMM. Mirrors the dense `RunDeviceRegion` ordering exactly (QK-norm before RoPE, #157), so it's argmax-stable vs the single-user loop.
- **`PrefillPackedMulti`** — per-sequence loop for now (cross-prompt packing is a follow-up).
- **`SupportsContinuousBatching`** — a single gate shared by the loader and the runtime guard so they can't diverge: dense-only, excludes MoE / Gemma-4 (per-layer head_dim) / TurboQuant / active SnapKV / final-logit softcap / non-GEMM-N-batchable Q4_0 weights. `KvBytesPerToken` is block-aware (`DTypeInfo.ByteSize`) so q8_0 KV doesn't throw.
- **Loader** (`InferenceEngineLoader`): the CUDA full-offload branch returns `cfwd.SupportsContinuousBatching`; engine construction is wrapped in the same dispose-on-throw guard as `BuildForwardPass`.

### Throughput (Qwen3-8B Q4_K_M, RTX 4070 Ti)

| batch N | aggregate t/s | × single-user (75 t/s) |
|--------:|--------------:|-----------------------:|
| 1 | 70 | 0.92× |
| 2 | 88 | 1.17× |
| 4 | 99 | **1.32×** |
| 8 | 106 | **1.41×** |

The GEMM-N matvec (grid Y = nTok) reads each weight tile once *per token-block*, so the win is launch-overhead amortization + L2 weight reuse. A/B against the compute-bound MMQ/GEMM path (`SHARPI_BATCH_DECODE_GEMM=1`) showed it is **worse** for small-N decode (N=4: 57 vs 99 t/s) — its fixed per-step costs (int8 activation conversion + Q6_K output-weight fp16 dequant every step) only amortize at the N=4096 prefill scale. So GEMM-N is the default; the toggle is kept for very high concurrency (curves converge near N=8).

## #183 follow-ups (from `ab6a398`, already on the branch)

- **#189** dequant-once CPU weight cache for the BLAS prefill path (bit-identical).
- **#191** narrowed-KV coherence tests.
- **#190** the backend-agnostic `IBatchedForwardPass` / `ISequenceKvCache` abstraction + CPU wiring (this PR adds the CUDA implementer).

## Review

An adversarial multi-agent review of the diff caught and fixed: a **blocker** (`KvBytesPerToken` threw `ArgumentException` for q8_0 KV → block-aware `ByteSize`), a `CreateCache` OOM partial-allocation leak, the non-Gemma Q4_0 gate gap (GEMM-N has no Q4_0 kernel → would crash at first decode), a latent final-logit-softcap divergence, and an int-overflow guard on the host logits buffer. The single-gate `SupportsContinuousBatching` resolves the loader-vs-runtime divergence those findings flagged.

## Testing (RTX 4070 Ti)

- **New** `CudaBatchForwardMultiTests` (5): batched N=2 / two decode steps / prefill-with-cache / chunked prefill — each vs single-user `Forward`/`Prefill` (argmax + top-5 + maxAbs<1.0). `CudaBatchedDecodeBench` (`SHARPI_BENCH_BATCH=1`) surfaces the throughput table above.
- **No regression**: Qwen3 dense prefill 7/7, full KV-dtype suite (bf16/q8_0 + Gemma-4 aliased-KV), CPU `ContinuousBatchingTests` 13/13, Server 103/103, CUDA SnapKV 3/3. Full-solution Release build clean (0 warnings, TreatWarningsAsErrors).

## Notes

- At long context (ctx ≳ 2048) VRAM auto-SnapKV engages and (correctly) routes the model to the single-user engine. To use batched CUDA serving at long context, run with `SHARPI_SNAPKV_BUDGET=0` or `--kv-type bf16`.
- Follow-ups: cross-prompt packed prefill (`PrefillPackedMulti` is sequential today); MoE / Gemma-4 / TQ / SnapKV batched serving remain out of scope.

### DIFF
--- a/benchmarks/SharpInference.Bench/ContinuousBatchingHarness.cs
+++ b/benchmarks/SharpInference.Bench/ContinuousBatchingHarness.cs
@@ -55,6 +55,16 @@ public static class ContinuousBatchingHarness
         using var backend = new CpuBackend();
         using var fwd = new ForwardPass(model, backend, hp);
 
+        // Issue #189: report the dequant-once weight-cache state so the prefill A/B below is
+        // interpretable. Force it off for the "before" numbers with SHARPI_PREFILL_DEQUANT_MB=0.
+        Console.WriteLine($"[cb] OpenBLAS: {(SimdKernels.BlasAvailable ? "LOADED" : "not found")}");
+        string budgetStr = fwd.PrefillDequantCacheBudgetBytes == long.MaxValue
+            ? "unlimited"
+            : $"{fwd.PrefillDequantCacheBudgetBytes / (1024 * 1024)} MiB";
+        Console.WriteLine($"[cb] dequant cache: {(fwd.PrefillDequantCacheActive
+            ? $"ACTIVE ({budgetStr} budget, covers model)"
+            : fwd.PrefillDequantCacheBudgetBytes > 0 ? "partial (budget < model)" : "off")}");
+
         // Long prompt: repeat filler text until the tokenizer crosses the target count.
         const string filler = "The quick brown fox jumps over the lazy dog near the quiet river bank. ";
         var sb = new System.Text.StringBuilder("<|im_start|>user\nSummarize the following text:\n");
@@ -77,7 +87,9 @@ public static class ContinuousBatchingHarness
 
         if (args.Contains("--packed"))
         {
-            RunPackedPrefillAb(fwd, tokenizer, longPrompt);
+            // --chunk N drives the per-prompt chunk size of the chunked modes so the A/B can
+            // reproduce the issue #189 chunk sweep (e.g. --packed --chunk 32 / 64). Default 64.
+            RunPackedPrefillAb(fwd, tokenizer, longPrompt, chunkArg > 0 ? chunkArg : 64);
             return;
         }
 
@@ -95,7 +107,7 @@ public static class ContinuousBatchingHarness
     /// (one PrefillWithCache per prompt) vs as ONE packed PrefillPackedMulti call.
     /// Isolates the weight-read amortization across prompts from the scheduling change.
     /// </summary>
-    private static void RunPackedPrefillAb(ForwardPass fwd, GgufTokenizer tokenizer, string longText)
+    private static void RunPackedPrefillAb(ForwardPass fwd, GgufTokenizer tokenizer, string longText, int perSeqChunk)
     {
         const int S = 4, T = 256;
         var all = tokenizer.Encode(longText).ToArray();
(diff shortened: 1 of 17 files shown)


pekkah:
Thanks — addressed the review in 4ba39ce:

**High (all applied, in `BatchForwardMulti`):**
- Moved the per-sequence view allocation loop **inside** the `try` and null-check each view in the `finally`, so a mid-loop `View` throw frees the views taken so far (same OOM-cleanup pattern as `CreateCache`).
- Deferred the `caches[n].Length` advance to **after** the pass completes + synchronizes (transactional — a mid-pass throw now leaves the logical length untouched), instead of rewriting it every layer.

**Medium (applied):** `ArgumentNullException.ThrowIfNull` guards on the batched-serving entry points (`BatchForwardMulti` / `PrefillWithCache` / `PrefillPackedMulti`), the `CudaSequenceKvCache` ctor, and the `ContinuousBatchingEngine` ctor.

**Medium (skipped, with rationale):** the dequant-cache buffer in `ForwardPass` already frees-on-throw around `Dequantize.ToFloat32` (the only realistic failure); extending the guard to the dict-insert path covers only a Dictionary-insert OOM, where the process is already failing — not worth restructuring the validated #189 code.

All guards are error-path only (happy path unchanged); 18 tests green (5 CUDA batched-parity + 13 CPU `ContinuousBatchingTests`).

Issue #198, opened 2026-06-10
perf(cuda): batched-decode routing above WS capacity — N>16 silently falls back to GEMM-N matvec (#194 follow-up)


pekkah:
## Context

The #194 weight-stationary kernels are stamped per compile-time batch capacity (2/4/8/16, `CudaWsKernels.Variants`); `MatMulBatchedWeightStationary` delegates `nTok > 16` to the GEMM-N matvec, which re-streams the weight per token. So with `SHARPI_MAX_BATCH > 16`, decode steps at N>16 lose the weight-amortization win exactly where it matters most.

Nobody runs that configuration today (a 12 GB card can't hold 16+ full-context KV caches for Qwen3-8B at ctx 2048 anyway — ~604 MB/seq), so this is **low priority**, but the cliff is silent.

## Options

1. **Sub-batch split**: drive N>16 as ⌈N/16⌉ weight-stationary calls — weight re-streamed once per sub-batch, still 16× better amortization than GEMM-N. Trivial, no new kernels; likely the right first move.
2. **Extend capacities** (NT=32): needs a register-pressure/occupancy check — 32 fp32 accumulators/thread on top of the dp4a working set may spill on the Q4_K 8-warp kernel.
3. **Compute-bound crossover**: re-bench `SHARPI_BATCH_DECODE_GEMM=1` at N=16/32 — the MMQ/GEMM fixed per-step costs (int8 activation conversion + Q6_K output fp16 dequant) may finally amortize there, and #190's "curves converge near N=8" note predates the WS bar moving from 106→183.

Either way, add an N=16 cell to `CudaBatchedDecodeBench` (needs a shorter prompt/ctx so 16 caches fit in 12 GB).

Refs: #194, #190.

Issue #197, opened 2026-06-10
perf(cuda): batch the per-sequence attention/RoPE/KV-append loop in batched decode — N=8 flattens at 2.42x (#194 follow-up)


pekkah:
## Context

#194 made the batched-decode matmuls weight-stationary, lifting aggregate decode (Qwen3-8B Q4_K_M, RTX 4070 Ti, single-user = 75 t/s):

| N | GEMM-N (#190) | weight-stationary (#194) | × single-user |
|--:|--:|--:|--:|
| 2 | 88.7 | 123.4 | 1.64× |
| 4 | 99.6 | 169.4 | 2.25× |
| 8 | 105.9 | 182.7 | 2.42× |

The curve still flattens hard toward N=8 (22.8 t/s/seq). With weight HBM reads now amortized N×, the residual per-step cost is the **serial per-sequence block** in `CudaForwardPass.BatchForwardMulti`: per layer, a `for (n = 0..N)` loop issues ~6 small launches per sequence (optional bias adds, `HeadNorm` ×2, `RoPE` ×2, `KvAppendKv`, `AttentionKv`) — at N=8 that's 36 layers × 8 seqs × ~6 ≈ **1.7K launches per decode step**, each a low-occupancy single-token kernel, plus the N attention kernels run back-to-back instead of concurrently.

## Goal

Ragged-batched kernels for the per-sequence ops so the per-layer launch count is O(1) in N:

- batched QK-norm / RoPE over N tokens at **per-sequence positions** (the existing `llm_head_norm_batched` / `llm_rope_neox_*_batched` assume one sequence with consecutive positions — need a positions array),
- batched KV-append + batched single-query attention over **N distinct caches** (device-side pointer table for the per-sequence K/V base pointers + per-sequence lengths; ragged `pos+1` extents).

Must stay argmax-stable vs the per-token decode oracle (`CudaBatchForwardMultiTests` argmax + top-5 + maxAbs<1.0); the attention reduction itself should keep the existing per-(head, position) chain so it stays bit-close.

## Acceptance

- N=8 aggregate well above 183 t/s (per-seq attention is ~the whole remaining gap to ~weight-bound N×75 scaling at small N).
- `CudaBatchForwardMultiTests` + `ContinuousBatchingTests` green.

Refs: #194, #190, PR #192.

PR #199, opened 2026-06-10
perf(cuda): #194 weight-stationary batched-decode matmul — N=8 aggregate 106 -> 183 t/s

### LINKED ISSUE #194: perf(cuda): weight-stationary batched-decode matmul — amortize weight HBM reads across the batch (#190 follow-up)
## Context

In PR #192 (#190 CUDA dense continuous batching), `BatchForwardMulti` routes its batched matmuls through the **GEMM-N matvec** (`CudaBackend.MatMulBatched`, grid `(rows+7)/8 × nTok`). That kernel reads each weight tile **once per token-block** (one block-row per token), so it amortizes *launch overhead* and gets *L2 weight reuse* across the N tokens — but it does **not** amortize the weight **HBM** read.

Measured (Qwen3-8B Q4_K_M, RTX 4070 Ti, aggregate t/s; single-user = 75):

| N | t/s | × |
|--:|----:|--:|
| 1 | 70 | 0.92× |
| 2 | 88 | 1.17× |
| 4 | 99 | 1.32× |
| 8 | 106 | 1.41× |

A batched step at N=8 takes ~5.7× a single step, i.e. weight reads are *not* shared — the win is L2-driven. The compute-bound MMQ/GEMM path (`SHARPI_BATCH_DECODE_GEMM=1`, which *does* read each weight once per batch) is **worse** at decode batch sizes (N=4: 57 vs 99 t/s) because its fixed per-step costs (int8 activation conversion + the Q6_K output-weight fp16 dequant **every step**) only amortize at the N=4096 prefill scale.

## Goal

A **weight-stationary** batched-decode matmul for small N (2–16): load each weight tile to shared mem / registers **once** and apply it to all N activation rows, so the weight HBM read is amortized N× without prefill-scale fixed costs. This is the real lever to push batched decode well past the current ~1.4× toward the memory-bandwidth ceiling (theoretically up to ~Nx until compute-bound).

Likely a small-N GEMM-N variant that swaps the grid so the weight tile (not the token) is the outer loop, reusing the existing per-dtype dp4a/Q4_K-SoA dequant readers. Must stay argmax-stable vs the per-token decode oracle.

Refs: #190, #162 (Qwen3-8B Q4_K decode gap), #149/#152 (MMQ tiling), PR #192.

pekkah:
## Context

#190 batched decode routed its trunk + output matmuls through the GEMM-N matvec (`CudaBackend.MatMulBatched`, grid Y = token), which re-streams every weight tile from HBM once per token — the batch win was launch overhead + L2 reuse only, capping aggregate throughput at ~1.4× single-user at N=8. The compute-bound MMQ/GEMM alternative is worse at decode batch sizes (fixed per-step costs only amortize at prefill scale). Closes #194.

## What

- **`CudaWsKernels`** (new): weight-stationary small-N matvec kernels — the token loop runs **inside** the thread block, so each weight read serves all N activation rows. Bodies for F32 / Q4_K (AoS + #156 SoA, dp4a over Q8_1) / Q5_K / Q6_K / Q8_0 (AoS + #149 SoA), stamped per compile-time batch capacity (2/4/8/16) so per-token accumulators stay register-resident; excess tokens are predicated off (block-uniform, no divergence).
- **`CudaBackend.MatMulBatchedWeightStationary`**: same contract/token-major layout as `MatMulBatched`; N=1, N>16, and non-batchable dtypes delegate to the GEMM-N path. `MatMulBatched` itself (the GDN/MTP byte-parity surface) and all prefill paths (`PrefillBatchedTrunk` / `GpuMatMulBatchedCore`) are untouched.
- **`CudaForwardPass.BatchDecodeMatMul`**: weight-stationary is the new default; `SHARPI_BATCH_DECODE_WS=0` restores the #190 GEMM-N matvec, `SHARPI_BATCH_DECODE_GEMM=1` still routes the compute-bound A/B path. Env-only toggles, same tier as `SHARPI_BATCH_DECODE_GEMM` — no CLI/server-options surface (no operator trade-off: bit-identical output, strictly faster at every measured N).

**Bit-exact, not just argmax-stable**: each (row, token) pair keeps the GEMM-N kernel's exact reduction chain — token-invariant factors are hoisted only at existing left-fold boundaries (`super_d*sc` like the N2 dual kernels, `d*q` for Q8_0, `sc*(dec-32)` for Q6_K, `d1*q-dm1` for Q5_K) — so output is bit-identical to N sequential per-token matvecs.

## Measured (Qwen3-8B Q4_K_M, RTX 4070 Ti, aggregate decode t/s, single-user = 75)

| N | before (#190) | after | × single-user |
|--:|--:|--:|--:|
| 1 | 69.8 | 69.7 | 0.92× (delegates to GEMM-N) |
| 2 | 88.7 | **123.4** | 1.17× → **1.64×** |
| 4 | 99.6 | **169.4** | 1.32× → **2.25×** |
| 8 | 105.9 | **182.7** | 1.41× → **2.42×** |

`SHARPI_BATCH_DECODE_WS=0` reproduces the old 89/100/106 exactly — the whole delta is the new kernels.

## Validation

- **`CudaMatMulBatchedWsTests`** (new, 7/7): bit-exactness oracle vs **sequential per-token `MatMul`** (the independent reference, not the GEMM-N path the kernels mirror) — every dtype + SoA layout, batch sizes 1–17 covering each capacity variant, round-up predication, and both delegates; rows not divisible by the 8-row block group.
- `CudaBatchForwardMultiTests` 5/5 (batched N=2 / two-step / prefill-with-cache / chunked vs single-user Forward, argmax + top-5 + maxAbs<1.0).
- `Qwen3CudaBatchedPrefillTests` 7/7, `ContinuousBatchingTests` 13/13, `CudaMatMulBatchedTests` 15/15 (GEMM-N regression).

## Follow-ups

- #197 — batch the per-sequence attention/RoPE/KV-append loop (the remaining N=8 flattening; ~1.7K launches per decode step).
- #198 — N>16 routing above WS capacity (silent GEMM-N fallback; sub-batch split / NT=32 / compute-bound crossover).

### DIFF
--- a/src/SharpInference.Cuda/CudaBackend.cs
+++ b/src/SharpInference.Cuda/CudaBackend.cs
@@ -173,6 +173,16 @@ public sealed unsafe class CudaBackend : IComputeBackend, IImageOpsBackend, IDis
     private nint   _matvecQ5KGemmNKernel;
     private nint   _matvecQ6KGemmNKernel;
     private nint   _matvecQ80GemmNKernel;
+    // Issue #194: weight-stationary small-N batched-decode matvecs — token loop inside
+    // the block so each weight read is amortized across the batch. One handle per
+    // compile-time batch capacity (CudaWsKernels.Variants: 2/4/8/16), indexed by position.
+    private readonly nint[] _matvecF32WsKernels    = new nint[CudaWsKernels.Variants.Length];
+    private readonly nint[] _matvecQ4KWsKernels    = new nint[CudaWsKernels.Variants.Length];
+    private readonly nint[] _matvecQ4KWsSoaKernels = new nint[CudaWsKernels.Variants.Length];
+    private readonly nint[] _matvecQ5KWsKernels    = new nint[CudaWsKernels.Variants.Length];
+    private readonly nint[] _matvecQ6KWsKernels    = new nint[CudaWsKernels.Variants.Length];
+    private readonly nint[] _matvecQ80WsKernels    = new nint[CudaWsKernels.Variants.Length];
+    private readonly nint[] _matvecQ80WsSoaKernels = new nint[CudaWsKernels.Variants.Length];
     // Issue #141: compute-bound prefill GEMM — dequant Q8_0 weight + convert
     // activations to fp16, then one cublasGemmEx (weight read once per batch).
     private nint   _dequantQ80F16Kernel;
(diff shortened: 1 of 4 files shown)


pekkah:
Reviewed the two bot findings — declining both: (1) `BitConverter.GetBytes(Half)` exists since .NET 5 (build is warnings-as-errors clean and the 7 oracle tests ran green); the helper is verbatim from the sibling CudaMatMulBatchedTests.cs, kept for consistency. (2) The Q6_K manual shift/mask is intentionally byte-for-byte identical to llm_matvec_q6k_gemm_n — the kernel the WS variant must mirror exactly for the bit-exactness contract.

PR #200, opened 2026-06-10
perf(cuda): #197 ragged-batched decode attention/RoPE/KV-append

### LINKED ISSUE #197: perf(cuda): batch the per-sequence attention/RoPE/KV-append loop in batched decode — N=8 flattens at 2.42x (#194 follow-up)

pekkah:
## Summary

After #194 (weight-stationary batched-decode matmuls), the residual per-step cost in `CudaForwardPass.BatchForwardMulti` was the serial per-sequence block: ~6 low-occupancy single-token launches per sequence per layer (QK-norm ×2, RoPE ×2, KV-append, single-query attention) — ~1.7K launches per decode step at N=8, with the N attention kernels running back-to-back instead of concurrently. This PR makes the per-layer launch count O(1) in N.

- **`CudaRaggedKernels`** — ragged RoPE (per-sequence positions array; the existing `*_batched` kernels assume one sequence at consecutive `basePos+t`), ragged KV-append + ragged single-query attention over **N distinct caches** at ragged lengths `pos[n]+1` (f32/bf16/q8_0 KV dtypes), and a broadcast bias-row add. Per-sequence positions and cache base pointers are passed as **by-value struct kernel parameters** (capacity 16, host chunks above that): the driver snapshots them with the launch command itself, so there is no device-side pointer table, no host→device upload, and no sync on the hot path. Batched decode never captures CUDA graphs (`CreateCache` disables them), so parameter-baked pointers are safe.
- **`CudaBackend`** — `RoPEBatchedRagged` / `KvAppendBatchedRagged` / `AttentionBatchedRagged` / `AddBiasBatched` chunking wrappers; the attention spill scratch becomes per-(sequence, head) rows `[N × numHeads × maxSeqLen]`, validated fail-loud.
- **`CudaForwardPass.BatchForwardMulti`** — ragged path default-on; the `[layer][seq]` cache tensor table rebuilds only when the batch composition changes (element identity), the spill scratch allocates only when a sequence exceeds the 4096-slot shared-memory fast path; QK-norm rides the existing `HeadNormQkBatched` (one launch for Q+K). **`SHARPI_BATCH_DECODE_RAGGED=0`** restores the #190 per-sequence loop.

Each ragged kernel keeps its per-token counterpart's exact per-(head, position) reduction chain — only the row/cache indirection is batched — so per sequence the output is **bit-identical** to the sequential kernels.

## Throughput (Qwen3-8B Q4_K_M, RTX 4070 Ti, single-user 75.7 t/s)

| N | #194 (ragged off) | ragged (this PR) | Δ | × single-user |
|--:|--:|--:|--:|--:|
| 1 | 69.8 | 69.2 | — | 0.91× |
| 2 | 123.1 | 128.8 | +4.6% | 1.70× |
| 4 | 169.2 | 207.5 | +22.6% | 2.74× |
| 8 | 188.5 | **246.1** | **+30.6%** | **3.25×** |

Acceptance met: N=8 well above 183 t/s, N=4 above 169. `SHARPI_BATCH_DECODE_WS=0` A/B intact (90/104/113 at N=2/4/8 — tracks the #190 GEMM-N curve; that path is matmul-bound so ragged attention moves it little).

## Validation

- **Kernel isolation** (`CudaRaggedDecodeKernelTests`, 10 tests): bit-exactness vs the **sequential per-token kernels** (the independent reference) — chunk-boundary batch sizes (17, 20), ragged positions, all three KV dtypes, q8_0 warp-collaborative append, and the >4096 spill-scratch path.
- **E2E parity**: `CudaBatchForwardMultiTests` 5/5 (argmax + top-5 + maxAbs<1.0 vs single-user `Forward`) and `ContinuousBatchingTests` 13/13 — green on the ragged path **and** with `SHARPI_BATCH_DECODE_RAGGED=0`.
- `CudaMatMulBatched*` suites (22 tests) still green.

Scope: dense batched decode only (the `SupportsContinuousBatching` gate). Per-token `Forward`, prefill paths, and `MatMulBatched` (GDN/MTP byte-parity surface) untouched. SnapKV remains excluded (`_kvEvictedCount == 0`, physical slot == pos).

Closes #197.

### DIFF
--- a/src/SharpInference.Cuda/CudaBackend.cs
+++ b/src/SharpInference.Cuda/CudaBackend.cs
@@ -281,6 +281,20 @@ public sealed unsafe class CudaBackend : IComputeBackend, IImageOpsBackend, IDis
     private nint   _fullSeqAttentionBf16Kernel;
     private nint   _fullSeqAttentionGlobalKernel;
     private nint   _fullSeqAttentionGlobalBf16Kernel;
+
+    // Issue #197: ragged-batched decode kernels — one launch covers all N sequences'
+    // RoPE / KV-append / single-query attention at per-sequence positions against N
+    // distinct per-sequence caches (positions + cache pointer tables ride in by-value
+    // struct kernel parameters; see CudaRaggedKernels).
+    private nint   _ropeNeoxRaggedKernel;
+    private nint   _ropeInterleavedRaggedKernel;
+    private nint   _kvAppendRaggedKernel;
+    private nint   _kvAppendRaggedBf16Kernel;
+    private nint   _kvAppendRaggedQ8Kernel;
+    private nint   _attentionRaggedKernel;
+    private nint   _attentionRaggedBf16Kernel;
+    private nint   _attentionRaggedQ8Kernel;
+    private nint   _addBiasRowsKernel;
     // Issue #141 (attention): memory-efficient flash-attention prefill (shared K/V
     // tiles reused across a query tile + online softmax) replacing the scalar
     // full_seq / swa_batched kernels' O(n²) global K/V re-reads.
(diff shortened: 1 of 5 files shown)


pekkah:
Profiled the residual N=8 flattening (nsys + ncu, details in #201): 85% of the 32.5 ms step is the two #194 weight-stationary matvec kernels at ~3.3x the weight-streaming floor — Q4_K is L1TEX/LSU-pipe-bound (94% pipe, 30% DRAM), Q6_K adds a latency-bound serial block walk and a 128-block grid on the attn_v shapes (35% occupancy, 8% DRAM). The ragged attention this PR adds is down to 5% of the step. Follow-up levers (SoA Q8_1 acts in the WS token loop, q6k geometry fix, small-N MMQ) are in #201. The last commit here adds the SHARPI_BENCH_BATCH_N/SHARPI_BENCH_SINGLE/SHARPI_BENCH_STEPS bench knobs used for the captures.

pekkah:
Re the powf-in-kernel suggestion: intentionally kept — the ragged kernels' contract is bit-identity with the sequential per-token RoPE kernels, which compute freq the same way (powf per pair), and the nsys profile in #201 puts llm_rope_neox_ragged at 0.12 ms of a 32.5 ms N=8 step (0.4%). A host-side freq table would trade away the bit-exactness oracle for noise-level perf; the actual hotspot (85% of the step) is the WS matvecs, tracked in #201.

PR #202, opened 2026-06-10
perf(cuda): #201 q6k scale-word WS decode + opt-in Q4_K decode MMQ (N=8: 246 -> 317 t/s)

### LINKED ISSUE #201: perf(cuda): batched-decode WS matvecs are LSU/latency-bound at N=8 — 27.7 of 32.5 ms/step, 3.3x over weight floor (#197 follow-up)
## Context

#197 (PR #200) batched the per-sequence attention/RoPE/KV-append ops; aggregate decode is now N=2/4/8 → 129/208/246 t/s (Qwen3-8B Q4_K_M, RTX 4070 Ti, single-user 75.7). But the curve still flattens: per-seq 30.8 t/s at N=8, and step time scales 14.5 → 15.5 → 19.3 → 32.5 ms for N=1/2/4/8 against a fixed ~9.7 ms whole-step weight-streaming floor (4.37 GB trunk + 0.51 GB output @ 504 GB/s).

Profiled the N=8 step (nsys per-kernel sums over 136 steps + ncu SOL/occupancy on the hot kernels; capture via the new `SHARPI_BENCH_BATCH_N` / `SHARPI_BENCH_SINGLE` / `SHARPI_BENCH_STEPS` bench knobs):

| kernel | inst/step | ms/step | share |
|---|--:|--:|--:|
| `llm_matvec_q4k_ws_soa_n8` (Q,K,O,gate,up + half V/down) | 216 | 19.0 | **58%** |
| `llm_matvec_q6k_ws_n8` (half V/down + output proj) | 37 | 8.8 | **27%** |
| `llm_attention_ragged` | 36 | 1.7 | 5% |
| `llm_quantize_q8_1` | 216 | 0.4 | 1% |
| norms/rope/append/silu/embed/etc. | ~270 | ~0.7 | 2% |

**85% of the step is the two #194 weight-stationary matvec kernels, running ~3.3× above the weight-streaming floor.** The #197 ops are a rounding error now. ncu says why (N=8, per shape):

| kernel / shape | vs floor | L1/TEX | DRAM | achieved occ | diagnosis |
|---|--:|--:|--:|--:|---|
| q4k gate/up [12288×4096] | 3.2× | **94.7%** | 32.8% | 80% | **LSU/L1TEX-pipe saturated** |
| q4k Q/O [4096×4096] | 3.5× | **89.5%** | 29.6% | 79% | same |
| q4k K/V [1024×4096] | 4.7× | 71.7% | 22.5% | 77% | pipe + small-grid latency |
| q6k output [151936×4096] | 3.0× | **84.1%** | 33.5% | 83% | LSU-pipe saturated |
| q6k ffn_down [4096×12288] | ~5× | 47.9% | 19.2% | 76% | serial per-warp block walk, latency-bound |
| q6k attn_v [1024×4096] | ~9× | 20.8% | **8.3%** | **35%** | grid = 128 blocks — GPU mostly idle |

(ncu locks clocks to base, durations ~1.4× nsys; ratios are the signal. Registers 48/thread, no spills — occupancy is not the limiter except via grid size.)

## Diagnosis

At NT=8 the WS kernels leave the weight-bandwidth-bound regime: the token loop multiplies per-weight-word **scalar load instructions** ×8 while the weight read stays fixed. Q4_K's inner loop per (lane, super-block, token) is 2 activation loads + 2 Q8_1-scale `__ldg` + 4 dp4a + ~8 fp ops → ~1 warp-instruction per weight byte at NT=8, and the L1TEX/LSU pipe hits 90-95% while DRAM idles at 30%. Q6_K is worse: scalar fp32 activation loads (8 per token per super-block per lane), a serial per-warp walk over all super-blocks (no warp striding like q4k_soa), and tiny grids on the [1024×4096] shapes (128 blocks on 60 SMs).

If the matvecs ran at the weight floor, N=8 would be ~14.5 ms/step → **~550 t/s aggregate** (2.2× headroom).

## Candidate fixes (in increasing ceiling / decreasing parity strength)

1. **Bit-exact, low risk — cut LSU instructions in the q4k WS token loop:** route the WS path through SoA Q8_1 activations (`llm_quantize_q8_1_soa` exists since #173): contiguous quants enable wider/paired loads, and the per-token fp16 scale loads+converts (currently 2 `__ldg`+2 converts per lane per token per block) come from a separate pre-converted array. Loads don't change the FP association, so bit-identity to the per-token oracle holds.
2. **Bit-exact — fix the q6k geometry:** adopt q4k_soa's layout (one row per block, 8 warps striding super-blocks + shared reduce) for `llm_matvec_q6k_ws_*`: 8× the blocks for attn_v/K-sized shapes (128 → 1024) and 8× shorter serial chains per warp. Note the warp/shared reduce order changes vs the current q6k WS kernel — it stays bit-identical only if stamped as the same reduction tree as `llm_matvec_q4k_ws_soa`'s contract (re-validate with the CudaMatMulBatchedWsTests pattern vs sequential `MatMul`, which is the actual contract).
3. **Argmax-stable, big ceiling — small-N tensor-core MMQ:** an m16n8k32 int8 MMA path with n-tile = 8/16 over SoA Q8_1 activations (the `llm_mmq_*_soa_acts` machinery from #173, retiled for decode-N) replaces ~32 scalar dp4a/load instructions per warp with one MMA. The #190 finding that "MMQ only amortizes at prefill scale" was measured for the full GEMM route incl. Q6_K dequant→fp16→cuBLAS (1.9 GB of fp16 temps per step — that is why `SHARPI_BATCH_DECODE_GEMM=1` loses); a direct small-N MMQ with no fp16 temps and a new Q6_K MMQ kernel avoids that fixed cost. Gate as an A/B (`SHARPI_BATCH_DECODE_MMQ=1`?) under the argmax-stable contract like the existing GEMM toggle. Half the q4k token-loop dp4a work (`sum_lo/hi` block-sums) is also redundant with the `s` field Q8_1 already stores — usable in an argmax-stable variant even without MMA.

Sanity bound for (1)+(2): even halving the over-floor excess (27.7 → ~18 ms matvec) lifts N=8 to ~300 t/s; (3) approaches the ~550 t/s weight-bound ceiling.

Refs: #194, #197, PR #200, #173 (SoA Q8_1 acts), #190 (GEMM/MMQ decode A/B).

pekkah:
## Summary

Implements #201. Qwen3-8B Q4_K_M on RTX 4070 Ti, aggregate batched-decode t/s (single-user 75.6 unchanged):

| config | N=1 | N=2 | N=4 | N=8 |
|---|--:|--:|--:|--:|
| before (#197 / PR #200) | 68.8 | 128.8 | 207.5 | 246.1 |
| **default (bit-exact)** | 68.8 | **138.8** | 208.9 | 248.2 |
| **`SHARPI_BATCH_DECODE_MMQ=1`** | 68.8 | 138.9 | 208.6 | **317.4** (4.2× single-user) |

### 1. `llm_matvec_q6k_ws_sw_n*` — Q6_K scale-word decode (default on, bit-exact)

The Q6_K WS serial walk was latency-bound on its per-super-block scale/d decode: 10 dependent byte-gather loads. The sw kernel loads the 18-byte tail (bytes 192..209) as **five aligned words + funnel-shift extracts** — same bytes, same chain, **bit-identical** to N sequential `MatMul` calls (`CudaMatMulBatchedWsTests` enforce it). N=8 per-shape: ffn_down 5.14→4.79, lm-head 2.84→2.65, attn_v 0.80→0.75 ms/step. Most of the win shows at N=2 (+7.8% aggregate). Kill-switch `SHARPI_BATCH_DECODE_WS_V2=0`.

### 2. `llm_mmq_q4k_soa_acts_n16` — decode MMQ (opt-in, argmax-stable)

85% of the N=8 step was the two WS matvec families at ~3.3× the weight-streaming floor, and per ncu the q4k kernel is **L1TEX-wavefront-bound by its lane geometry, which the bit-identity contract freezes** (see dead ends below). The opt-in path relaxes to the argmax-stable contract (same as the prefill MMQ / `SHARPI_BATCH_DECODE_GEMM` toggle) and re-tiles the prefill MMQ to BN=16: SoA Q8_1 activations, one m16n8k32 s8 mma per warp per sub-block, each weight byte read from HBM **once per step**.

The K-step matters: a first cut with one sub-block (32 elems) per barrier pair measured **1.4–1.7× slower than the WS matvec** — one mma (~50 cycles) cannot hide a staging-load latency epoch. The shipped kernel stages one **256-element super-block** per barrier pair (raw quants, linear coalesced copies, nibble/scale decode at fragment-read time, +1-word padded strides against 8-way bank conflicts): gate/up 11.6→6.9 ms/step (1.7× floor), Q/O 6.5→4.5.

Per-tensor dispatch falls back to bit-exact WS for Q6_K, rows<2048 (grid-starved K/V shapes), and N<5 (BN=16 tile mostly predicated off: N=2 measured 123 vs 139 t/s).

### Measured dead ends (implemented, profiled, removed)

- **Token-warp WS geometry** (bit-exact by construction, one warp per (row, token)): loses everywhere — NT× weight-decode replication outweighs the occupancy win (N=8: 243→239 routed only to the grid-starved attn_v shape; →182 incl. ffn_down). The issue's "8 warps striding super-blocks" idea is bit-incompatible with the per-token `llm_matvec_q6k` serial chain, so token-warp was the only bit-exact geometry available — and it doesn't pay.
- **SoA-Q8_1-activation WS reads (xs)**: genuinely cut L1TEX 94.7→82%, but the second activation-region pointer costs 7 registers in the NT-unrolled token loop (48→55 → occupancy 80→63%): net −5 t/s; pinning 48 regs via launch bounds spills (−36 t/s).

### Validation

- `CudaMatMulBatchedWsTests`: bit-identity vs N sequential `MatMul` (batch sizes 1,2,3,4,5,8,11,16,17; all dtypes; `WS_V2` kill-switch both ways) — green.
- `CudaDecodeMmqTests` (new): fp32-CPU-reference tolerance for the n16 kernel incl. partial tile / grid.y=2 / WS-fallback shapes — green.
- New `BatchForwardMulti` MMQ argmax/top-5 e2e oracle vs single-user; `CudaBatchForwardMultiTests`, `CudaRaggedDecodeKernelTests`, `ContinuousBatchingTests` — 39 tests green.
- A/B sanity unchanged: `SHARPI_BATCH_DECODE_WS=0` → 89.9/103.9/112.3 (was 90/104/113); `SHARPI_BATCH_DECODE_RAGGED=0` → ~199 @ N=8.

Closes #201.

### DIFF
--- a/src/SharpInference.Cuda/CudaBackend.cs
+++ b/src/SharpInference.Cuda/CudaBackend.cs
@@ -183,6 +183,9 @@ public sealed unsafe class CudaBackend : IComputeBackend, IImageOpsBackend, IDis
     private readonly nint[] _matvecQ6KWsKernels    = new nint[CudaWsKernels.Variants.Length];
     private readonly nint[] _matvecQ80WsKernels    = new nint[CudaWsKernels.Variants.Length];
     private readonly nint[] _matvecQ80WsSoaKernels = new nint[CudaWsKernels.Variants.Length];
+    // #201: scale-word Q6_K WS variant + Q4_K decode MMQ (BN=16 int8 mma).
+    private readonly nint[] _matvecQ6KWsSwKernels    = new nint[CudaWsKernels.Variants.Length];
+    private nint _mmqQ4kSoaActsN16Kernel;
     // Issue #141: compute-bound prefill GEMM — dequant Q8_0 weight + convert
     // activations to fp16, then one cublasGemmEx (weight read once per batch).
     private nint   _dequantQ80F16Kernel;
@@ -415,6 +418,18 @@ public sealed unsafe class CudaBackend : IComputeBackend, IImageOpsBackend, IDis
     /// </summary>
     public bool ActSoaCpaEnabled { get; set; } =
         Environment.GetEnvironmentVariable("SHARPI_ACT_SOA_CPA") == "1";
+
+    /// <summary>
+    /// Issue #201: decode the Q6_K WS per-super-block scale/d tail (bytes 192..209)
+    /// via five aligned word loads + funnel-shift extracts (<c>llm_matvec_q6k_ws_sw_n*</c>)
+    /// instead of 10 dependent byte-gather loads. Same bytes, same chain —
+    /// bit-identical (CudaMatMulBatchedWsTests verify against N sequential MatMul
+    /// calls); cuts the LSU instruction count and the gather latency the serial walk
+    /// stalls on. <c>SHARPI_BATCH_DECODE_WS_V2=0</c> restores the #194 byte-gather
+    /// kernel. Settable so tests can A/B both generations on one backend.
+    /// </summary>
+    public bool WsV2Enabled { get; set; } =
+        Environment.GetEnvironmentVariable("SHARPI_BATCH_DECODE_WS_V2") != "0";
 
     /// <summary>Total VRAM on the active CUDA device, in bytes. Queried once at backend creation.</summary>
     public ulong VramBytes
@@ -2010,11 +2025,12 @@ public sealed unsafe class CudaBackend : IComputeBackend, IImageOpsBackend, IDis
         }
 
         // F32-input kernels: same (rows+7)/8 × 256-thread geometry as the GEMM-N
-        // dispatch, minus the token grid dimension.
+        // dispatch, minus the token grid dimension. #201: Q6_K defaults to the
+        // scale-word variant (bit-identical, word-loaded scale/d tail).
         bool q80Soa = weightDType == DType.Q8_0 && _soaHandles.ContainsKey(matrix.Handle);
         nint kernel = weightDType switch
         {
-            DType.Q6_K => _matvecQ6KWsKernels[variant],
+            DType.Q6_K => WsV2Enabled ? _matvecQ6KWsSwKernels[variant] : _matvecQ6KWsKernels[variant],
             DType.Q5_K => _matvecQ5KWsKernels[variant],
             DType.Q8_0 => q80Soa ? _matvecQ80WsSoaKernels[variant] : _matvecQ80WsKernels[variant],
             _          => _matvecF32WsKernels[variant],
@@ -2045,16 +2061,16 @@ public sealed unsafe class CudaBackend : IComputeBackend, IImageOpsBackend, IDis
 
         int subBlocks = cols / 32;                          // per token
         long totalSub = (long)subBlocks * nTok;
-        nuint q81Bytes = (nuint)(totalSub * 36L);
-        EnsureQ81BatchBuf(q81Bytes);
+        if ((long)cols * nTok > int.MaxValue)
+            throw new InvalidOperationException(
+                $"MatMulBatchedWeightStationary: cols*nTok ({(long)cols * nTok}) exceeds int range.");
+        int qN = (int)((long)cols * nTok);
+
+        EnsureQ81BatchBuf((nuint)(totalSub * 36L));
 
         {
             nint qInPtr  = xPtr;
             nint qOutPtr = _q81BatchBuf;
-            int  qN      = (int)((long)cols * nTok);
-            if ((long)cols * nTok > int.MaxValue)
-                throw new InvalidOperationException(
-                    $"MatMulBatchedWeightStationary: cols*nTok ({(long)cols * nTok}) exceeds int range.");
             nint* args = stackalloc nint[3]
             {
                 (nint)(&qInPtr), (nint)(&qOutPtr), (nint)(&qN)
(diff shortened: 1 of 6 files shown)


pekkah:
Review feedback disposition:

- **HalfToUInt16Bits in tests**: applied (226af31).
- **ElementCount > int.MaxValue guard before the rows/cols cast**: declined — `MatMulBatchedDecodeMmq` deliberately mirrors the exact argument-validation pattern of its siblings (`MatMulBatchedWeightStationary`, `MatMulBatchedMmq`, `MatMulBatchedGemm`); the callers are the internal decode trunk where rows ≤ vocab (151,936). If the guard is wanted it should land across all four methods in one pass, not diverge in one.
- **"Redundant" cols*nTok overflow check**: declined — it is only redundant if the previous suggestion is adopted; as written it matches `DispatchMatVecQ4KWs` verbatim and protects the quantize launch count.

Issue #213, opened 2026-06-11
perf(cuda): decode throughput degrades at long context with q8_0 KV (12 vs 65 tok/s)


pekkah:
## Problem

Decode throughput on the dense CUDA path with **q8_0 KV** degrades sharply as context grows: ~**12 tok/s at 23k context** vs ~**65 tok/s at low context** on Gemma 4 E4B (4070 Ti).

Live trace:
```
request prompt=23678tok ... decode=185tok (12 tok/s) total=44033ms
# vs low-context:
request prompt=335tok   ... decode=11tok  (65 tok/s)
```

## Notes / hypotheses

- Each decode step's attention reads the full KV for the global (full-causal) layers. At 23k positions × q8_0 dequant, the per-token KV read + dequant may dominate over the weight stream (the ~70 tok/s HBM ceiling for the 8 GB Q8_0 weights).
- Memory note suggested q8_0 KV *preserves* decode speed at long context (vs bf16 spilling the tied embed), so 12 tok/s is lower than expected — worth profiling whether it's the q8_0 attention kernel, the SWA ring handling, or genuine KV-bandwidth.
- Compare q8_0 vs bf16 vs fp32 KV decode t/s at 8k / 16k / 32k context to isolate the dtype contribution.

## Scope

- [ ] Profile a long-context decode step (Nsight) — split attention vs matvec vs KV read.
- [ ] A/B q8_0 / bf16 / fp32 KV decode t/s across context lengths.
- [ ] If it's the q8_0 attention kernel, optimize the dequant/read path; if it's inherent KV bandwidth, document the expected curve.

## References

- Surfaced alongside the prefix-cache issue while profiling Claude Code traffic on Gemma 4 E4B Q8_0, ctx capped at 90614 for q8_0 KV VRAM fit.

pekkah:
## Phase 1 profiling — the degradation is inherent KV bandwidth, not the q8 kernel

RTX 4070 Ti, Gemma 4 E4B Q8_0 weights, dense `CudaForwardPass`. Steady-state decode t/s (32 timed steps after a long prefill), per KV dtype:

| ctx | fp32 | bf16 | q8_0 |
|---|---|---|---|
| 2048 | 49.2 | 49.6 | 43.8 |
| 8192 | 31.5 | 31.9 | 24.5 |
| 16384 | 21.2 | 21.5 | 15.4 |
| 32768 | 12.7 | 12.7 | 8.6 |

[...]

Issue #212, opened 2026-06-11
perf(engine): prefix-cache gets 0 reuse under agentic clients — single-slot KV evicted by interleaved auxiliary requests


pekkah:
## Problem

Under an agentic client (Claude Code), every turn re-prefills the **entire** system prompt from scratch — the prefix cache gets **0 reuse** even though the ~23.7k-token system prompt + tools is identical across requests.

Live trace from the server (`SharpInference.Engine` Debug log, added in the tokenizer PR):

```
request prompt=6527tok  (reused=0 ...)            # my warmup
request prompt=335tok    (reused=0 ...)           # Claude Code auxiliary request
request prompt=23678tok  (reused=0) prefill=27755ms (853 tok/s) ttft=27958ms total=44033ms
```

Result: TTFT ≈ **28 s on every turn**, dominated by re-prefilling the unchanged 23.7k prefix.

## Root cause

`InferenceEngine` keeps a **single** previous-sequence slot (`_prevTokens`) and the dense `CudaForwardPass` KV cache holds **one contiguous sequence**. `FindCacheablePrefix` matches the new prompt against `_prevTokens` only.

Claude Code interleaves **small auxiliary requests** (topic/title/quota checks on a cheaper path — the 335-token request above) between the real prompts. Each one overwrites `_prevTokens` (and the KV cache content), so the next real request finds no reusable prefix and re-prefills the whole 23.7k system prompt.

If the cache survived the interleaving, turns 2+ would reuse ~23.7k tokens → **TTFT ~28s → near-zero** (decode only).

## Scope / options

- [ ] **Paged / multi-prefix KV cache** (vLLM-style block table) so multiple sequences' prefixes stay resident and the stable system-prefix isn't evicted by a short interleaved request. This is the real fix; it's a substantial change to the dense `CudaForwardPass` KV layout (currently one contiguous buffer + `TruncateTo`).
- [ ] **Cheaper mitigation:** keep the *longest* recent prefix resident and refuse to let a much shorter request evict it — but correctness requires `_prevTokens` to always reflect actual KV cache content, so this needs at least a 2-slot cache (one "pinned system prefix" + one "scratch"), not just a heuristic on `_prevTokens`.
- [ ] Either way: keep the existing single-sequence partial-rewind fast path; add multi-slot on top.
- [ ] Validate with the `reused=` field in the per-request Debug trace under a real Claude Code multi-turn session.

## References

[...]

Issue #211, opened 2026-06-11
feat(loader): support Gemma 4 E4B QAT q4_0 GGUF (missing attn_k_norm tensor layout)


pekkah:
## Problem

Google's official Gemma 4 **E4B** QAT `q4_0` GGUF fails to load. The `Q8_0` E4B model loads and runs fine on the same code path, so this is a GGUF tensor-layout gap specific to the QAT `q4_0` repack — not a quant-decode issue.

```
$ sharpi-cli -m E:/models/gemma-4-E4B_q4_0-it.gguf -g -1 --kv-type q8_0 -p "..." --temp 0
Error: Missing bias tensor: blk.24.attn_k_norm.weight
```

- Repo: `google/gemma-4-E4B-it-qat-q4_0-gguf`
- File: `gemma-4-E4B_q4_0-it.gguf` (5.15 GB; note the upstream filename uses `_q4_0-it`, not the 12B's `-it-qat-q4_0` pattern)
- The repo also ships an `mmproj` (vision) file we don't load.

## Why it's worth doing

E4B decode is HBM-bandwidth bound (~70 t/s on a 4070 Ti at Q8_0, i.e. at the `8 GB / 504 GB/s` ceiling). The `q4_0` weights are **5.15 GB vs 8.19 GB** → ~1.6× fewer bytes/token → ~1.6× faster decode at near-identical quality (QAT). It also frees ~3 GB VRAM (room for `fp32`/`bf16` KV at moderate context).

The 12B QAT `q4_0` (`#124`) already loads and runs the dense gemma4 Q4_0 path, so the decode kernels exist — this is purely about the **E4B loader** accepting this file's tensor set.

## Likely cause

E4B differs from the 12B by having **per-layer input embeddings (PLE)** and **per-layer QK-norm** (`attn_k_norm` / `attn_q_norm`, sized `numKvHeads*headDim` / `numHeads*headDim`). The gemma4 loader currently treats `blk.N.attn_k_norm.weight` as **required** and throws when absent. This QAT `q4_0` repack appears to either:
- omit the QK-norm tensors at some/all layers, or
- name/store them differently (e.g. a different suffix, or folded), or
- store them at a dtype/shape the loader's "bias tensor" lookup rejects.

The error message ("Missing **bias** tensor") also suggests the lookup path is the generic required-tensor check, not a QK-norm-aware one.

## Scope

[...]

PR #214, opened 2026-06-11
perf(tokenizer): O(n²) → O(n log n) SPM encode + per-request perf trace

pekkah:
## Summary

Fixes the real cause of "the server feels slow" under agentic clients (Claude Code): the **SPM tokenizer was O(n²)**, so big system prompts took tens of seconds to *tokenize* before any compute. Adds a per-request perf trace (via `ILogger`) that made the diagnosis possible and surfaces the remaining levers.

### 1. `perf(tokenizer)`: O(n²) → O(n log n) SPM encode
`GgufTokenizer.EncodeSpm` rescanned all adjacent pairs to find one best merge, applied it (O(n) `RemoveAt`), and repeated — O(n²) over the whole prompt. Replaced with llama.cpp's priority-queue approach (symbol linked-list + min-heap, O(1) staleness check via monotonic length). **Bit-identical output** (134 Tests.Core pass, incl. tokenizer parity).

| | before | after |
|---|---|---|
| encode 7,826-tok prompt | 30,832 ms | **61 ms** |
| encode 23,678-tok Claude Code prompt | ~90 s | **107 ms** |
| first request (real Claude Code) | minutes | **~44 s** (now real prefill+decode, not tokenize) |

### 2. `feat(server)`: per-request perf trace via `ILogger`
Debug-level structured log splitting encode / prefill / decode with throughput + prefix-reuse:
```
request prompt=23678tok (reused=0 new=23678) encode=107ms prefill=27755ms (853 tok/s) ttft=27958ms decode=185tok (12 tok/s) total=44033ms
```
`InferenceEngine` gets an `ILogger` wired from the host's `ILoggerFactory`; Engine takes a `Microsoft.Extensions.Logging.Abstractions` ref (abstractions only, AOT-safe). Off unless `"SharpInference.Engine": "Debug"` is enabled. Null logger (CLI/tests) = no-op.

## Follow-ups surfaced by the trace
- **#212** — prefix-cache gets 0 reuse under agentic clients (single-slot KV evicted by interleaved auxiliary requests). Biggest remaining multi-turn win (~28s/turn).
- **#213** — decode degrades at long context with q8_0 KV (12 vs 65 tok/s).
- **#211** — Gemma 4 E4B QAT q4_0 loader gap (the download entry above is for this).

## Test plan
- [x] Tests.Core (134) green — tokenizer parity holds
- [ ] Tests.ForwardPass / Tests.Server green (running)
- [x] Live: 23,678-token Claude Code request, encode 107ms, coherent output

### DIFF
--- a/scripts/download-model.ps1
+++ b/scripts/download-model.ps1
@@ -5,7 +5,7 @@
     Downloads from HuggingFace to the models/ directory. Skips if already present.
     Supports: smollm2, qwen3-8b, olmoe-1b-7b, llama31-70b, qwen3-coder-30b-a3b, qwen36-35b-a3b,
               qwen36-27b-mtp, qwen36-27b-mtp-q5, qwen36-35b-a3b-mtp, carnice-35b-a3b-mtp,
-              gemma4-12b-qat, gemma4-12b-q4km,
+              gemma4-12b-qat, gemma4-12b-q4km, gemma4-e4b-qat,
               llama4-scout, z-image-turbo, z-image-turbo-q8, realesrgan-x4
 .PARAMETER Model
     Which model to download. Default: downloads all text models (skips large image models).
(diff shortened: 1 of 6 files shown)


pekkah:
Thanks for the review. Addressed:

- **`ids` list capacity** — resolved: the merge core was extracted to `SpmMergePieces` and `EncodeSpm` now does `new List<int>(merged.Count)` (tighter bound than `n`, since surviving symbols ≤ merged.Count ≤ n).
- **Capture `Logger` in a local (thread-safety)** — applied: `var logger = Logger;` is captured once at request start; both the `IsEnabled` gate and the `finally` `LogDebug` use the local, removing the read-twice race and the null-forgiving deref risk.
- **Pin `Microsoft.Extensions.Logging.Abstractions` version** — keeping `Version="*"` to match the repo convention (`Directory.Build.props` + the existing package refs all use `*`; there is no central package management here). Pinning just this one would be inconsistent.

Also added a model-free parity test (`SpmMergeTests`) — a 5000-case fuzz asserting the new O(n log n) merge is byte-identical to the original O(n²) algorithm, since the existing SPM tests require a local GGUF.

Issue #220, opened 2026-06-11
perf(engine): TierPlanner prices Gemma 4 KV at uniform full-context fp32 — runtime uses SWA rings, KV-share aliasing and per-layer head_dim, so auto -g/ctx underestimates capacity ~4-6×


pekkah:
## Problem

`TierPlanner.Plan` prices GPU KV uniformly: `2 × gpuLayers × NumKvHeads × HeadDim × sizeof(float) × ctx` for every layer (`TierPlanner.cs:97-126`). The Gemma 4 runtime allocates nothing like that:

- **SWA layers get window-sized rings**, not full-context buffers: `CudaForwardPass.cs:809,833-834` — `layerCtx = SwaRingSize(_maxSeqLen, swaWindow)` when `hp.IsSwaLayer[i]`. Gemma 4 is ~5:1 SWA:global with a window far smaller than typical contexts.
- **KV-share layers allocate no KV at all** — they alias the source layer's K/V (`CudaForwardPass.cs:802-803`: "KV-share layers (KvSourceLayer[L] >= 0) alias the…").
- **Per-layer head_dim** (Gemma 4's mixed dims, cf. #195) is priced at the single `hp.HeadDim`.
- KV dtype is priced fp32 even when `SHARPI_KV_DTYPE=bf16/q8_0` narrows it 2–4× (also noted in #215; #213's live config runs q8_0 KV).

Net effect: for Gemma 4 the planner overestimates per-token KV cost by roughly **4–6×** (more with narrowed KV). On a 12 GB card with `-g -1`, that means fewer GPU layers and/or a much smaller `RecommendedCtxSize` than what actually fits — the loader's `ClampGemma4KvShareBoundary` (`InferenceEngineLoader.cs:216`) already patches one Gemma-4 blind spot in the planner, and this is the rest of the family.

This matters most for the hybrid configs (Gemma 4 12B-class on 12 GB, where every reclaimed GB is another GPU layer) and for auto-context on E4B long-context runs.

## Proposed work

1. Teach `TierPlanner` the per-layer KV shape: when `hp.IsSwaLayer` is populated, price SWA layers at `SwaRingSize(ctx, hp.SlidingWindowSize)` positions; skip KV-share layers (`hp.KvSourceLayer[L] >= 0`); use per-layer head_dim where the hyperparams carry it. Reuse the same helpers `CudaForwardPass` uses so the two can't drift (single source of truth for ring sizing).
[...]

Issue #215, opened 2026-06-11
perf(cuda,engine): CPU-MoE mode + expert-aware TierPlanner for CudaHybridForwardPass — auto placement is structurally wrong for pure-attention MoE on 12 GB


pekkah:
## Context

Target: single-user decode/prefill on 12 GB VRAM / 64 GB RAM. The GDN-hybrid class (`CudaHybridGdnForwardPass`) already has the right low-VRAM MoE configuration: **all trunk layers on GPU + routed experts on CPU**, with a capacity-driven auto-router (`CudaHybridGdnForwardPass.cs:886-909`: SLRU capacity ratio &lt; 0.5 → CPU MoE, `SHARPI_CPU_MOE` override). The #100 verification showed why this matters: on a 4070 Ti 12 GB, CPU MoE is **11.8 t/s vs 6.1 t/s** GPU-SLRU on the structurally identical qwen3.6-A3B.

The **pure-attention MoE hybrid** (`CudaHybridForwardPass` — Qwen3-Coder-30B-A3B, OLMoE, qwen2moe-class models) has neither piece:

## Problem 1 — TierPlanner books all routed experts per layer, but the runtime never uploads them eagerly

`TierPlanner.MeasureLayerBytes` (`TierPlanner.cs:165-180`) includes `ffn_gate_exps` / `ffn_up_exps` / `ffn_down_exps` — i.e. **every expert** — in each MoE layer's byte cost. But `CudaHybridForwardPass` uploads routed experts **lazily via the SLRU slot manager**, not per layer (`CudaHybridForwardPass.cs:2805-2806`: "Routed-expert weights are uploaded lazily by CudaExpertSlotManager on cache miss"; the shared expert and router stay resident).

Consequence of `-g -1` (`InferenceEngineLoader.cs:209-242`, same pattern in `RunCommand.cs`): the greedy packer charges ~N×expert bytes per layer, stops packing far too early, and trunk layers whose *actual* eager footprint is tiny (attn q/k/v/o + norms + router + shexp) land on CPU. Each CPU layer then pays CPU attention plus the per-token pinned-buffer boundary crossings in `Forward` (`CudaHybridForwardPass.cs:890-936`), while the "saved" VRAM just becomes SLRU headroom for fewer GPU layers.

## Problem 2 — no "GPU trunk + CPU routed experts" mode at all

[...]

Issue #100, opened 2026-05-30
perf(cuda): llm_matvec_q3_k NVRTC + UploadRaw Q3_K/Q8_0 + TierPlanner byte accounting


pekkah:
## Context
Sibling of rank-1 / rank-2 issues. Targets the **GPU side** of Carnice mixed-precision routed-expert path.

CUDA kernels for Q3_K and Q8_0 routed-expert matvec are **missing**:
- `CudaBackend.cs:1187-1193` throws `NotSupportedException`
- `CudaTextKernels.cs` only ships Q4_K / Q5_K / Q6_K / F32
- `CudaExpertSlotManager.cs:273-281` and `CudaHybridGdnForwardPass.cs:2755-2762` silently dequant-to-F32 -> ~25% VRAM waste

`TierPlanner.cs:197-203` also has stale Q5_K accounting and zero Q3_K/Q8_0 accounting.

## Honest framing — zero Carnice lift on 12 GB cards
Adversarial verification showed: per-expert footprint is ~1.43-1.51 MB (not 880 KB) -> SLRU lands at **39-43% slot capacity** on RTX 4070 Ti 12 GB -> below the 50% auto-router threshold at `CudaHybridGdnForwardPass.cs:654` -> squarely in the documented regression band (`:636-637`: 6.1 t/s GPU SLRU vs 11.8 t/s CPU MoE on structurally identical qwen3.6-A3B).

**Auto-router correctly keeps Carnice on CPU MoE on a 4070 Ti.** This work is valuable for two reasons:
1. **VRAM reclaim** — removes ~25% dequant-to-F32 waste on every machine.
2. **Larger-card unblock** — RTX 4090 / 5090 16+ GB cards cross the 50% threshold; estimated 7-11 t/s on Carnice there.

## Scope (v1)
- New NVRTC kernel `llm_matvec_q3_k` in `CudaTextKernels.cs` modeled on `llm_matvec_q5_k` (Q3_K block: 32 B hmask + 64 B qs + 12 B 6-bit scales + 1 fp16 d = 110 B / 256 elem)
- Extend `UploadRaw` to Q3_K and Q8_0 (today only Q4_K / Q5_K go raw; rest dequant to F32)
- Fix `TierPlanner` byte accounting for Q5_K (already shipped), Q3_K, Q8_0
- **Auto-router stays on** — Carnice on 4070 Ti continues to pick CPU MoE (correct decision)
[...]

PR #223, opened 2026-06-12
perf(engine): TierPlanner KV pricing (#220) + CPU-MoE execution mode for CudaHybridForwardPass (#215)

pekkah:
Two commits closing **#220** and the compute-half of **#215**.

## #220 — TierPlanner prices KV at real dtype + per-layer shape
The layer planner priced GPU KV with a flat `2·gpuLayers·NumKvHeads·HeadDim·sizeof(float)·ctx` formula — wrong for Gemma-4-class models (SWA window-rings, KV-share aliasing, per-layer head_dim/kv-heads) and blind to narrowed KV dtypes (bf16/q8_0). It also charged every routed MoE expert against the trunk weight budget even though the runtime streams them lazily via the SLRU slot manager (or runs them on CPU), so the greedy packer stopped placing trunk layers far too early.

- Reuse the runtime allocator's own `CudaForwardPass.EstimateKvCacheBytes` (now gpuLayers-scoped) for per-layer-shape models, with a monotonic binary search (`SolveGpuCtxForPerLayerKv`) for the inverse context solve — planner and allocator can no longer drift. Uniform models keep the flat formula, dtype-scaled via `DTypeInfo.ByteSize` (byte-identical at fp32).
- New `kvDtype` param on `Plan`, threaded through the CUDA dense/hybrid call sites only (Vulkan/GDN/bench stay fp32).
- `MeasureLayerBytes` excludes routed experts (`ffn_*_exps`) for MoE; leftover VRAM after trunk+KV reported as `LayerPlacement.ExpertCacheBudgetBytes`.

## #215 — CPU-MoE execution mode for CudaHybridForwardPass (compute-half)
`CudaHybridForwardPass` (pure-attention MoE hybrid — Qwen3-Coder-30B-A3B, OLMoE, qwen2moe-class) gains a CPU-MoE mode mirroring the GDN sibling: GPU runs attention/norms (+shared expert when present), the post-FFN-norm hidden is downloaded, the MoE router + routed experts run on the CPU, result uploaded back. At a 12 GB profile the SLRU expert cache is only ~13% resident on Qwen3-Coder-30B, so miss-driven PCIe traffic dominates the GPU-SLRU path; running routed experts on the CPU is faster (the #100 precedent).

- New `_cpuMoe` auto-router (`SHARPI_CPU_MOE=0|1` override, else `predictedSlots/totalExperts < 0.5`), predicted from the planner's `ExpertCacheBudgetBytes`.
- `GpuTrunkCpuMoeFfn`: GPU shexp kick → `Download` → `CpuMoeRouted` (factored out of the existing `CpuMoeFfn` — no new dot kernel) → `UploadInto` → GPU shexp add. Dispatched in both decode and per-token prefill.
- Fixes a `-g -1` auto-select regression the #215 planner-half introduced: excluding routed experts from the trunk budget made `TierPlanner` return `GpuLayers == NumLayers` for big MoE, so the CLI `wantHybrid` gate went false and Qwen3-Coder-30B fell through to full-offload `CudaForwardPass` (the silent-OOM/thrash case TierPlanner exists to avoid). New `LayerPlacement.MoeRoutedExpertBytes`; RunCommand routes auto MoE to hybrid when that exceeds the expert-cache budget. OLMoE (experts fit) stays full-GPU.

## Validation
- `CudaHybridForwardPass_CpuMoeMode_MatchesCpuReference`: top-5 logit overlap vs the independent CPU `ForwardPass` (strict argmax parity is unreliable for CUDA MoE — Q4_K activation quant flips the router top-K near the boundary). OLMoE: 4/5 overlap, top-1 match, maxAbs 1.61.
- `CudaForwardPassKvDtypeTests`: gpuLayers scoping + binary-search contract (fits / never under-reserves / ≥ reference / maximal / floors at 512).
- A/B on Qwen3-Coder-30B-A3B @ 12 GB: decode 18.6→25.0 t/s (1.34×), prefill 14.2→20.8 t/s (1.46×).
- Full Release build clean under `TreatWarningsAsErrors`; forward-pass + planner tests green.

## Follow-ups (separate issues)
The #215 commit message lists "batched CPU-MoE prefill / Q8_K-input routed kernels" as follow-ups. These were **investigated this session and measured to regress** the Q4_K targets (the per-token sequential core is the optimum); issues will be filed documenting the negative results so they aren't re-attempted blindly.

### DIFF
--- a/src/SharpInference.Cli/RunCommand.cs
+++ b/src/SharpInference.Cli/RunCommand.cs
@@ -447,7 +447,7 @@ public sealed class RunCommand : Command<RunCommand.Settings>
                     var hwProfile = HardwareProfile.Detect(cuda);
                     AnsiConsole.MarkupLine($"[dim]Hardware: {hwProfile.Summary()}[/]");
                     var placement = TierPlanner.Plan(model, hp, hwProfile, settings.TurboQuant,
-                        requestedCtxSize: ctxSize);
+                        requestedCtxSize: ctxSize, kvDtype: CudaForwardPass.ResolveConfiguredKvDType());
                     cudaGpuLayers = placement.GpuLayers;
 
                     // Gemma 4 KV-share constraint: the shared-KV source layers (E4B:
@@ -487,7 +487,7 @@ public sealed class RunCommand : Command<RunCommand.Settings>
                 {
                     var hwProfile = HardwareProfile.Detect(cuda);
                     var placement = TierPlanner.Plan(model, hp, hwProfile, settings.TurboQuant,
-                        requestedCtxSize: ctxSize);
+                        requestedCtxSize: ctxSize, kvDtype: CudaForwardPass.ResolveConfiguredKvDType());
                     if (nGpuLayers != -1)
                         placement = placement with { GpuLayers = cudaGpuLayers, CpuLayers = hp.NumLayers - cudaGpuLayers };
 
--- a/src/SharpInference.Cli/RunCommand.cs
+++ b/src/SharpInference.Cli/RunCommand.cs
@@ -442,6 +442,7 @@ public sealed class RunCommand : Command<RunCommand.Settings>
                 // a model bigger than VRAM (e.g. Qwen3-Coder 30B-A3B in 12 GB) would
                 // attempt full-offload via CudaForwardPass and silently OOM.
                 int cudaGpuLayers;
+                bool moeAutoNeedsHybrid = false;
                 if (nGpuLayers == -1)
                 {
                     var hwProfile = HardwareProfile.Detect(cuda);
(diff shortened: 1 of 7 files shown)


pekkah:
Review cycle (high-effort, 3 finder angles + verification): **no correctness or crash bugs.**

Refuted candidates: the `TierPlanner.Plan(kvDtype:)` "missing at Vulkan/bench/test call sites" cluster is by-design (only CUDA dense/hybrid thread `kvDtype`; Vulkan/GDN/bench use fp32 KV — verified CLI 450/506 and server 213/239 pass it); `MoeRoutedExpertBytes` summing all layers is correct in its only use (gated on full GPU offload).

Surviving (low-severity, non-blocking):
- `ExpertCacheBudgetBytes` is stale after a `with { GpuLayers = N }` override on explicit `-g N` → may mis-select CPU-MoE vs GPU-SLRU (perf only, both modes correct, self-correcting via `SHARPI_CPU_MOE`). Filed as #224.
- Minor: `TierPlanner.Plan` runs twice in the auto MoE-hybrid path (pre-existing pattern).

`#215` batched-MoE follow-ups were investigated and measured to regress (per-token core is the optimum) — documented in #225. Merging.

PR #226, opened 2026-06-13
perf(cuda): batched-trunk prefill for CPU-resident embedding (#218)

### LINKED ISSUE #218: perf(cuda): batched-trunk prefill falls back to per-token when the embedding is CPU-resident — exactly the low-VRAM config (#123 follow-up)
## Problem

The #123 batched-trunk prefill in `CudaHybridForwardPass` is gated OUT when the embedding table stays on the CPU:

- `IsBatchedPrefillSupported` requires `_gpuEmbedding is not null` (`CudaHybridForwardPass.cs:1020-1027`).
- The doc comment (`:1013-1017`) explains why: the per-token CPU-embed path stages each row through the single shared `_pinnedHidden` buffer, and a batched embed loop would race write-after-read on it without draining the stream per token.
- The fallback is the full per-token `Forward` loop (`:1001-1003`), which pays the pinned map/copy/unmap + `EndRecordAndSubmit` boundary cost **per prompt token** (`:848-970`) — the exact per-position launch cascade #123 was built to collapse.

`ShouldKeepFixedWeightsOnCpu` / the tier planner push the embedding to CPU precisely on low-VRAM cards, so on the target 12 GB machine the batched prefill is effectively never active for this class, while a 16 GB+ card gets it for free.

## Proposed work

1. **Batched CPU embed + single upload:** dequantize the chunk's N embedding rows on the CPU into an N×embDim staging block (a dedicated pinned buffer or a ping-pong pair, NOT the shared `_pinnedHidden`), then one async H2D into the batched trunk's hidden buffer. This removes the WAR hazard by construction — no per-token drains.
2. Drop the `_gpuEmbedding is not null` term from `IsBatchedPrefillSupported`; keep the other gates (Gemma4-like, TurboQuant, non-NEOX RoPE, L2 QK-norm, non-batchable trunk dtypes) as-is.
3. Keep the `_faulted` latch + bit-exactness contract: the batched embed must produce byte-identical rows to `CpuEmbedToken` (same dequant routine, just batched).

## Acceptance

- [ ] `BatchedPrefill_BitwiseMatchesSequential`-style oracle for a CPU-resident-embedding config (force via the 2 GB storage-buffer rule or a test hook), single + multi-chunk.
- [ ] Prefill A/B on a 12 GB-like profile (CPU embedding + CPU layers &gt; 0): expect the same order-of-magnitude win #123 showed for the GPU-embedding config.
- [ ] GPU-embedding configs unchanged (same kernels, same parity tests green).

## References

- #123 (the batched-trunk prefill this gates), #176 (prefill perf context), #215 (placement — more trunk on GPU makes this matter more).

pekkah:
## Summary

Closes #218. The #123 batched-trunk prefill in `CudaHybridForwardPass` was gated out when the embedding table is CPU-resident (`_gpuEmbedding is null`) — **exactly the low-VRAM / 12 GB-default config** the tier planner produces. Those models fell back to the per-token `Forward` loop and paid the pinned map/copy/unmap + `EndRecordAndSubmit` launch cascade **per prompt token**, while a 16 GB+ card got the batched path for free.

The gate existed to avoid a write-after-read hazard: a naive batched embed loop reusing the shared `_pinnedHidden` buffer would race each token's CPU write against the prior token's still-in-flight async H2D read.

## Fix

When the embedding is CPU-resident, dequantize **all N rows** on the CPU (same `CpuEmbedToken` per-row dequant → byte-identical to the per-token path) into a **dedicated** pinned block `_pinnedEmbedAll` `[N × embDim]`, then upload the whole block in **one** async H2D (`CopyDevice`, UVA pinned→device) into the batched trunk's stream. All CPU writes complete before the single copy is enqueued, so the WAR hazard is removed by construction — no per-token drains.

- Dropped `_gpuEmbedding is not null` from `IsBatchedPrefillSupported`; every other gate (Gemma4-like, TurboQuant, non-NEOX RoPE, L2 QK-norm, attn-bias, non-batchable trunk dtypes) is unchanged.
- The `_faulted` latch + bit-exactness contract hold.
- Test/bench hook `ForceCpuResidentEmbedding` (env `SHARPI_FORCE_CPU_EMBED=1`) forces the CPU-resident config even where the embedding would fit in VRAM.

## Perf

Qwen3-Coder-30B-A3B Q4_K_M, RTX 4070 Ti 12 GB, 48 GPU / CPU-MoE, N=512 prefill:

| Config | tok/s |
|---|---|
| CPU-embed, batched-OFF (pre-#218 fallback) | 18.5 |
| **CPU-embed, batched-ON (#218)** | **25.1** (1.35×) |
| GPU-embed, batched-ON (what 16 GB+ got free) | 24.9 |

#218 lands the 12 GB config at **101% of the GPU-embed batched throughput** — i.e. it now matches what 16 GB+ got for free. The 1.35× ceiling here is the per-token CPU-MoE (identical in both arms); #218 collapses the embed + attention-launch cascade onto the GPU-embed curve.

## Tests (`Tests.ForwardPass` / `CudaHybridBatchedPrefillTests`)

New bit-parity oracles, all bit-identical to the sequential `Forward` loop:
- `BatchedPrefill_CpuEmbedding_BitwiseMatchesSequential_Coder` — single-chunk
- `BatchedPrefill_CpuEmbedding_MultiChunk_BitwiseMatchesSequential_Coder` — `startPos>0` + pinned-buffer reuse
- `BatchedPrefill_CpuEmbedding_CpuLayersSplit_BitwiseMatchesSequential_Coder` — synthesized `CpuLayers>0` split (the planner gives 48/0 here, so this deterministically exercises the CPU-embed staging + per-token CPU-layer loop together)
- `BatchedPrefill_GpuEmbedding_CpuLayersSplit_BitwiseMatchesSequential_Coder` — closes a pre-existing #123 gap: GPU-embed + `CpuLayers>0` had no committed guard on this box

Also fixed stale `~29 GPU + 19 CPU` split comments (the planner gives 48/0 via CPU-MoE). Existing GPU-embed oracles unchanged; full class 9/9 green. Solution builds clean under TreatWarningsAsErrors (Debug + Release).

### DIFF
--- a/src/SharpInference.Engine/CudaHybridForwardPass.cs
+++ b/src/SharpInference.Engine/CudaHybridForwardPass.cs
@@ -182,8 +182,16 @@ public sealed unsafe class CudaHybridForwardPass : IForwardPass
     // Test observable (issue #123): set by Prefill to whether the last call dispatched the
     // batched-trunk path (true) or fell back to the per-token loop (false). Lets the parity
     // oracles assert the batched path actually ran instead of passing vacuously when the
-    // config is gated out (e.g. CPU-resident embedding / non-batchable dtype on a given box).
+    // config is gated out (e.g. non-batchable dtype on a given box).
     internal bool LastPrefillWasBatched;
+
+    // Issue #218 test/bench hook: force the low-VRAM "fixed weights on CPU" config
+    // (embedding + output table CPU-resident, _gpuEmbedding/_gpuOutputWeight null) even on a
+    // box where ShouldKeepFixedWeightsOnCpu would keep them on the GPU. Lets the batched-CPU-
+    // embed prefill path (#218) be exercised and A/B-benched on a model whose embedding fits in
+    // VRAM. Read once in the constructor; OR'd into cpuEmbeddingOutputOnly. Default off.
+    internal static bool ForceCpuResidentEmbedding =
+        Environment.GetEnvironmentVariable("SHARPI_FORCE_CPU_EMBED") == "1";
 
     // Device scratch for the batched attention trunk, allocated lazily by
     // EnsureBatchedTrunkScratch(N) and sized to the largest N seen so far. The
@@ -201,6 +209,13 @@ public sealed unsafe class CudaHybridForwardPass : IForwardPass
     // Pinned host buffer for the batched GPU→CPU hidden transfer [N × embDim].
     private Tensor? _pinnedHiddenAll;
     private int _pinnedHiddenAllN;
+    // Issue #218: dedicated pinned host buffer for the batched CPU→GPU embedding upload
+    // [N × embDim], used only when the embedding is CPU-resident (_gpuEmbedding is null).
+    // Distinct from _pinnedHidden/_pinnedHiddenAll so the embed staging never races the
+    // per-token hidden transfer; the whole block is CPU-written then uploaded in one async
+    // H2D, so there is no per-token write-after-read on a shared buffer.
+    private Tensor? _pinnedEmbedAll;
+    private int _pinnedEmbedAllN;
 
     // Non-transactional fault latch (mirror of CudaHybridGdnForwardPass._faulted):
     // set at batched-prefill entry, cleared only after the KV/position counters have
@@ -276,7 +291,7 @@ public sealed unsafe class CudaHybridForwardPass : IForwardPass
         if (_isGemma4Like)
             Gemma4ValidateHybridSplit(hp, _nGpuLayers);
 
-        bool cpuEmbeddingOutputOnly = ShouldKeepFixedWeightsOnCpu(
+        bool cpuEmbeddingOutputOnly = ForceCpuResidentEmbedding || ShouldKeepFixedWeightsOnCpu(
             model.FindTensor("token_embd.weight")!.Value,
             model.FindTensor("output.weight"));
         _tqEnabled = enableTq;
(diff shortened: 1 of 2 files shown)


pekkah:
## Review cycle

Ran three independent review lenses on the diff (general correctness/CLAUDE.md, silent-failure/error-handling, test-coverage).

**Findings & resolution:**
- **Correctness / silent-failure**: no issues at or above threshold. Confirmed: `_pinnedEmbedAll`↔`stream` element-count match, UVA pinned→device `CopyDevice` semantics (same proven idiom as the existing GPU→CPU copy), the WAR-hazard removal, the `_faulted` latch held across the new branch, no leak/double-free, the removed `throw` replaced by a complete fail-fast branch, dropping the `_gpuEmbedding` gate lets no unsupported config through.
- **Test coverage — addressed**: added `BatchedPrefill_CpuEmbedding_ComputeRouting_ArgmaxStableVsMatvec_Coder` (`99f1b06`) — the combination that actually ships by default on a 12 GB box (CPU-resident embedding + compute-routed MMQ/GEMM projections) was previously asserted only by inference. Now argmax-stable-verified (greedy=2585, maxAbs=2.28). Also already added the GPU-embed `CpuLayers>0` split oracle (`458bb57`) closing a pre-existing #123 gap.

**Documented follow-up (not in this PR):** the `_faulted` non-transactional latch (introduced in #123, shared with `CudaHybridGdnForwardPass`) has no fault-injection coverage anywhere. The only way to force a throw inside `PrefillBatchedTrunk` step 1 is an out-of-range token id → that's an OOB read (UB / likely AV), not a clean managed throw, so a faithful test needs a dedicated throw seam. Out of scope here; worth a small follow-up with a test hook.

**Final gate:** `CudaHybridBatchedPrefillTests` 10/10 green (incl. the >4096 wave path); solution builds clean under TreatWarningsAsErrors (Debug + Release).

Issue #228, opened 2026-06-13
perf(cuda): EstimateAvailableKvVram reserves max(VRAM/3, 2GB) — far over-conservative, auto-context lands ~8× below what fits (Gemma 4 12B q8_0: 30K auto vs 256K real)


pekkah:
## Problem

The auto-context the dense CUDA path picks (`-g -1`, no `-c`) is much smaller than what actually fits. On a 4070 Ti 12 GB with `gemma-4-12b-it-qat-q4_0` + `--kv-type q8_0`:

| | ctx |
|---|---|
| auto (`-g -1 --kv-type q8_0`) | **30,840** |
| forced (`-c 262144 --kv-type q8_0`, = model max) | **262,144** — constructs + runs a token |

So the full 256K q8_0 KV cache fits, but auto picks ~8× less. (Found while verifying #220, which fixed a *separate* bug — auto ignoring `--kv-type` entirely.)

## Root cause: a blunt, over-sized VRAM reserve

`CudaForwardPass.EstimateAvailableKvVram` (`CudaForwardPass.cs:3946-3982`) computes the KV budget as `VRAM − estimatedWeights − scratch − reserve`, where:

```
long reserved = Math.Max(vramBytes / 3, 2L * 1024 * 1024 * 1024);   // :3978
```

On a 12 GB card that reserve is **4094 MiB** — and it's the dominant term. A `SHARPI_TRACE_VRAM=1` run shows the reality:

```
constructor entry:        free = 10854 MiB   (~1.4 GB already used by the CUDA context)
before per-layer upload:  free =  9316 MiB
after ALL weight uploads: free =  2061 MiB   ← actual headroom for KV
```

The whole 256K q8_0 KV cache fits in that 2061 MiB (Gemma's ~5:1 SWA layers cap at their ~5K ring, so only the few global layers scale with context — cheap at q8_0). But the auto formula subtracts a 4 GB reserve on top of the weight estimate, leaving only ~1.3 GB of *computed* budget → 30,840. It's reserving ~2.5–3 GB it doesn't use.

## Why the reserve is load-bearing (do NOT just delete it)

[...]

PR #229, opened 2026-06-13
perf(cuda): dtype-aware auto-context for Gemma 4 (#220)

### LINKED ISSUE #220: perf(engine): TierPlanner prices Gemma 4 KV at uniform full-context fp32 — runtime uses SWA rings, KV-share aliasing and per-layer head_dim, so auto -g/ctx underestimates capacity ~4-6×

pekkah:
## Summary

Fixes #220. On a 12 GB card, Gemma 4 12B's auto-context (`-g -1`, no `-c`) was pinned at **1770 tokens for fp32 AND bf16 AND q8_0** — `--kv-type` bought no extra context, despite the model supporting 262144.

The dense full-GPU path's auto-context comes from `CudaForwardPass.EstimateMaxContext`, which already had a Gemma SWA / KV-share / per-layer-headdim branch but **hardcoded `sizeof(float)`** (and used `NumKvHeads`, no pow2 round-up) — so it was dtype-blind and could drift from the real allocator. (`TierPlanner.RecommendedCtxSize` is computed but discarded for the full-GPU case; the live source is `EstimateMaxContext`.)

## Fix

Replace `EstimateMaxContext`'s hand-rolled SWA math with a binary search against `EstimateKvCacheBytes(hp, ctx, kvDType)` — the single source of truth the ctor's allocation and `TierPlanner.SolveGpuCtxForPerLayerKv` both use (dtype + SWA ring + KV-share skip + per-layer KV heads + pow2 round-up). The core is factored into the pure, GPU/GGUF-free `CudaForwardPass.SolveMaxCtxForKv` (mirrors the `ResolveKvDType` factoring); the call site passes the requested KV dtype (`ResolveConfiguredKvDType()`). **Uniform-attention models keep the flat fp32 formula unchanged.**

## Result (4070 Ti 12 GB, `gemma-4-12b-it-qat-q4_0`, `-g -1`)

| `--kv-type` | before | after |
|---|---|---|
| fp32 | 1770 | 2048 (allocator-exact) |
| bf16 | 1770 | **4096** (2×) |
| q8_0 | 1770 | **30840** (17×) |

q8_0 exceeds 4× because SWA layers stop growing past their ring cap, so beyond it only the global layers consume KV per token. Verified the q8_0 30840-ctx run constructs + decodes coherently (53 t/s, no OOM); Qwen3-8B dense auto-ctx unchanged (12080).

## Tests

3 new GGUF-free unit tests for `SolveMaxCtxForKv` (`CudaForwardPassKvDtypeTests`):
- dtype response in the linear regime — bf16 = 2× fp32, q8_0 ~3.76× (q8_0's 34/32-byte blocks fall short of a clean 4×);
- allocator-maximal on the SWA shape (fits, never under-reserves, maximal);
- uniform models ignore dtype (the "no change for uniform models" guard).

Existing KV-dtype parity + Gemma coherence + SWA ring-wrap tests stay green (they pass explicit ctx, so are unaffected). Release build clean under TreatWarningsAsErrors + AOT analyzers.

## Follow-up (separate)

The auto value (30840) is still ~8× below what actually fits (full 256K runs with `-c`), because `EstimateAvailableKvVram` reserves a blunt `max(VRAM/3, 2GB)`. That's tracked in **#228** (gated on a spill-cliff A/B) — deliberately out of scope here.

### DIFF
--- a/src/SharpInference.Engine/CudaForwardPass.cs
+++ b/src/SharpInference.Engine/CudaForwardPass.cs
@@ -4103,12 +4103,18 @@ public sealed unsafe class CudaForwardPass : IForwardPass, IBatchedForwardPass
         const int floorCtx = 512;
         int cap = hp.ContextLength;
 
+        // A model whose context is at/below the floor clamps to the cap (and avoids the
+        // Math.Clamp(_, 512, cap) below throwing when cap < 512). Mirrors the floor convention:
+        // return the small ctx and let the ctor's allocation fail loudly if even that won't fit.
+        if (cap <= floorCtx)
+            return cap;
+
         if (hp.LayerHeadDim is not null && hp.IsSwaLayer is not null)
         {
             // EstimateKvCacheBytes is monotonic non-decreasing in ctx, so an upper-bound binary
             // search converges. Floor 512 mirrors the uniform clamp below.
-            if (cap <= floorCtx || EstimateKvCacheBytes(hp, floorCtx, kvDType) > availableKvBytes)
-                return Math.Min(floorCtx, cap);
+            if (EstimateKvCacheBytes(hp, floorCtx, kvDType) > availableKvBytes)
+                return floorCtx;
             int lo = floorCtx, hi = cap;
             while (lo < hi)
             {
--- a/src/SharpInference.Engine/CudaForwardPass.cs
+++ b/src/SharpInference.Engine/CudaForwardPass.cs
@@ -566,7 +566,14 @@ public sealed unsafe class CudaForwardPass : IForwardPass, IBatchedForwardPass
         else if (_tqEnabled)
             _maxSeqLen = EstimateMaxContextTq(model, gpu, hp, tqFp32Window, tqBits);
         else
-            _maxSeqLen = EstimateMaxContext(model, gpu, hp);
+            // Size the auto-context for the KV dtype the operator requested (#220): a bf16/q8_0
+            // KV store fits 2×/4× the positions of fp32, but EstimateMaxContext previously priced
+            // fp32 unconditionally, so --kv-type was silently ignored for auto-context. Pass the
+            // *requested* dtype (pre-auto-narrow): an explicit narrowed choice should expand the
+            // window, and the fp32 default still yields the fp32-fit context (the auto-narrow
+            // below never fires for an fp32-sized auto-context, since fp32 fits there by
+            // construction).
+            _maxSeqLen = EstimateMaxContext(model, gpu, hp, ResolveConfiguredKvDType());
         // The KV-append/attention kernels index the cache at `pos % _maxSeqLen` (the ring
         // modulo, identity for full caches), so a zero context — e.g. a malformed GGUF with
         // context_length=0 reached via an explicit ctx-size — would be an in-kernel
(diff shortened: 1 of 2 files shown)


pekkah:
## Review cycle complete

Ran three independent review lenses (correctness, silent-failure/OOM, test-coverage) plus the automated gemini-code-assist review.

**Correctness + silent-failure:** no high-confidence issues. Confirmed the binary search is allocator-exact and **strictly safer than master** on OOM — master's old Gemma formula summed raw bytes with no pow2 round-up and could hand back a context that cudaMalloc-fails; the new search rounds each buffer exactly like the pool. Requested-vs-actual KV dtype can't diverge to under-reserve (auto-narrow only narrows, never widens). No integer overflow (long arithmetic; the `(int)` cast is only on the unchanged uniform path).

**Addressed findings:**
- **gemini (HIGH):** `Math.Clamp(_, 512, cap)` in the uniform branch threw `ArgumentException` for `ContextLength < 512`. Fixed with a top-level `if (cap <= floorCtx) return cap;` guard (95b5d27) — both branches now safe + consistent; added a uniform-shape `cap<512` test.
- **pr-test-analyzer (2× must-fix):** the SWA-saturation super-linear claim and the floor/cap clamp had no coverage. Added `SolveMaxCtxForKv_SwaSaturation_DtypeGainExceedsWidthRatio`, `_ClampsToFloorAndModelMax`, `_MixedPerLayerHeadDim_IsAllocatorMaximal` (1f73171).

Final: 6/6 `SolveMaxCtxForKv` unit tests green, existing KV-dtype/coherence/ring tests green, Release build clean under TreatWarningsAsErrors + AOT analyzers. Merging.

Issue #235, opened 2026-06-13
perf(cuda): flash-decoding (split-KV) for long-context decode attention — 8-block launch is occupancy-bound (~6% of HBM peak)


pekkah:
## Problem

Single-user decode throughput on the CUDA path collapses as context grows — ~**49→12.7 tok/s from 2K→32K** on Gemma 4 E4B (RTX 4070 Ti), a ~3.9× drop. This is **dtype-independent**: fp32, bf16, and q8_0 all degrade with the same curve (after #234 closed the q8-specific dequant overhead, q8 now tracks fp32). #213 documented this as "inherent KV bandwidth" — that was only half right.

## Root cause (profiled): the decode-attention kernel is occupancy-bound, not bandwidth-bound

Every decode-attention kernel (`llm_attention`, `llm_attention_bf16`, `llm_attention_q8_0`, and the `_swa` variants) launches with:

```
grid  = (numHeads, 1, 1)
block = 256
```

(see `CudaBackend.cs:3657, 3944, 4224, 4274, 4620, 4665`)

At decode (1 query token), that is **`numHeads` blocks total** — **8 blocks for E4B** (`gemma4.attention.head_count = 8`). Each block serially scans its head's *entire* KV sequence. Consequences:

- Only **8 of 60 SMs** are touched (~13% of the GPU). The **sequence dimension has zero parallelism** — as ctx grows, each block's serial scan just gets longer, with nothing to hide the latency.
- Achieved KV-read bandwidth at 16K: ~1.0 GB/token of KV ÷ ~36 ms (the Phase-1 attention-phase time) ≈ **~28 GB/s**, vs the 4070 Ti's ~500 GB/s HBM peak → **~6% of peak**.

So long-ctx decode attention is **latency/occupancy-bound**. There is ~10× of headroom the 8-block launch leaves on the floor. (The weight matvecs — qkv/ffn — are a separate, already-saturated ~90%-of-peak phase and are dtype-independent; they are *not* the long-ctx problem.)

### E4B geometry (for reference)
- head_count 8, head_count_kv 2 (GQA 4:1), head_dim 512 (global) / 256 (SWA)
- 42 layers, sliding_window 512, ~5:1 SWA:global pattern → ~7 global layers (these grow to full ctx and dominate the long-ctx KV read; SWA layers stay window-bounded)
- shared_kv_layers 18 (cross-layer KV sharing — reduces distinct KV but does not change the parallelism story)

[...]

PR #236, opened 2026-06-13
perf(cuda): flash-decoding (split-KV) for long-context decode attention (#235)

pekkah:
## What

Flash-decoding (split-KV) for long-context decode attention on the dense CUDA path (#235). Fixes the decode-attention occupancy collapse profiled in Step 0.

## Why

The decode-attention kernels launch `grid=(numHeads,1,1)` — **8 blocks for Gemma 4 E4B** (~13% of the 60 SMs) — and each block **serially scans the whole KV sequence**. The sequence dimension has zero parallelism, so as context grows each block's O(ctx)/token scan just lengthens while the GPU idles: ~**6% of HBM peak** at 16K. This is the root of the long-context decode collapse from #213 (49→12.7 t/s, 2K→32K, dtype-independent).

## How

Flash-decoding (Tri Dao 2023): partition each head's sequence into fixed 512-token chunks across `numHeads × nSplits` blocks; each emits the un-normalized online-softmax partial `(m_i, l_i, Õ_i)` for its chunk; a combine kernel LSE-merges the partials per head.

- **Scalar, not tensor-core.** At GQA ratio 4 (E4B) the Step-0 research and llama.cpp both stay scalar: query-head packing only wins at G≥8, and at G=4 a 16-row mma tile is ≥75% padding while still memory-bound. Full prior-art writeup is posted on #235.
- **One templated `llm_attention_splitkv<KV>`** (fp32/bf16/q8_0 thunks) reusing `sharpi_kv_dot`/`sharpi_kvload`; plus a dtype-agnostic `llm_attention_combine`.
- **CUDA-graph-compatible.** Fixed grid (`nSplits = ceil(maxSeqLen/512)`) with `seqLen` as the only per-replay-updated param and early-exit for out-of-range splits — slots into the existing `TrackPositionNode`/`PositionPlus1` mechanism (combine has no varying args). Out-of-range splits write `(m=−inf, l=0)` so a stale partial is never merged.
- **Gated** in `CudaForwardPass.AttentionKv`: split when `seqLen > 2048` (below that the single-block kernel is already near the weight-matvec floor) and `maxSeqLen ≤ 131072` (`nSplits ≤ 256` fits the combine's shared array). Global layers only; SWA stays single-block (window-bounded KV). `SHARPI_SPLIT_DECODE=0` kill-switch.

## Result — decode A/B (Gemma 4 E4B Q8_0, RTX 4070 Ti, 32 timed steps after a long prefill)

| ctx | fp32 off→on | q8 off→on |
|---|---|---|
| 8192 | 31.4 → **52.9** (1.68×) | 29.1 → **55.2** (1.90×) |
| 32768 | 12.8 → **29.7** (2.32×) | 11.3 → **45.2** (4.0×) |

The long-context curve flattens substantially. And the #213 insight is now fully realized: once occupancy is fixed the kernel is **bandwidth-bound**, so q8 (¼ the bytes) now **beats** fp32 at 32K — the inverse of the single-block kernel where q8 was the slowest dtype.

## Validation

- **4 new split-vs-single-block parity tests** (`CudaForwardPassKvDtypeTests`): fp32/bf16/q8 on E4B + fp32 on Qwen3-8B (different geometry — 32 heads/d128 all-global vs 8 heads/d512 SWA+global). A 5000-token prompt puts every decode step at `seqLen > 2048` → the split path; the single-block run (`SHARPI_SPLIT_DECODE=0`) is the trusted reference. Argmax-stable with tight logit budgets (≤1.0 fp32, ≤2.0 bf16, ≤2.5 q8); greedy `firstTok` identical split-on vs off.
- **Full KvDtype suite 59/59**, **Gemma4 graph-parity + dense forward-pass 8/8**, attn-kernel NVRTC compile 5/5.
- Combine reorders the softmax reduction → **argmax-stable, not bit-identical** (same contract as the TC2 flash-prefill path).
- Release build clean (`TreatWarningsAsErrors` + NativeAOT analyzers).

## Scope / follow-ups

Dense `CudaForwardPass`, global layers. Noted Phase-B follow-ups (in #235): in-kernel GQA group-shared KV loading (load each KV head once across its query group), SWA split, and the hybrid/GDN paths.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,11 +13,10 @@ OpenBLAS in `tools/openblas/` for faster batched GEMM. Build with `dotnet build 
 
 Supported architectures: `llama`, `llama4`, `olmoe`, `qwen3`, `qwen3moe`, `qwen35moe`
 (hybrid Gated-DeltaNet + attention + MoE), `gemma4` (per-layer head_dim, SWA + global, PLE).
-Benchmarked on AMD Zen 4 (12c/24t, DDR4-3200) + RTX 4070 Ti (12 GB), Q4_K_M, `--temp 0`.
-**Prefill t/s** is the warm-cache rate at a ~1K-token prompt; **decode t/s** is the near-zero-ctx
-generation rate (forward-pass iterations / time, so thinking tokens count). All outputs verified
-coherent (`scripts/bench-all.ps1`); top-1 parity vs llama.cpp b8585 verified on Qwen3-8B (byte-identical
-60-token greedy decode).
+Benchmarked on AMD Zen 4 (12c/24t) + RTX 4070 Ti (12 GB), Q4_K_M, `--temp 0`. Prefill t/s is the
+warm-cache rate at a ~1K-token prompt; decode t/s is near-zero-ctx (forward-pass iterations / time, so
+thinking tokens count). Outputs verified coherent; Qwen3-8B is byte-identical to llama.cpp b8585 (60-token
+greedy decode).
 
 | Model | Repo | Size | Backend | Prefill t/s | Decode t/s | Notes |
 |---|---|---:|---|---:|---:|---|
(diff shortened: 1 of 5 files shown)


pekkah:
Addressed the 3-lens + gemini review in `caa244c`:

**gemini HIGH** (nSplits > 256 → combine shared-array overrun): added a `nSplits ≤ SplitKvMaxSplits (256)` guard in `AttentionSplitKv` that throws, and tied `SplitKvMaxCtx = SplitKvMaxSplits × SplitKvChunk` so the host gate can't drift from the kernel's bound.

**gemini MEDIUM ×2** (hoist invariant offsets): hoisted the KV base offset out of the splitkv Phase 1/3 inner loops and the combine d-loop. Bit-identical (same offsets, same read order) — confirmed by the chunked-prefill tests (which use the split decode path) staying green.

**silent-failure-hunter LOW** (TQ wasted VRAM): the partials allocation now skips `_tqEnabled` runs — TQ decode uses the direct `TqAttention`/`Attention` path, never `AttentionKv`, so the buffer was never used.

**silent-failure-hunter HIGH** (perf, not correctness): a graph captured *below* `SplitKvMinSeq` bakes the single-block node, so short-prompt/long-generation stays single-block (correct, just unaccelerated). The primary case (#213/#235 = long prompts) captures above the threshold and accelerates. Documented at the gate; re-capture-on-crossing is a noted #235 Phase-B follow-up.

**test coverage** (pr-test-analyzer): added two tests —
- `Gemma4_E4B_SplitKv_GraphReplayCrossesBoundary`: asserts `GraphReady` (so it can't vacuously pass on a direct-launch fallback) and decodes across a 512-token split boundary under graph replay — exercising the out-of-range→active split transition the steady-state parity tests don't hit. Gates on top-5 stability + a blow-up ceiling (the tight fp32 logit bound stays the parity test's job).
- `Gemma4_12B_Q8SplitKv_MatchesSingleBlock`: the `k_eq_v` / per-layer-KV global geometry (riskiest untested path).

Final validation: 6 split-KV parity tests + 4 chunked-prefill green on the hoisted kernel; the earlier full KvDtype (59) + Gemma4 graph-parity/dense (8) runs cover the unchanged paths.

Issue #238, opened 2026-06-13
perf(cuda): extend flash-decoding split-KV to the hybrid forward passes (MoE / GDN)


pekkah:
## Context

Follow-up to #235 (flash-decoding split-KV, Phase A — shipped in PR #236). The split-KV decode path is wired into the **dense** `CudaForwardPass.AttentionKv` only. The hybrid forward passes still launch the single-block-per-head decode attention, so their GPU-resident attention layers hit the same occupancy collapse at long context.

## Problem

`CudaHybridForwardPass` (layer-split MoE: Qwen3-Coder-30B, OLMoE) and `CudaHybridGdnForwardPass` (GDN hybrid: Qwen3.6-35B/27B) don't benefit from split-KV. Their GPU attention layers' per-token KV read is still serialized across `numHeads` blocks → long-ctx decode degrades there too.

## Proposed change

Wire the existing `CudaBackend.AttentionSplitKv` (+ partials buffer + the `seqLen > 2048` gate + `SHARPI_SPLIT_DECODE` switch) into the hybrid passes' attention decode for the GPU-resident attention layers — the same change made to `CudaForwardPass.AttentionKv`, reusing the validated kernels. If #237 (GQA head-sharing) lands first, the hybrid path inherits it automatically.

## Measurement gate

Hybrid decode is often **MoE/expert-bound**, not attention-bound (the routed experts dominate the step). So profile (`SHARPI_CUDA_PROFILE`) whether attention is a meaningful share of hybrid decode at long ctx **before** claiming a win — the speedup is bounded by attention's fraction of the step. Decode A/B (split on/off) at long ctx per model; merge only if it moves the needle. The GDN hybrid disables CUDA graphs on the batched-decode path, so confirm the split path's eager launch is correct there (no graph dependency).

## Scope

- [ ] Allocate the split partials + wire `AttentionSplitKv` into the hybrid attention decode dispatch (both `CudaHybridForwardPass` and `CudaHybridGdnForwardPass`), gated like the dense path.
- [ ] Confirm graph interaction per pass (CudaHybridForwardPass uses graphs; the GDN batched path doesn't).
- [ ] Per-model long-ctx decode A/B (Coder-30B, OLMoE, Qwen3.6-35B/27B) — keep only where attention is a real share.

## Validation

[...]

Issue #237, opened 2026-06-13
perf(cuda): GQA head-sharing in split-KV decode attention — load each KV head once across its query group


pekkah:
## Context

Follow-up to #235 (flash-decoding split-KV, Phase A — shipped in PR #236). The split kernel `llm_attention_splitkv<KV>` fixed the decode-attention occupancy collapse and made the kernel **bandwidth-bound** at long context. The next bandwidth lever is GQA head-sharing.

## Problem

The split kernel runs **one block per (query-head, KV-split)**. With GQA, the `G = numHeads / numKvHeads` query heads in a group share one KV head — but they're `G` separate blocks, each re-reading the *same* KV slice. At long context the KV exceeds L2 (~48 MB on the 4070 Ti), so those become **G× redundant HBM reads** of K and V. Since the kernel is now bandwidth-bound, that redundancy directly inflates decode time. E4B is `G = 4` (8 query heads / 2 KV heads); Qwen3-8B is also 4 (32/8).

## Proposed change

One block handles a **whole query group** for a KV split: load each K/V slice element **once**, dot it against all `G` query rows, and maintain `G` running online-softmax states + `G` output accumulators. Each query head still emits its own `(m_i, l_i, Õ_i)` partial, so `llm_attention_combine` is unchanged (still indexed per query head). KV-read traffic drops ~G× — potentially a large further long-ctx decode win, biggest for fp32/bf16 (more bytes, worse L2 fit).

Stay **scalar** (GQA-4, per #235's analysis — tensor cores need G≥8 to fill an mma tile). The win is the shared *load*, not tensor cores.

## Design tension (measurement-gated)

Sharing reduces the block count by G× (E4B: `numKvHeads × nSplits = 2 × nSplits` vs `8 × nSplits`). With only 2 KV heads that can **re-starve occupancy** — the exact problem #235 fixed. So the split-count heuristic must be re-tuned (smaller chunk / more splits, or sub-grouping a KV head's work across blocks) to keep ~enough blocks to fill the SMs. Net should be positive (fewer bytes per block × more blocks), but it is **not guaranteed to beat Phase A** — gate the merge on the decode-vs-ctx A/B showing a win over the current split kernel, per dtype. If it regresses occupancy more than it saves traffic, document and stop.

## Scope

- [ ] Group-shared `llm_attention_splitkv` variant: grid keyed on KV head (+ sub-split), G query rows per block, G partials out. Keep the combine kernel and partials layout.
[...]

PR #239, opened 2026-06-13
perf(cuda): GQA head-sharing for split-KV decode (auto fp32 long-ctx) (#237)

pekkah:
## What

GQA head-sharing for the split-KV decode-attention path (#237, follow-up to #235). One block handles a KV head's whole query group, loading each K/V slice **once** instead of G separate blocks each re-reading it.

## Why

The per-head split kernel runs one block per `(query-head, KV-split)`. With GQA, the `G = numHeads/numKvHeads` query heads sharing a KV head are G separate blocks each re-reading the *same* KV slice → at long ctx (KV exceeds L2) that's **G× redundant HBM reads**, which inflate the now-bandwidth-bound long-ctx decode.

## How

- `llm_attention_splitkv_grouped<KV>` (fp32/bf16/q8 thunks) + `sharpi_kv_dot_multi`: grid `(num_kv_heads, n_splits)`; each block reads each K/V element **once** and fuses it into all G query rows' online-softmax states (dynamic shared `G × CHUNK`). Emits the **same per-query-head partials** → `llm_attention_combine` and the partials layout are unchanged.
- Argmax-stable, not bit-identical (re-derived reduction order; same contract as the per-head split).

## Measurement (E4B, head_dim 512, 2 KV heads, RTX 4070 Ti) — decode A/B vs the per-head split

| ctx | fp32 grouped/per-head | q8 grouped/per-head |
|---|---|---|
| 8192 | 49.2 / 53.2 (lose) | 51.0 / 55.5 (lose) |
| 24576 | 41.1 / 37.2 (**+11%**) | — |
| 32768 | 36.8 / 29.9 (**+23%**) | 40.8 / 45.2 (lose) |

The G× traffic cut only beats the G× fewer blocks when the read is **genuinely bandwidth-bound**. The 32K row is decisive: at identical 128 blocks, **fp32 wins but q8 loses** → it's bandwidth, not occupancy. q8/bf16 read too few bytes to be bandwidth-bound, and below ~24K fp32 loses to occupancy.

## Decision: auto-enable only where it wins

Grouped is **auto-selected only for fp32 at seqLen ≥ 24576** with an eligible head config (`numKvHeads ≥ 2`, `G ∈ [2,8]`, divides evenly). Everything else (q8/bf16, short ctx, MQA) keeps the per-head split. `SHARPI_SPLIT_DECODE_GROUPED`: `0` forces per-head, `1` forces grouped (bisect), unset = auto. The crossover is tuned to E4B-class shapes (head_dim 512); other head_dims shift it, hence the override.

## Validation

- **4 grouped-vs-single-block parity tests** (forced grouped): fp32/bf16/q8 on E4B + fp32 on Qwen3-8B's 8-KV-head geometry — argmax-stable, tight logit budgets (1.0/2.0/2.5). The kernel is correct for all dtypes even though auto only uses fp32.
- **6 per-head split + graph-boundary tests** stay green — the dispatch change is inert below the fp32/24K gate.

## Scope

Dense `CudaForwardPass` global layers, E4B-class GQA. Hybrid paths remain #238. The chunk stays 512 (no combine-cap change); a smaller chunk wouldn't rescue q8 (not bandwidth-bound), so it wasn't pursued.

### DIFF
--- a/src/SharpInference.Cuda/CudaBackend.cs
+++ b/src/SharpInference.Cuda/CudaBackend.cs
@@ -331,6 +331,11 @@ public sealed unsafe class CudaBackend : IComputeBackend, IImageOpsBackend, IDis
     private nint   _attentionSplitKvBf16Kernel;
     private nint   _attentionSplitKvQ8Kernel;
     private nint   _attentionCombineKernel;
+    // Issue #237 (GQA head-sharing): split-KV variant that processes a KV head's whole
+    // query group per block (grid num_kv_heads × n_splits), loading each K/V slice once.
+    private nint   _attentionSplitKvGroupedKernel;
+    private nint   _attentionSplitKvGroupedBf16Kernel;
+    private nint   _attentionSplitKvGroupedQ8Kernel;
 
     // Grow-only global score scratch for the wave-based >4096 batched-query SDPA
     // (issue #118). Sized W × num_heads × score_stride floats; W is chosen so this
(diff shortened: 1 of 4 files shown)


pekkah:
Addressed the 3-lens + gemini review (no correctness blockers found):

**gemini MEDIUM** (div-by-zero): `group = numHeads / numKvHeads` ran before the validation guard → guarded the division behind `numKvHeads ≥ 2` (`9586e67`). The guard's `< 2` already short-circuits the subsequent `%`.

**pr-test-analyzer** (the one pre-merge gap — the shipping default had no regression fence, and its mis-fire mode is the host `throw`): extracted the dispatch decision to a pure `CudaForwardPass.ShouldUseGroupedSplit(...)` and added a **no-GPU unit test** fencing the full table — q8/bf16 never auto, fp32 <24K off / ≥24K on, MQA/MHA/non-divisible/G=9 rejected, **G=8 eligibility boundary ok**, "0"/"1" overrides (`3211f1f`).

**code-reviewer** (minor): tightened the `AttentionSplitKv` host guard to the exact contract (`numKvHeads ≥ 2`, `G ∈ [2,8]`, divides).

**Deferred (both reviewers agreed):**
- G=8 kernel array bound (`dots[8]/acc[8]`) is not exercised end-to-end — no on-disk model has G=8 with numKvHeads≥2. The eligibility boundary is unit-tested; noted in the test comment.
- Grouped graph-replay *across* a split boundary — the grouped parity tests already exercise capture+replay and the early-exit empty-meta write (splits 10–11 are out-of-range at capture), just not the empty→active transition mid-replay; it's structurally identical to the validated per-head boundary test and provably correct by the meta-masking argument.

Validation: gate unit test + grouped fp32 parity (through the refactored dispatch) green; 4 grouped parity (all dtypes) + 6 per-head/boundary green; build clean.

PR #240, opened 2026-06-13
perf(cuda): split-KV for the layer-split MoE hybrid pass (#238)

pekkah:
## What

Extend flash-decoding split-KV (#235/#237) to the layer-split MoE hybrid pass (`CudaHybridForwardPass`: Qwen3-Coder-30B, Llama-4 Scout, …). It was dense-only; the hybrid's GPU-resident attention layers still hit the long-ctx occupancy collapse #235 fixed.

## How

Wire the existing `CudaBackend.AttentionSplitKv` (+ partials buffer, dtype dispatch, the #237 grouped auto-select) into `CudaHybridForwardPass.AttentionKv`, mirroring the dense gate. **Reuses the validated #235/#237 kernels unchanged** — only the dispatch/buffer wiring is new.

## Measurement (the gate) — Qwen3-Coder-30B-A3B Q4_K_M, q8 KV, 4070 Ti, split on/off

| ctx | off | on | on/off |
|---|---|---|---|
| 4096 | 16.5 | 17.1 | 1.03 (break-even) |
| 6144 | 13.9 | **20.7** | **1.49** |
| 8192 | 12.0 | **19.5** | **1.63** |
| 16384 | 7.7 | **16.1** | **2.08** |

Attention **is** a meaningful share of the (CPU-MoE-bound) long-ctx hybrid decode — a clear win for the big layer-split model.

## The threshold: seqLen > 4096 (not the dense 2048)

A small fast all-GPU MoE — **OLMoE** (16 heads, context-capped at 4096) — **regresses 2×** (59→32 t/s) when split at 4096: attention is a tiny fraction of its fast decode, so the split's fixed overhead dominates. Raising the hybrid threshold to `seqLen > 4096` (`HybridSplitMinSeq`) excludes OLMoE's entire operating range (no regression) while keeping every confirmed Coder win (break-even at 4096, growing to 2.08× at 16K). `SHARPI_SPLIT_DECODE=0` kill-switch; skipped under TurboQuant.

## Validation

- **`Coder30B_Q8SplitKv_MatchesSingleBlock`**: split-vs-single-block at >4096 ctx, argmax-stable + tight logit budget (the hybrid wiring is correct). The A/B also confirmed q8 first-token matched single-block at all 4 ctx points.
- **OLMoE short-ctx** hybrid tests unaffected (ctx 2048 < gate → split off).
- Kernel correctness carries from the dense #235/#237 bit-faithful parity tests (same `AttentionSplitKv`).

## Scope / follow-up

`CudaHybridForwardPass` (layer-split MoE). The **GDN hybrid pass** (`CudaHybridGdnForwardPass`, Qwen3.6) has a different attention dispatch and GDN-dominated decode — deferred to its own measurement-gated follow-up (it has only ~10 attention layers, so the payoff is uncertain).

### DIFF
--- a/src/SharpInference.Engine/CudaForwardPass.cs
+++ b/src/SharpInference.Engine/CudaForwardPass.cs
@@ -508,11 +508,12 @@ public sealed unsafe class CudaForwardPass : IForwardPass, IBatchedForwardPass
     private Tensor? _splitKvPartialMeta;
     // Only split once the context is long enough that the 8-block launch underutilizes the
     // GPU; below this, the single-block kernel is already near the weight-matvec floor.
-    // Internal so CudaHybridForwardPass reuses the same gate (#238).
-    internal const int SplitKvMinSeq = 2048;
+    // (CudaHybridForwardPass uses its own higher HybridSplitMinSeq — #238.)
+    private const int SplitKvMinSeq = 2048;
     // Upper bound on maxSeqLen for the split path: nSplits = ceil(ctx/chunk) must fit the
     // combine kernel's SPLITKV_MAX_SPLITS shared array. Derived from the backend constants so
-    // the gate can't drift from the kernel bound (= 256 × 512 = 131072).
+    // the gate can't drift from the kernel bound (= 256 × 512 = 131072). Internal so the hybrid
+    // pass reuses the same bound (#238).
     internal const int SplitKvMaxCtx = CudaBackend.SplitKvMaxSplits * CudaBackend.SplitKvChunk;
     // SHARPI_SPLIT_DECODE=0 reverts to the single-block-per-head decode attention (the A/B
     // baseline; lets parity tests compare the split path against it). Default on.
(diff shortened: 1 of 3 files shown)


pekkah:
Addressed the 3-lens review (all three found no correctness blockers — the wiring is a faithful mirror of the dense path, and every hybrid edge — per-layer head dims / kv heads, eager-vs-graph decode, TQ exclusion, disposal, fall-through — was verified safe):

- **code-reviewer** (minor): `SplitKvMinSeq` was widened to `internal` but the hybrid uses its own `HybridSplitMinSeq` — the widening was unused. Reverted to `private`; kept `SplitKvMaxCtx` internal (genuinely reused).
- **pr-test-analyzer**: tightened the parity max-abs from 6.0 → **4.0** — this is same-dtype (q8-vs-q8) split-vs-single, where only the combine's reduction reorder diverges (the 6.0 was inherited from the cross-dtype q8-vs-fp32 test), so the bound is now load-bearing; re-validated green. Corrected the test docstring's "graph interaction" overclaim — the MoE hybrid runs decode **eager** (no graph capture), so the test covers the eager wiring; split-under-graph-replay is validated on the dense path (`Gemma4_E4B_SplitKv_GraphReplayCrossesBoundary`).

**Deferred (acceptable per reviewers):** a Gemma4-partial-offload hybrid graph-captured split test (narrow: graphs default off in the hybrid; the dense graph test + the hybrid eager test cover the pieces) and a bf16-on-hybrid case (the kernels are dtype-validated on dense; the hybrid passes `_kvDType` straight through). Both noted as #238 follow-ups.

silent-failure-hunter confirmed no reachable overrun/null/double-free/graph-flip.

PR #241, opened 2026-06-14
perf(cuda): split-KV for the GDN hybrid pass — completes #238

pekkah:
## What

Completes #238 — extends flash-decoding split-KV to `CudaHybridGdnForwardPass` (Qwen3.6-35B/27B), the last forward pass that still ran single-block decode attention on its GPU attention layers. (The dense path was #235/#237; the layer-split MoE hybrid was #240.)

## How

Wire `CudaBackend.AttentionSplitKv` into `GpuAttnBlockAt` — the per-token attention site used by both `Forward` (single-user) and `BatchForward2` (MTP k-token verify) — mirroring the dense/MoE-hybrid gate (incl. the #237 grouped auto-select). Reuses the validated #235/#237 kernels; only the dispatch/partials-buffer/dispose wiring is new. GDN KV is fp32/bf16 (no q8).

## Measurement (Qwen3.6-35B-A3B, bf16 KV, 4070 Ti, split on/off)

| ctx | off | on | on/off |
|---|---|---|---|
| 8192 | 16.4 | 24.5 | 1.50 (cold-confounded) |
| 16384 | 20.6 | 24.6 | **1.20** |

Clean **+19% at 16K**. The 8K ratio is **cold-cache-confounded**: this is a CPU-MoE mmap model and the bench opens a fresh model per run, so the *first* run (8K-off) reads cold (`off@16K=20.6 > off@8K=16.4` — backwards for a real ctx effect — confirms it). By the 16K runs the OS page cache is warm for both off and on, so 1.20× is the clean number. Correctness: first-token matched single-block (561/561) at both ctx.

The win is smaller than the MoE hybrid's 2.08× — expected, since GDN decode has only ~10 attention layers vs the GDN scan + MoE.

## Threshold: seqLen > 8192 (conservative)

The attention share is small, so the split overhead amortizes only at genuinely long ctx; with so few attention layers a moderate-ctx split risks the OLMoE-style overhead regression. `GdnSplitMinSeq = 8192` targets the clean-win region (4096–8192 is unverified for this pass). `SHARPI_SPLIT_DECODE=0` kill-switch.

## Validation

- **`CudaHybridGdnForwardPass_SplitKv_MatchesSingleBlock`** — split-vs-single-block at >8192, argmax-stable + blow-up ceiling (top-5 is the hard gate; the GDN stack carries ~1e-3/layer Q8_1 noise over 40 layers, so a tight logit bound isn't meaningful here). The A/B also matched first-token 561/561 at 8192 and 16384.
- Existing GDN tests (ctx ≤ 4096 < gate → split off) unaffected.

With this, #238 is complete (both hybrid passes) and the flash-decoding arc spans dense (#235), GQA head-sharing (#237), MoE hybrid (#240), and GDN hybrid (this).

### DIFF
--- a/src/SharpInference.Engine/CudaHybridGdnForwardPass.cs
+++ b/src/SharpInference.Engine/CudaHybridGdnForwardPass.cs
@@ -3830,8 +3830,10 @@ public sealed unsafe class CudaHybridGdnForwardPass : IForwardPass
             _gpu.KvAppend(_gpuK, _gpuV, _gpuKCache[layer]!, _gpuVCache[layer]!, kvDim, kvPosition, _maxSeqLen);
 
         int seqLen = kvPosition + 1;
-        // Flash-decoding split-KV (#238) at long ctx; else the single-block kernel. Mirrors the
-        // dense/MoE-hybrid gate (incl. the #237 grouped auto-select).
+        // Flash-decoding split-KV (#238) at long ctx; else the single-block kernel. Uses the
+        // #237 grouped auto-select like the dense/MoE-hybrid passes. (No `maxSeqLen <= _maxSeqLen`
+        // guard as on the dense path: this site always passes _maxSeqLen as the cache extent, so
+        // nSplits == the buffer's nSplitsMax exactly — there is no per-call maxSeqLen to bound.)
         if (_splitKvPartialO is { } splitO && _splitKvPartialMeta is { } splitMeta
             && seqLen > GdnSplitMinSeq)
         {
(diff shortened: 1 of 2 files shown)


pekkah:
Addressed the 3-lens review (all three: no correctness blockers — the `GpuAttnBlockAt` KvAppend restructure is behavior-preserving, the partials buffer reuse across `BatchForward2`'s MTP tokens is serial-stream-safe, buffer sizing is exact, and the non-split fall-through is byte-identical):

- Noted the **MTP draft head** (`GpuMtpAttnBlock`) deliberately stays single-block — it's a single attention layer, so the split occupancy win is negligible (the main-model verify via `GpuAttnBlockAt` does get the split). (code-reviewer obs. A)
- Clarified the comment that the GDN dispatch omits the dense `maxSeqLen <= _maxSeqLen` guard **on purpose**: this site always passes `_maxSeqLen` as the cache extent, so `nSplits` equals the buffer's `nSplitsMax` exactly — there's no per-call `maxSeqLen` to bound. (silent-failure obs.)

**Deferred (per reviewers, noted as #238 follow-ups):** an MTP×split parity test on the 27B-MTP at >8192 ctx (reachable but a niche config; the shared partials buffer is serial-stream-safe by inspection — `split→combine→o-proj` completes per token before the next overwrites) and an fp32-on-GDN case (shared with the dense fp32 split coverage; the GDN passes `_kvDType` straight through). The max-abs 10.0 ceiling is intentionally a blow-up bound — top-5 stability is the hard gate, since the 40-layer GDN stack's Q8_1 noise makes a tight logit bound infeasible (matches the dense chunked-prefill tests).

PR #242, opened 2026-06-14
docs: long-context decode flash-decoding spans dense + hybrid passes

pekkah:
## What

Update the README "Long-context decode" note: it said "(dense CUDA)" but flash-decoding (split-KV) now covers all three CUDA forward passes after #240 (MoE hybrid) and #241 (GDN hybrid).

## Changes

- Note now reads "on every CUDA path — dense, MoE-hybrid, and GDN-hybrid".
- Added the new long-ctx model numbers: Qwen3-Coder-30B (MoE hybrid) q8 7.7→16.1 t/s @16K (2.08×); Qwen3.6-35B-A3B (GDN hybrid) bf16 +19% @16K; kept the dense Gemma 4 E4B figures.
- Noted the per-path engage thresholds (dense ~2K / MoE-hybrid 4K / GDN 8K).

## No table changes

The headline perf table is near-zero-ctx decode / ~1K-ctx prefill. Split-KV only engages at long ctx (>2K–8K), so no table cell changes — same as the original #235 doc update.

Docs-only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,9 +60,9 @@ same box). Llama-4 Scout and Qwen3-Coder Vulkan-hybrid keep prior values (not re
 
 **Long-context decode** uses flash-decoding (split-KV) on every CUDA path — dense, MoE-hybrid, and
 GDN-hybrid: the per-token KV read parallelizes across SMs instead of one block per head, so decode no longer
-collapses as context grows. At long ctx: Gemma 4 E4B (dense) fp32 13→30 / q8 11→45 t/s @32K; Qwen3-Coder-30B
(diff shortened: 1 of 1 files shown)


pekkah:
Addressed gemini's consistency nits in `e1568e3` (full name `Qwen3-Coder-30B-A3B`; `GDN-hybrid` in the threshold list). The comment-analyzer independently verified every claim (wiring in all 3 passes, thresholds 2048/4096/8192, the `SHARPI_SPLIT_DECODE=0` kill-switch, the perf numbers, and the no-table-change rationale) is accurate.

PR #244, opened 2026-06-14
feat(cli): llama.cpp-compatible --ngl/--device GPU selection flags

pekkah:
Reimplements stale PR #144 fresh on current master (that branch had conflicts in hot files; the core here is a clean new `GpuDevice` helper).

## What

Adds llama.cpp flag aliases and **single-GPU device selection** to both the `run` and `image` commands.

- **`--ngl` / `--gpu-layers`** — aliases for the existing `-g` / `--n-gpu-layers`. **`-g` is kept** (non-breaking; #144 had dropped it).
- **`--model-draft`** (alias of `--draft-model`) and **`--repeat-penalty`** (alias of `--rep-penalty`) — llama.cpp parity.
- **`--device <auto|none|cpu|INDEX|NAMEINDEX>`** — e.g. `0`, `CUDA0`, `Vulkan1`. Parsing lives in the new `GpuDevice.Resolve`:
  - pins **CUDA** via `CUDA_VISIBLE_DEVICES` (process-wide; only set when the user hasn't already constrained it),
  - returns the index for **Vulkan**'s explicit physical-device selector,
  - `none` / `cpu` forces the CPU path (overrides `--ngl`),
  - **single-device only** — comma-separated multi-GPU lists are rejected (SharpInference has no tensor/row split).

`VulkanBackend` gains `VulkanBackend(int deviceIndex = -1)` + `SelectPhysicalDevice(int)` with a bounds-checked, compute-queue-checked explicit-index path. The default `-1` preserves the prior discrete-GPU-preferred auto-select, so **all existing callers (server, tests, FLUX) are byte-for-byte unchanged**.

## Verification

- Full solution builds clean (TreatWarningsAsErrors, 0 warnings).
- CLI smoke tests of every `--device` parse branch:
  - `--device foo` → "expected a device index…"
  - `--device 0,1` → "multi-device split is not supported"
  - `--device=-1` → rejected (won't mis-parse as device 1)
  - `--device CUDA3` / `--device none` → valid, proceeds to model load
  - `--ngl`, `--model-draft`, `--repeat-penalty` recognized and shown in `--help`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,9 @@ SHARPI_MODEL=models/SmolLM2-1.7B-Instruct-Q4_K_M.gguf \
   dotnet run --project src/SharpInference.Server.Host -c Release
 ```
 
+GPU flags mirror llama.cpp: `-g`/`--ngl`/`--n-gpu-layers` are interchangeable, and `--device <0|CUDA0|none>`
+pins a single GPU (no multi-GPU split).
+
 ## Image generation
 
 Two pipelines, auto-detected from model filename. Benchmarked on AMD Zen 4 + RTX 4070 Ti (CUDA, 4 steps,
--- a/SharpInference.slnx
+++ b/SharpInference.slnx
@@ -23,6 +23,7 @@
     <Project Path="src/SharpInference.Vulkan/SharpInference.Vulkan.csproj" />
   </Folder>
   <Folder Name="/tests/">
+    <Project Path="tests/SharpInference.Tests.Cli/SharpInference.Tests.Cli.csproj" />
     <Project Path="tests/SharpInference.Tests.Core/SharpInference.Tests.Core.csproj" />
     <Project Path="tests/SharpInference.Tests.ForwardPass/SharpInference.Tests.ForwardPass.csproj" />
     <Project Path="tests/SharpInference.Tests.Pipeline/SharpInference.Tests.Pipeline.csproj" />
--- a/src/SharpInference.Cli/GpuDevice.cs
+++ b/src/SharpInference.Cli/GpuDevice.cs
@@ -1,0 +1,67 @@
+namespace SharpInference.Cli;
+
+/// <summary>
+/// Parses the llama.cpp-style <c>-dev/--device</c> option into a single GPU selection and
+/// applies it. Unlike llama.cpp's comma-separated multi-GPU list, SharpInference targets a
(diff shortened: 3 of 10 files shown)


pekkah:
Addressed the review feedback across three commits:

**Silent-failure-hunter (CRITICAL #1/#2):** An explicit `--device N` that was invalid/absent was being swallowed into a silent CPU fallback in `ImageCommand` (Z-Image printed the actively-wrong "no GPU detected"; FLUX printed nothing). Fixed by gating every backend-creation fallback `catch` on `deviceIndex < 0` — auto-select still falls back, but an explicitly-named device now propagates its error to the outer handler and is reported. Verified: `image --device 9 --backend vulkan` → `Error: --device 9: only 2 Vulkan device(s) present (valid indices 0..1).`

**Silent-failure-hunter (#5):** `--device none` contradicting an explicit GPU request now prints a `Note:` (run: vs `--ngl/-g`; image: vs `--backend`) instead of silently dropping to CPU.

**pr-test-analyzer (#9):** Added `SharpInference.Tests.Cli` (25 cases) covering every `GpuDevice.Resolve` parse branch — including the `-1`-must-not-mis-parse-as-device-1 guard and int overflow — plus the `CUDA_VISIBLE_DEVICES` pin/preserve side effect. Exposed via `InternalsVisibleTo`; collection parallelism is off so the env-mutating cases are deterministic.

**gemini (ImageCommand.cs:249):** Z-Image now logs the upscaler CPU fallback (`Upscaler backend: CPU (--device none)` / `CPU`) to match FLUX.

**gemini (GpuDevice.cs:57):** Declined — the suggested refactor is byte-for-byte identical to the current loop (same digit-strip + letters-only prefix check). The existing logic is now locked down by the 25 unit tests above (the `-1`/`0x1F`/`CUDA`/overflow rejection cases specifically).

**silent-failure-hunter (#3, CUDA index validation):** Not adding a CUDA-side device-count check. `GpuDevice.Resolve` runs before backend selection and can't know whether the target is CUDA or Vulkan, so a CUDA-count check there would wrongly reject a valid Vulkan-only index. After the #1/#2 fix every invalid explicit device *does* surface an error (Vulkan validates in-ctor; an explicit CUDA index that doesn't exist fails loudly at `cublasCreate`). The residual is message wording on the CUDA path, not a silent failure — left as a known limitation.

PR #249, opened 2026-06-14
FlashQLA-style chunked Gated DeltaNet prefill (CPU + CUDA)

pekkah:
## Summary

Ports the technique behind Qwen's [FlashQLA](https://github.com/QwenLM/FlashQLA) — the chunk-parallel `chunk_gated_delta_rule` — into SharpInference's existing Gated DeltaNet (GDN) path. FlashQLA the library is Hopper-only (SM90) TileLang/PyTorch, so it can't run on this stack or on consumer Ada cards; this PR implements the **algorithm** across the CPU and CUDA backends instead. The GDN prefill was previously a pure per-token sequential scan — the chunked form (UT-transform / forward substitution) is the matmul-shaped optimization that was missing.

## What's here

| Layer | Change | Validation |
|---|---|---|
| **CPU kernel** | `GdnKernels.GdnRecurrenceChunkedPrefill` — chunked delta rule, float32 + AVX2/FMA SIMD, log-space decay carry | ✅ parity tests vs sequential scan; **~5.3× faster** (`--gdn` bench) |
| **GDN block** | composition validated end-to-end (conv1d-prefill → SiLU → split → L2norm → tile → chunked recurrence) | ✅ `GdnBlockBatched_MatchesPerTokenBlock` |
| **CPU prefill wiring** | `HybridGdnForwardPass.PrefillChunked` — layer-major prefill, gated behind `SHARPI_GDN_CHUNKED_PREFILL` (default **off**) | ✅ composition-validated; model-gated end-to-end parity test ready |
| **CUDA kernel** | `llm_gdn_chunked_prefill` NVRTC kernel + `CudaBackend.GdnChunkedPrefill` wrapper (sm_89-compatible, 34 KB shared) | ⏳ compiles; **needs GPU run** — `GdnChunkedPrefill_MatchesCpuSequentialReference` is the on-hardware gate |

## Design notes

- **Layer-major CPU prefill is provably equivalent** to the token-major `Forward` loop: each token's per-layer work is independent given the prior layer's output, and the KV/GDN caches are written in position order. Attention and MoE/FFN run per-token through the same kernels in the same order (byte-identical); only the GDN recurrence changes algorithm (chunked), differing from the scan only by FP reduction order.
- **CUDA kernel**: one block per v-head, thread `j` owns value-column `j`, so forward substitution / S0 projections / state carry are fully column-local (no cross-thread sync); only the per-chunk `K·K`/`K·Q` dot matrices and the per-token RMSNorm reduction cooperate via shared memory.
- Everything new is **gated off by default** and does not touch the validated per-token paths, so it cannot regress existing behavior.

## Validation gap (honest)

Developed in a sandbox with **no GPU and no model on disk**, so the two end-to-end checks can't run in CI here:
- `HybridGdnChunkedPrefill_MatchesSequentialPrefill` — needs the qwen35moe GGUF; asserts chunked CPU prefill is argmax-identical + numerically close to the sequential loop.
- `GdnChunkedPrefill_MatchesCpuSequentialReference` — needs a CUDA GPU; compiles the NVRTC kernel and checks it vs the CPU oracle.

Both pass-by-construction from the validated unit-level pieces; they're the gates to flip the flags on. Full solution builds clean (TreatWarningsAsErrors); all sandbox-runnable tests pass.

## Not in this PR (next step)

Wiring `GdnChunkedPrefill` into `CudaHybridGdnForwardPass`'s batched-prefill path (replacing the `GdnRecurrenceScan` call) — recommended after the CUDA kernel parity test is confirmed green on hardware.

https://claude.ai/code/session_01YDvDsqib3d7seUoA6uR5qf

---
_Generated by [Claude Code](https://claude.ai/code/session_01YDvDsqib3d7seUoA6uR5qf)_


pekkah:
Thanks — addressed in eb9affa:

- **qkAt int overflow (medium):** fixed. Widened the offset math to 64-bit (cast the token index to `long` before the multiply) in `qkAt` and the 7 sibling `((c0+t/s)*hv+h)*d` sites in `GdnRecurrenceChunkedPrefill`. In-range behaviour is identical (33 GDN kernel parity tests green); this only fixes the >=512K single-call prefill case. The CUDA sibling already casts early.
- **Env-var test parallelism (medium):** no change needed — this is a false positive for this assembly. `tests/SharpInference.Tests.ForwardPass/xunit.runner.json` sets `parallelizeAssembly: false` and `parallelizeTestCollections: false`, so test classes never run concurrently. The new test follows the same env-var pattern as the existing `CudaHybridGdnBatchedPrefillTests` (which sets the same `SHARPI_CPU_MOE`) and restores state in `finally`.

Issue #247, opened 2026-06-14
Prefill optimization attack plan: Gemma 4 on RTX 4070 Ti (12 GB, Ada / sm_89)


pekkah:
> ## Status (2026-06-14) — verified against current code
> Tracing the actual prefill **dispatch** (not just the kernel wrappers) shows the headline Tier-1 item #2 is already done:
>
> - **#2 windowed FlashAttention-Tc2 for SWA — ALREADY IMPLEMENTED.** The Gemma 4 batched-prefill attention dispatch routes **both SWA and global layers to `FlashAttentionPrefillTc2`** (`CudaForwardPass.cs:2837-2863`), passing `isSwa ? window : 0` so the TC kernel masks to the sliding window itself. Scalar `AttentionSwaBatched` (`:2869`) is only the flash-*disabled* fallback. Windowed Tc2 is parity-tested (the #151 SWA configs). SWA prefill is already on tensor cores — not "1/6".
>
> **Still genuinely open (re-verified):**
> - **#1 PLE GPU gather+dequant** — still a CPU `Parallel.For` dequant + host→device upload (`CudaForwardPass.cs:2682-2700`; the "~30% of batched prefill" in-code note stands). Top lever, and Gemma-4-unique.
> - **#5 prefill CUDA graphs** — capture is decode-only (`CudaForwardPass.cs:1549`); prefill chunks aren't captured.
> - **#6 VRAM-resident weight cache** — `MatMulBatchedGemm` still re-dequants per batch.
>
> Net: lead with #1 (PLE), then #5/#6; #2 is done.
>
> ---

## Goal

Improve prefill (TTFT) throughput for Gemma 4 models on a 12 GB Ada GPU (RTX 4070 Ti, sm_89). This issue captures a ranked attack plan based on profiling the current prefill path and surveying the external state of the art.

## Context: what's already solved (do not re-chase)

The prefill path is already mature. The following are implemented and should **not** be treated as opportunities:

- Batched all-GPU trunk prefill — `CudaForwardPass.PrefillBatchedTrunk` (`CudaForwardPass.cs:2585`)
- Chunked prefill, 4096-token windows (`CudaForwardPass.cs:200`)
- FlashAttention tensor-core Tc2 (multi-warp d-split, +27–40%) — `CudaBackend.cs:4234`
- int8 tensor-core MMQ + SoA repack (Q8_0/Q4_K/Q4_0) — `CudaBackend.cs:2308`
[...]

Issue #141, opened 2026-06-05
Gemma 4 GPU prefill is ~70x slower than llama.cpp — adopt cuBLAS GEMM for prefill


pekkah:
> ## ✅ RESOLVED (2026-06-14)
> Prefill went from ~108 → ~3444 t/s (E4B Q8, RTX 4070 Ti): dequant→cuBLAS/MMQ GEMM (#143, 108→1564), tensor-core flash-attention prefill (#146/#147, PR #148), and the SoA Q8_0 repack (#149). That is ~2× of llama.cpp's ~8475 t/s — within the **"2–3×, not 70×"** acceptance bar. The remaining ~2× compute lever (a different MMQ tiling regime) is tracked in **#152** (split-K & cp.async already ruled out there). Decode was left untouched per scope (companion #142). Closing.
>
> ---

## Finding

Gemma 4 GPU **prefill is ~70× slower than llama.cpp** — the single largest gap, and the one
most worth closing.

Benchmark (RTX 4070 Ti, `gemma-4-E4B-it-Q8_0`, all layers on GPU):

| | llama.cpp b9529 (cuda-12.4) | SharpInference | gap |
|---|---|---|---|
| Prefill (pp512) | **~8,000 t/s** | ~108 t/s | **~70×** |

(llama.cpp via `llama-bench -ngl 99 -p 512`; ours via `bench-textgen` warm prefill.)

## Root cause

Our batched-trunk prefill (#136) collapses per-position launches into **matvec / GEMM-N**
kernels — but those are **memory-bound**: they re-stream the full weight matrix roughly every
~2 tokens (prefill ≈ 2× decode, which is all #136 bought). llama.cpp runs prefill as a
**compute-bound cuBLAS GEMM** that reads each weight **once per ~512-token batch**, so on a
compute-rich GPU it's bandwidth-bound vs compute-bound — hence the ~70×.

## Proposal

Route the prefill trunk matmuls (Q/K/V/O projections, FFN gate/up/down) through a real GEMM
for the multi-token batch:

- We already link cuBLAS (`cublas64_12`, used for DiT) — use `cublasGemmEx` with the prompt
  activations as `[N × dim]`.
- Q8_0 weights need a real GEMM input: dequantize to fp16/bf16 once per matmul (or tile on the
  fly), or add a quantized-GEMM (MMQ-style) kernel. llama.cpp uses dequant→cuBLAS for large
  batches and its MMQ int8 kernels otherwise.
[...]

PR #256, opened 2026-06-15
perf(gemma4): GPU-side PLE pre-pass dequant for batched prefill (#247 item 1)

pekkah:
Implements **item #1** of the #247 Gemma 4 prefill attack plan — the PLE pre-pass dequant — and reports findings on the rest.

## What changed
The Gemma 4 batched-prefill PLE pre-pass (`BuildPerLayerProjectionsBatched`) dequanted each token's `per_layer_token_embd` row on the **CPU** (`Parallel.For` over N×10752 elements) and uploaded the full f32 rows. This moves the dequant to the **GPU**:

- Gather the chunk's **raw packed** rows on the CPU (a memcpy, no FP math).
- Upload the **4×-smaller** quant bytes (vs f32 rows).
- Dequant on-device into `_bpPleRowAll` via new contiguous-row kernels.

The PLE table stays **CPU-resident** (TierPlanner budget + auto-context untouched, no extra VRAM), so this is a **no-regression** change on the 12 GB card. New kernels `llm_dequant_rows_q8_0` / `llm_dequant_rows_q6k`; backend `DequantRowsQ8_0/Q6K` + `UploadRawInto`. Kill-switch `SHARPI_PLE_GPU_DEQUANT=0` (also the F32-table fallback).

## Correctness
- GPU dequant is **bit-identical** to CPU `Dequantize.ToFloat32` (both `(d·scale)·q`, exact `cvt.f32.f16`), so the strict `Gemma4_E4B_BatchedPrefill_GemmOff_MatchesSequentialBitExact` oracle still holds.
- New `CudaDequantRowsTests` proves bit-identity directly for Q8_0 + Q6_K incl. the real 10752-wide row (**6/6 pass**).
- Full `Gemma4CudaBatchedPrefillTests` suite passes **8/8** (incl. the bit-exact oracle, flash/GEMM/MMQ argmax oracles, past-window, chunked >4096).
- Solution builds clean (0 warnings under `TreatWarningsAsErrors`).

## Measured (E4B-Q8_0, `-g -1`, warm)
PLE phase **80 → 63 ms @ N=4096 (~21%)**, **8 → 4 ms @ N=972** — at zero VRAM cost, plus reduced host RAM and PCIe traffic. The `ple` phase is dominated by the unavoidable per-layer-proj GEMM, so the dequant was the removable slice; the original "~30% of batched prefill" (#141) doesn't reproduce on the current warm, layer-optimized path.

## Notes on the rest of #247
Investigated and reported in the issue:
- **#2 (windowed SWA flash) is already done** — both SWA and global layers route to `FlashAttentionPrefillTc2`.
- **#6 (VRAM weight cache) is a no-op on the local models** — they all use the MMQ path (cached SoA repack, no per-call weight work); it only helps Q6_K/Q5_K `_M`-quant trunks on multi-chunk/repeated prefills, none available to measure.
- **#5 (prefill CUDA graphs) buys <1.5% at N≥512** — prefill is compute-bound (matmul 567 ms + attn 310 ms of ~1063 ms @N=4096); graphs cut launch overhead (~4–12 ms), large + risky to implement (instrument all batched kernels).

A 12B note for the issue: the **12B has no PLE table** (`embedding_length_per_layer_input=0`) — PLE is E2B/E4B only.

### DIFF
--- a/src/SharpInference.Cuda/CudaBackend.cs
+++ b/src/SharpInference.Cuda/CudaBackend.cs
@@ -841,6 +841,7 @@ public sealed unsafe class CudaBackend : IComputeBackend, IImageOpsBackend, IDis
             throw new InvalidOperationException($"UploadRawInto: handle {dst.Handle} not registered.");
         if ((nuint)src.Length > entry.byteSize)
             throw new ArgumentException($"UploadRawInto: source ({src.Length} bytes) exceeds destination capacity ({entry.byteSize} bytes).");
+        if (src.IsEmpty) return;   // nothing to copy; avoids fixed → null-ptr on an empty span
         fixed (byte* s = src)
             UploadViaStaging(entry.devPtr, s, (nuint)src.Length);
     }
--- a/src/SharpInference.Cuda/CudaBackend.cs
+++ b/src/SharpInference.Cuda/CudaBackend.cs
@@ -129,6 +129,8 @@ public sealed unsafe class CudaBackend : IComputeBackend, IImageOpsBackend, IDis
     private nint   _embedLookupQ6KKernel;   // #124: Q6_K tied embedding (Gemma 4 12B)
     private nint   _embedLookupQ80Kernel;
     private nint   _embedLookupQ80BatchedKernel;
+    private nint   _dequantRowsQ80Kernel;   // #247: GPU-side PLE pre-pass (Q8_0 rows → f32)
+    private nint   _dequantRowsQ6KKernel;   // #247: GPU-side PLE pre-pass (Q6_K rows → f32)
     private nint   _matvecF32Kernel;
     private nint   _matvecQ4KKernel;
     private nint   _matvecQ4KSoaKernel;     // #156: scale-pre-unpacked SoA decode matvec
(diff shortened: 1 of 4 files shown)


pekkah:
Thanks for the review. Disposition of the findings:

**Applied** (commit 8af6870):
- `UploadRawInto` — early-return on an empty span (kept after the handle/capacity validation) to avoid `fixed` → null-ptr.
- Test `HalfToUshort` — switched to `BitConverter.HalfToUInt16Bits` (no per-call `byte[]`).

Also covered the full int8 range incl. `-128` in the test and documented the int-safe `N*bpr` bound at the 4096 chunk cap (commit edd5cae).

**Not applied (with rationale):**
- *Hoisting the `Span`/pinned-pointer out of the `Parallel.For`* — these don't compile: `Span<byte>` is a `ref struct` and can't be captured in a lambda (CS8175), and a `fixed`-pinned pointer can't either (CS1686). The current `new Span<byte>(array).Slice(...)` inside the loop is allocation-free (it's a stack struct over the existing array ref), so there's no heap/GC overhead to remove.
- *`N == 0` guard in `BuildPerLayerProjectionsBatched`* — unreachable: a prompt is never empty, and the surrounding `PrefillBatchedTrunk` (embed all N, final norm on token N-1) doesn't support N=0 either, so a local guard would be false comfort rather than making N=0 work.

Issue #203, opened 2026-06-10
perf(cuda): batched-decode MMQ follow-ups — Q6_K MMQ kernel, Q/O split-K, cp.async staging (#201 leftover headroom: ~317 -> ~450+ t/s at N=8)


pekkah:
**Umbrella tracker** for the batched-decode headroom left after #201 / PR #202. Work items are split into child issues:

- [ ] #204 — Q6_K decode MMQ kernel (per-16 int8 scales vs the m16n8k32 fragment; ~5 ms/step)
- [ ] #205 — n16 MMQ kernel efficiency: split-K / BM=32 for the [4096×*] shapes + cp.async staging (+ small K/V & attn_v shapes; ~3-4 ms/step)
- [ ] #206 — evaluate `SHARPI_BATCH_DECODE_MMQ` default-on for batched serving (N=5-7 cells + quality soak)

## Context

#201 / PR #202 landed the opt-in decode MMQ (`SHARPI_BATCH_DECODE_MMQ=1`): N=8 aggregate 246 -> 317 t/s. The N=8 step is now ~25.2 ms against a ~9.7 ms whole-step weight-streaming floor. Per-shape (nsys, 40-step capture, Qwen3-8B Q4_K_M @ 4070 Ti, N=8):

| shape | ms/step | vs floor | path |
|---|--:|--:|---|
| q6k ffn_down [4096x12288] | 4.77 | ~3.2x | WS-sw (bit-exact) -> #204 |
| q6k lm-head [151936x4096] | 2.76 | ~2.7x | WS-sw -> #204 |
| q4k gate/up [12288x4096] | 6.89 | 1.7x | MMQ n16 -> #205 (cp.async) |
| q4k Q/O + q4k-half ffn_down [4096x*] | 4.47 | ~2.7x | MMQ n16, grid 64 blocks -> #205 (split-K) |
| q4k K/V [1024x4096] | 0.91 | ~6.6x | WS (rows<2048 fallback) -> #205 (low prio) |
| q6k attn_v [1024x4096] | 0.77 | ~16x | WS-sw -> #205 (low prio) |

Combined estimate if #204 + #205 land near floor: **~450-500 t/s aggregate at N=8**.

Measured dead ends recorded in #201 / PR #202 (do not retry without new evidence): token-warp WS geometry; SoA-Q8_1-act WS reads (register-bound); sub-block-sized MMQ K-step (latency-serial).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #262, opened 2026-06-16
perf(cuda): default-on int8 decode MMQ for multi-user batched serving (#206)

### LINKED ISSUE #206: perf(cuda): evaluate SHARPI_BATCH_DECODE_MMQ default-on for batched serving (#203 item 5)
Child of #203. The #201 decode MMQ ships **opt-in** (`SHARPI_BATCH_DECODE_MMQ=1`): N=8 aggregate 248 -> 317 t/s, argmax-stable (same contract as the prefill MMQ and the `SHARPI_BATCH_DECODE_GEMM` toggle), with per-tensor fallback to the bit-exact WS path for Q6_K / rows<2048 / N<5.

Before flipping the default for batched serving (`SHARPI_MAX_BATCH` / ContinuousBatchingEngine users):

1. **N=5-7 cells**: the dispatch threshold N>=5 was set from N=2 (123 vs 139, loses), N=4 (wash), N=8 (+28%); the 5-7 range is interpolated, not measured. Bench those cells (`SHARPI_BENCH_BATCH_N=5,6,7`) and adjust the threshold if 5/6 are washes.
2. **Quality soak**: argmax-stable is a kernel-level contract; run a multi-sequence generation comparison (greedy, ~100+ tokens per seq, several prompts) MMQ=1 vs default and confirm token-level agreement or benign divergence only.
3. **Decision**: default-on for BatchForwardMulti only (single-user Forward is untouched either way), keeping `SHARPI_BATCH_DECODE_MMQ=0` as the bit-exact kill-switch — or keep opt-in and document the knob in the server/CLI docs.
4. Re-evaluate after #204/#205 land — if Q6_K joins the MMQ path the default-on win grows from +28% to an estimated +60-80% at N=8.

pekkah:
## Summary

Resolves #206 (umbrella #203). The #201 int8 tensor-core batched-decode MMQ tile (`SHARPI_BATCH_DECODE_MMQ=1`) shipped opt-in. Re-benched and flip it **default-on for multi-user continuous-batching decode**, while keeping single-user speculative-decode `BatchVerify` bit-exact.

## Why (measured, Qwen3-8B Q4_K_M @ RTX 4070 Ti)

Aggregate decode t/s, MMQ vs the bit-exact weight-stationary path it replaces:

| N | WS (was default) | MMQ (now default) | Δ |
|---|---|---|---|
| 5 | 204.9 | 227.6 | **+11.1%** |
| 6 | 225.4 | 262.7 | **+16.8%** |
| 7 | 242.3 | 297.3 | **+22.8%** |
| 8 | 249.7 | 318.9 | **+27.6%** |

The N=5 cell (the dispatch threshold) is a solid win, not a wash — confirms the existing `N≥5` gate. All cells are far above the 3–4% bar.

## Scoping (the load-bearing detail)

`BatchVerify` (single-user spec decode) shares `BatchForwardMulti`, and `CudaSpecBatchVerifyTests.Qwen3_8B_SpecDecode_GreedyParity_E2E` asserts **exact** token equality over 48 tokens. The MMQ path is argmax-stable, **not** bit-exact — a naive global default-flip would put that guarantee at risk. So:

- `BatchForwardMulti` gains `bool allowDecodeMmq = true`. Continuous batching (via the `IBatchedForwardPass` interface impl) uses the default → MMQ on. `BatchVerify` passes `false` → stays on the bit-exact WS kernels.
- `SHARPI_BATCH_DECODE_MMQ` becomes tri-state: **unset** = on for multi-user decode / off for spec-verify; **`0`** = off everywhere (bit-exact kill-switch); **`1`** = on everywhere incl. spec-verify (argmax-stable, perf A/B only).

## Test gap closed

The pre-existing #201 "MMQ oracle" ran at **N=2**, below the `N≥5` dispatch threshold, so it silently fell back to WS and never exercised the int8 mma kernel. New `Qwen3_8B_BatchForwardMulti_DecodeMmq_N6_MatchesSingleUser` runs N=6 (3×PromptA + 3×PromptB, two decode steps) so the big ≥2048-row Q4_K matmuls actually take the MMQ path, asserting per-sequence argmax/top-5 parity with the single-user reference + KV-append/position indexing.

## Verification (4070 Ti, Qwen3-8B-Q4_K_M)

- Default-on bench reproduces the win automatically (318.9 t/s @ N=8, no env var).
- `SHARPI_BATCH_DECODE_MMQ=0` reverts to 249.0 t/s (bit-exact WS).
- New N=6 MMQ oracle: green.
- All 5 `CudaSpecBatchVerifyTests` green, incl. the 48-token greedy parity E2E (spec-verify stayed bit-exact).
- `CudaBatchForwardMultiTests` (7 tests) green.
- Release build clean (0 warnings, TreatWarningsAsErrors).

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,10 +147,12 @@ dotnet run --project src/SharpInference.Cli -c Release -- \
   -p "Write a binary search in Rust" --temp 0
 
 # API server (OpenAI /v1/chat/completions + Anthropic /v1/messages, port 5000)
-# Multi-user: SHARPI_MAX_BATCH=8 enables continuous batching (CPU backend). Long-prompt
-# admission prefills in SHARPI_PREFILL_CHUNK-token slices (default 256, 0 = blocking)
-# interleaved with decode, packed across prompts; SHARPI_KV_BUDGET_MB caps total KV
-# memory for admission backpressure (default: half of RAM). See issue #183.
+# Multi-user: SHARPI_MAX_BATCH=8 enables continuous batching (CPU or CUDA backend). On CUDA,
+# batched decode uses int8 tensor-core matmuls by default at N>=5 (Qwen3-8B Q4_K_M @ 4070 Ti:
+# +11% to +28% aggregate t/s from N=5 to N=8); SHARPI_BATCH_DECODE_MMQ=0 forces the bit-exact
+# path. Long-prompt admission prefills in SHARPI_PREFILL_CHUNK-token slices (default 256,
+# 0 = blocking) interleaved with decode, packed across prompts; SHARPI_KV_BUDGET_MB caps total
+# KV memory for admission backpressure (default: half of RAM). See issues #183 / #206.
 SHARPI_MODEL=models/SmolLM2-1.7B-Instruct-Q4_K_M.gguf \
   dotnet run --project src/SharpInference.Server.Host -c Release
 ```
--- a/src/SharpInference.Engine/CudaForwardPass.cs
+++ b/src/SharpInference.Engine/CudaForwardPass.cs
@@ -438,14 +438,27 @@ public sealed unsafe class CudaForwardPass : IForwardPass, IBatchedForwardPass
         Environment.GetEnvironmentVariable("SHARPI_BATCH_DECODE_GEMM") == "1";
     private readonly bool _batchDecodeWeightStationary =
         Environment.GetEnvironmentVariable("SHARPI_BATCH_DECODE_WS") != "0";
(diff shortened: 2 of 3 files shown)


pekkah:
Addressed the Gemini review (66e2275): `SHARPI_BATCH_DECODE_MMQ` is now parsed once into `readonly bool` fields at construction instead of computed properties, so `BatchDecodeMatMul` does no string compare on the per-matmul hot path. Functionally identical; N=6 MMQ oracle still green.

Issue #145, opened 2026-06-06
perf(cuda): apply PR #143 review micro-opts (16-bit scale loads, shift/mask staging, overflow-check ordering)


pekkah:
Automated review (Gemini code-assist) on PR #143 left 7 medium-priority perf/correctness micro-opts. They are non-blocking — PR #143 is tested, CI-green, and review-toolkit-clean — so deferring them here rather than destabilizing the merge.

**CUDA kernels (`src/SharpInference.Cuda/CudaTextKernels.cs`):**
- [ ] `llm_flash_attn_prefill_f32` K-staging loop (~L3137): replace `idx / hd2` + `idx - kk*hd2` with shift/mask. **Caveat:** bakes a power-of-two `head_dim` assumption into an otherwise-general kernel. Current callers (Gemma 4: 256/512) are all power-of-two, but guard or document before applying.
- [ ] Same kernel, V-staging loop (~L3147): shift/mask on `head_dim`. Same caveat.
- [ ] `llm_matvec_q8_0` scale load (~L1342): single 16-bit `ld.global.u16` instead of two `sharpi_byte_at` (b0 is always 2-byte aligned, stride 34).
- [ ] `MMQ_LOAD_TILE` macro scale load (~L1456): same single-16-bit-load.
- [ ] `llm_dequant_q8_0_to_f16` scale load (~L1594): same.

**C# (`src/SharpInference.Cuda/CudaBackend.cs`, `src/SharpInference.Engine/CudaForwardPass.cs`):**
- [ ] `MatMulBatchedMmq` (~CudaBackend L1827): compute `(long)cols*nTok` and overflow-check *before* the `(int)` cast (currently truncates first, then checks).
- [ ] `EmbedLookupQ8_0Batched` token upload (~CudaForwardPass L1826): use `CollectionsMarshal.AsSpan` for the `List<int>` path to avoid the `ToArray()` allocation.

All are perf/cleanliness; the only correctness-flavored one is the MMQ overflow-check ordering (cosmetic in practice — `cols*nTok` can't realistically exceed int range for current models). Each needs a managed rebuild + the focused `Gemma4Cuda` / `CudaMmqQ8_0Tests` / `CudaFlashAttnTests` pass.

Source: PR #143 review thread.

pekkah:
[...]

PR #264, opened 2026-06-16
perf(cuda): Q6_K decode MMQ via SoA reader-port (free AoS) — +17% N=8 (#204)

### LINKED ISSUE #204: perf(cuda): Q6_K decode MMQ kernel — per-16 int8 scales vs the m16n8k32 fragment (#203 item 1, ~5 ms/step at N=8)
Child of #203 (umbrella; per-shape data there). After #201/PR #202, the Q6_K tensors are the largest batched-decode chunk still on the bit-exact WS matvecs at N=8:

| shape | ms/step | vs floor |
|---|--:|--:|
| ffn_down [4096x12288] (Q6_K half) | 4.77 | ~3.2x |
| lm-head [151936x4096] | 2.76 | ~2.7x |
| attn_v [1024x4096] | 0.77 | ~16x |

The Q4_K n16 decode MMQ (`llm_mmq_q4k_soa_acts_n16`) showed the shape of the win (gate/up 11.6 -> 6.9 ms/step); Q6_K needs its own kernel because its **int8 scales are per 16 elements** while the m16n8k32 mma reduces k=32 into one s32 accumulator — the two scale groups inside a fragment cannot be folded after the dot.

Candidate designs (pick after a probe):
1. **k-permuted fragments**: permute the k-dimension (same permutation on A and B) so each mma`s k=32 covers... still mixes two scale groups in the accumulator — only works combined with (2).
2. **Two mmas per 32 elements with zero-masked halves** — 2x mma count, simple, scales fold per-16 cleanly; mma is cheap at decode tile sizes (the n16 kernel is staging-bound, not mma-bound).
3. **Lift llama.cpp`s mmq q6_K tiling** (their `vec_dot_q6_K_q8_1_mma` path) wholesale.
4. Skip the lm-head: it runs once per step and is already 2.7x floor; ffn_down/attn_v are the per-layer wins.

Contract: argmax-stable under the existing `SHARPI_BATCH_DECODE_MMQ=1` gate, same per-tensor dispatch fallback as Q4_K. A Q6_K SoA weight repack (like #156`s Q4_K one) may be a prerequisite for coalesced staging — the 210-B AoS block is word-hostile.

Validation pattern: `CudaDecodeMmqTests` (fp32 CPU reference per-RMS tolerance) + the `BatchForwardMulti` MMQ argmax oracle.

pekkah:
## Summary

Resolves #204 (umbrella #203). Routes the Q6_K trunk weights (lm-head + ffn_down-half) through the int8 tensor-core decode-MMQ tile — the last big batched-decode chunk (~8.3 ms of the ~25 ms N=8 step was on the bit-exact WS-sw matvec). Measured on Qwen3-8B Q4_K_M @ RTX 4070 Ti (aggregate decode t/s vs the AoS/WS-sw baseline):

| N | baseline (AoS) | SoA reader-port | Δ |
|---|---|---|---|
| 6 | 269.0 | 304.0 | **+13.0%** |
| 7 | 305.1 | 346.9 | **+13.7%** |
| 8 | 327.2 | 382.1 | **+16.8%** (reproduced 324.7 → 379.6) |

Peak VRAM **11394 / 12282 MiB — no spill.**

## Why a reader-port (the two contained approaches failed on 12 GB)

The decode MMQ needs **coalesced staging** → a SoA weight layout. Keeping that SoA as a *companion* beside the interleaved AoS weight ~2.3×'d the Q6_K weight VRAM and **spilled** the 12 GB card at N=8 (325 → 169 t/s). A zero-VRAM **AoS-direct** staging avoided the spill but **regressed −7…−11%** (uncoalesced 210-B block stride). 

The fix: repack each 2-D Q6_K **matmul** weight to the int8 SoA layout once at upload and **free the AoS copy** (exactly like Q4_K's `RepackQ4KSoa`) — so SoA is the only copy (~+0.4 GB net, fits N=8). To free the AoS, **every** Q6_K reader was ported to SoA:

- New `_soa` variants of `llm_matvec_q6k` / `_n2` / `_gemm_n` / `_ws` / `_ws_sw` and `llm_dequant_q6k_to_f16`. Each is **BIT-IDENTICAL** to its AoS counterpart: the SoA stores `(q6−32)` int8 + verbatim scales + d, so `d·scale·(q6−32)` and the per-element reduction order are unchanged.
- Dispatch picks the `_soa` kernel iff the weight handle is SoA-repacked, else AoS.
- The **token embedding** table (read by `embed_lookup`, untied from `output.weight`) is excluded from the repack and stays AoS.
- `SHARPI_Q6K_SOA=0` reverts to AoS everywhere (kill-switch); `SHARPI_Q6K_DECODE_MMQ=0` keeps SoA readers but disables just the decode-MMQ tile.

The decode-MMQ tile itself (per-16 scales via two `m16n8k32` mmas split at the a0/a1|a2/a3 k-boundary, symmetric no-min fixup) is argmax-stable — the same contract as the Q4_K tile.

## Validation

- **Bit-identical** (not just argmax-stable) for every value-carrying reader: `CudaGemmQ6KTests.Q6KSoaReaders_AreBitIdenticalToAos` (MatMul / N2 / Batched / WS / **MatMulBatchedGemm** SoA == AoS, byte-for-byte) + `MatMulBatchedGemm_Q6K_TracksCpuReference`.
- Decode MMQ vs `SimdKernels.DotQ6K`: `CudaDecodeMmqTests.DecodeMmq_Q6K_Soa_TracksCpuReference`.
- e2e: all 8 `CudaBatchForwardMultiTests` + all 5 `CudaSpecBatchVerifyTests`, incl. the **48-token single-user `SpecDecode_GreedyParity_E2E`** (the ported single-user matvec/WS stays byte-stable) and the prefill-parity oracle.
- Release build clean (0 warnings, TreatWarningsAsErrors). All 18 relevant CUDA tests green on the 4070 Ti.

Closes #204. Refs #203 (umbrella). Caps the batched-decode arc: #262 (+11-28%) → #263 (+6.6%) → this (+17% @ N=8).

### DIFF
--- a/src/SharpInference.Cuda/CudaBackend.cs
+++ b/src/SharpInference.Cuda/CudaBackend.cs
@@ -143,8 +143,10 @@ public sealed unsafe class CudaBackend : IComputeBackend, IImageOpsBackend, IDis
     private nint   _matvecQ4KKernel;
     private nint   _matvecQ4KSoaKernel;     // #156: scale-pre-unpacked SoA decode matvec
     private nint   _q4kRepackSoaKernel;     // #156: one-time Q4_K → SoA repack
+    private nint   _q6kRepackSoaKernel;     // #204: one-time Q6_K → SoA repack for decode MMQ
     private nint   _matvecQ5KKernel;
     private nint   _matvecQ6KKernel;
+    private nint   _matvecQ6KSoaKernel;     // #204: bit-identical Q6_K decode matvec over the SoA layout
     // Q4_0 matvec (issue #124, Gemma 4 12B QAT): keeps the q4_0 weights packed on
     // the GPU. Without it q4_0 falls to the F32-dequant upload (~4× VRAM — a 7 GB
     // model would need ~28 GB, defeating full offload). 8 rows/block × 32 thr/row.
@@ -172,6 +174,7 @@ public sealed unsafe class CudaBackend : IComputeBackend, IImageOpsBackend, IDis
     private nint   _matvecQ4KN2SoaKernel;   // #156: N=2 over scale-pre-unpacked SoA weight
     private nint   _matvecQ5KN2Kernel;
     private nint   _matvecQ6KN2Kernel;
+    private nint   _matvecQ6KN2SoaKernel;   // #204: N=2 over the Q6_K SoA layout
     // Issue #111: batched GEMM-N variants — one weight matrix, N input vectors,
     // N output rows in a single launch. Each (row, token) runs the identical
     // per-row reduction as the GEMV so results are bit-identical to N sequential
@@ -182,6 +185,7 @@ public sealed unsafe class CudaBackend : IComputeBackend, IImageOpsBackend, IDis
     private nint   _matvecQ4KGemmNSoaKernel;  // #156: GEMM-N over scale-pre-unpacked SoA weight
     private nint   _matvecQ5KGemmNKernel;
     private nint   _matvecQ6KGemmNKernel;
+    private nint   _matvecQ6KGemmNSoaKernel;  // #204: GEMM-N over the Q6_K SoA layout
     private nint   _matvecQ80GemmNKernel;
     // Issue #194: weight-stationary small-N batched-decode matvecs — token loop inside
     // the block so each weight read is amortized across the batch. One handle per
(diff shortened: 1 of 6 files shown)


pekkah:
Applied the 16-bit aligned `d` load across all 5 Q6_K SoA readers (commit pushed). It's a per-super-block load (amortized 1/256), so e2e-neutral on these bandwidth-bound kernels — consistent with the #145 finding — but it's bit-identical and a clean instruction-count reduction, and the bit-identity + 48-token single-user greedy-parity oracles stay green.

Issue #253, opened 2026-06-15
feat(server): multimodal image input endpoint (Gemma 4 gemma4uv content blocks)


pekkah:
Follow-up to #250 / PR #251. The CLI supports Gemma 4 12B image input (`--image`); the API server does not.

## Scope
Accept image **content blocks** and route them through the existing vision splice:
- [ ] `/v1/messages` (Anthropic): `image` content blocks (base64 `source`).
- [ ] `/v1/chat/completions` (OpenAI): `image_url` parts (data URL / base64; URL fetch optional).
- [ ] Decode → `ImagePreprocessor` → `GemmaUvVisionEmbedder` → splice via `ForwardEmbedding` at the `<|image|>` placeholder positions (reuse the CLI's `RunImagePrompt` logic).
- [ ] Multi-image per message (markers already supported in the CLI path).
- [ ] Smoke tests in `Tests.Server`.

## Dependencies / notes
- `SharpInference.Server` would need a reference to `SharpInference.Vision` and an mmproj-configured engine (new option, e.g. `SHARPI_MMPROJ`).
- Server currently runs the engine on its configured backend; image input needs a pass with `SupportsEmbeddingInput` (CPU or full CUDA today — see the hybrid/Vulkan follow-up).
- Only `gemma4uv` (12B) is wired; reject image content for other models with a clear error.

Issue #93, opened 2026-05-29
Server: surface SHARPI_CPU_MOE as a SharpInferenceServerOptions field (mirror #80)


pekkah:
## Background

After the server refactor (PR landing this session) most CLI MoE-cache knobs reach
the engine via options → env vars in `InferenceEngineLoader.ApplyMoeEnvironment`:
`SHARPI_MOE_WARMPIN`, `SHARPI_MOE_WARMPIN_AFTER`, `SHARPI_MOE_PREDICT_PREFETCH`,
`SHARPI_EXPERT_STATS`. One placement knob is conspicuously missing:
**`SHARPI_CPU_MOE`** — the all-or-nothing override that forces routed experts to
the CPU side.

It matters in practice: per the README perf table, `Qwen3.6-35B-A3B-MTP` only lands
its 22.9 t/s on the CUDA hybrid path when `SHARPI_CPU_MOE=1`. An operator running
the server has to know to export the env var before `dotnet run`, which defeats the
purpose of the options surface that mirrors the CLI for everything else.

## Cause

`SharpInferenceServerOptions` has no field for this; `ApplyMoeEnvironment` never
writes `SHARPI_CPU_MOE`. The engine's `CudaHybridGdnForwardPass` reads the env var
directly at construction time, so the option has to be translated before model load.

## Scope

1. Add `bool? CpuMoe { get; set; }` to `SharpInferenceServerOptions` (nullable so
   default behaviour — engine auto-selects from SLRU sizing — is preserved when
   unset). XML doc cross-references the CLI's `--cpu-moe` (issue #80).
2. In `InferenceEngineLoader.ApplyMoeEnvironment`, when set, write
   `SHARPI_CPU_MOE=0|1` early.
3. One unit test that verifies the option round-trips into the env var ahead of
   model load (matches the existing MoE-knob coverage pattern).

## Acceptance

- [ ] `SharpInference:CpuMoe=true` in `appsettings.json` produces the same forward-
      pass routing as `SHARPI_CPU_MOE=1`.
- [ ] Unset / null preserves the auto-select default (no env-var write).
- [ ] CLI `--cpu-moe` (issue #80) and the server option resolve to the same engine
      behaviour.

## Related

- #80 — CLI `--cpu-moe` flag wrapping the same env var
- `src/SharpInference.Server/InferenceEngineLoader.cs` — `ApplyMoeEnvironment`
[...]

PR #266, opened 2026-06-16
perf(engine): 2-slot prefix cache so short agentic aux requests don't evict the long prefix (#212)

### LINKED ISSUE #212: perf(engine): prefix-cache gets 0 reuse under agentic clients — single-slot KV evicted by interleaved auxiliary requests

pekkah:
## Problem (#212)

Under an agentic client (Claude Code), the single-user `InferenceEngine` got **0 prefix reuse**. Live server trace:

```
request prompt=335tok   (reused=0 ...)            # Claude Code auxiliary request
request prompt=23678tok  (reused=0) prefill=27755ms (853 tok/s) ttft=27958ms
```

The engine keeps **one** previous-sequence slot (`_prevTokens`) and the dense `CudaForwardPass` KV cache holds **one** contiguous sequence. Short interleaved auxiliary requests (topic/title/quota checks, ~335 tokens) overwrite the KV that held the long stable system prefix (~23.7k tokens), so every real turn re-prefills the whole system prompt → **~28 s TTFT on every turn**.

## Fix

An **opt-in 2-slot prefix cache**:
- **Slot 0** = the dense `CudaForwardPass`'s native owned KV cache (full `_maxSeqLen`).
- **Slot 1** = a **bounded scratch** cache (`SHARPI_PREFIX_SCRATCH_TOKENS`, default 4096).

A short aux request is routed to scratch and **leaves slot 0's long prefix intact**, so the next real turn reuses it (TTFT → decode-only). Per request the engine picks the slot with the longest page-aligned shared prefix; on no match, a request whose worst-case footprint (`prompt + maxNewTokens`) fits the scratch cap goes to scratch, otherwise to the full-size owned slot.

### Mechanism (Engine-internal; no `SharpInference.Core` changes)
- New internal `IMultiSlotKvCache` capability; `CudaForwardPass` implements it by reusing the existing owned-cache view + `BindCache` machinery. The only per-slot state on the dense path is `(_gpuKCache, _gpuVCache, _kvLength)` — `_kvCache` is write-only bookkeeping — so a scoped owned-pointer swap is bit-safe.
- **CUDA graphs** bake the owned cache's device pointers, so the captured graph is valid only while slot 0 is active: `_activeSlotIsOwned` gates both graph paths to direct launches during scratch decode. No permanent graph disable → slot 0 keeps its ~9% graph win.
- Gated to the dense `SupportsPartialRewind` path; **GDN/snapshot, MTP, image (#253), TurboQuant, SnapKV, and MoE all stay single-slot**, exactly as before.

### Safety
- **Default OFF** (`SHARPI_PREFIX_SLOTS` unset / ≠ 2) → byte-identical to current behavior; every new branch is guarded by a null/`-1` sentinel.
- A scratch-routed request can never append KV past the scratch allocation: `SelectSlot` bounds routing by `prompt + maxNewTokens ≤ scratchCap` (caught in review — would otherwise be an OOB VRAM write).
- Scratch-alloc OOM degrades to single-slot rather than failing engine construction (auto-context doesn't budget the extra region).

## Knobs
- `SHARPI_PREFIX_SLOTS=2` — enable (default 1 = single-slot).
- `SHARPI_PREFIX_SCRATCH_TOKENS=N` — scratch capacity (default 4096).

## Tests
`InferenceEnginePrefixCacheTests` gains a before/after pair:
- **single-slot**: the aux request evicts the long prefix → **0 reuse** on the next long turn (the bug).
- **2-slot**: the aux request is served from scratch → the next long turn **reuses 48 tokens** of slot 0's prefix.

14/14 green; full solution builds clean under `TreatWarningsAsErrors`.

## Scope / follow-ups (this is the "cheaper mitigation" from #212)
- This is the bounded 2-slot mitigation, not the full vLLM-style paged/multi-prefix cache (two interleaved *large* conversations still thrash slot 0).
- Not GPU-validated end-to-end here (no-CPU-inference / no-GPU-run dev constraint); the default path is inert and shipping-default. A 4070 Ti A/B (slot-0 reuse hits the graph; scratch decode correct via direct launches; multi-turn `reused=` under a real Claude Code session) is the recommended validation before flipping default-on or adding a `SharpInferenceServerOptions` field (mirrors #93).

Closes #212 (mitigation tier).

### DIFF
--- a/src/SharpInference.Engine/CudaForwardPass.cs
+++ b/src/SharpInference.Engine/CudaForwardPass.cs
@@ -830,6 +830,25 @@ public sealed unsafe class CudaForwardPass : IForwardPass, IBatchedForwardPass, 
             // isn't wired here yet (#179). Don't auto-enable eviction on top.
             _snapKvEffectiveBudget = 0;
         }
+        else if (MultiSlotPrefixRequested())
+        {
+            // Issue #212: the operator opted into the multi-slot prefix cache
+            // (SHARPI_PREFIX_SLOTS=2), which keeps a long prefix resident for cross-request
+            // reuse — the opposite of SnapKV's prefill eviction, and incompatible with its
+            // compaction (which discards prefix positions and reindexes the cache so logical
+            // position != physical slot). For the agentic workload #212 targets (a huge system
+            // prefix re-sent every turn) prefix reuse beats SnapKV's memory savings, so let the
+            // multi-slot request suppress the SnapKV AUTO-enable; SupportsMultiSlotPrefix can
+            // then see budget==0 and engage. An explicit SHARPI_SNAPKV_BUDGET>0 still wins above
+            // (IsBudgetExplicit) and keeps SnapKV (with multi-slot disabled). Memory-safe: auto
+            // context sizes _maxSeqLen to fit a full KV cache, so dropping the auto-compaction
+            // can't OOM the base cache (the bounded scratch slot has its own OOM fallback).
+            _snapKvEffectiveBudget = 0;
+            Console.Error.WriteLine(
+                "[CudaForwardPass] SnapKV auto-enable suppressed for the multi-slot prefix cache " +
+                "(SHARPI_PREFIX_SLOTS=2, issue #212). Set an explicit SHARPI_SNAPKV_BUDGET>0 to " +
+                "use SnapKV instead (disables multi-slot).");
+        }
         else
         {
             long fullCacheBytes = (long)_hp.NumLayers * _maxSeqLen
@@ -3521,6 +3540,17 @@ public sealed unsafe class CudaForwardPass : IForwardPass, IBatchedForwardPass, 
 
     /// <inheritdoc/>
     public bool SupportsMultiSlotPrefix => _snapKvEffectiveBudget == 0 && DenseBatchedDecodeSupported();
+
+    /// <summary>
+    /// Issue #212: whether the operator requested the multi-slot prefix cache via
+    /// <c>SHARPI_PREFIX_SLOTS=2</c>. Read in the constructor so the SnapKV auto-enable can defer
+    /// to it (the two are mutually exclusive — SnapKV compaction destroys the positional cache
+    /// invariant prefix reuse relies on). Mirrors the engine's own parse of the same var.
+    /// </summary>
+    private static bool MultiSlotPrefixRequested() =>
+        int.TryParse(Environment.GetEnvironmentVariable("SHARPI_PREFIX_SLOTS"),
+            System.Globalization.NumberStyles.Integer,
+            System.Globalization.CultureInfo.InvariantCulture, out int s) && s == 2;
 
     /// <inheritdoc/>
     public ISequenceKvCache OwnedSlot
--- a/src/SharpInference.Engine/CudaForwardPass.cs
+++ b/src/SharpInference.Engine/CudaForwardPass.cs
@@ -38,7 +38,7 @@ namespace SharpInference.Engine;
 ///   • MoE + TurboQuant combination not validated (no test model has both;
 ///     should compose but the dispatch never sees the combination today).
 /// </summary>
-public sealed unsafe class CudaForwardPass : IForwardPass, IBatchedForwardPass
+public sealed unsafe class CudaForwardPass : IForwardPass, IBatchedForwardPass, IMultiSlotKvCache
 {
     private readonly CudaBackend _gpu;
     private readonly GgufModel _model;
@@ -485,6 +485,16 @@ public sealed unsafe class CudaForwardPass : IForwardPass, IBatchedForwardPass
     // against the owned cache. Never disposed (the owned tensors are freed by Dispose);
     // every layer is marked aliased so even an accidental Dispose can't free them.
     private CudaSequenceKvCache? _ownedCacheView;
+
+    // Issue #212: multi-slot prefix cache (single-user). The engine binds a non-owned
+    // "scratch" slot via ActivateSlot to serve a short interleaved request without evicting
+    // the long resident prefix in the owned cache (slot 0). _activeSlot is the bound slot
+    // (null = owned/no active bind); _savedKvLengthForSlot restores the owned _kvLength on
+    // DeactivateSlot; _activeSlotIsOwned gates the per-token CUDA graph (a captured graph
+    // bakes the owned pointers, so replay is only valid while slot 0 is active).
+    private ISequenceKvCache? _activeSlot;
+    private int _savedKvLengthForSlot;
+    private bool _activeSlotIsOwned = true;
 
     // Ragged attention spill scratch [N × numHeads × maxSeqLen] — the ragged kernel
     // spills per-(sequence, head) score rows when a sequence's length exceeds the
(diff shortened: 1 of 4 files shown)


pekkah:
Thanks — adopted the guard in b6e4f15.

To be precise about severity though: this wasn't actually breaking owned-slot reuse. The owned slot's resident length round-trips through its `CudaSequenceKvCache.Length`: `DeactivateSlot` writes the advanced `_kvLength` into `((CudaSequenceKvCache)_activeSlot).Length` *before* the rewind, and the next `ActivateSlot` reloads `_kvLength = c.Length`. So the next owned-slot request saw the full resident prefix; `_savedKvLengthForSlot` only ever set the (don't-care) field value between requests.

This is covered by `GenerateAsync_TwoSlot_ShortAuxRequestDoesNotEvictLongPrefix`, which drives owned → scratch → owned and asserts the third turn reuses 48 tokens — it passed both before and after this change.

I still applied your suggestion because it makes the invariant clearer (keeps `_kvLength` at the owned slot's advanced length instead of momentarily reading 0) at zero cost.

pekkah:
## End-to-end verification on the server host (RTX 4070 Ti, Qwen3-8B-Q4_K_M, CUDA, single-user `InferenceEngine`)

Reproduced the exact #212 agentic interleave against `SharpInference.Server.Host` via `/v1/chat/completions`: a ~636-token "system prompt" request, then a short 9-token aux request, then the long request again. Per-request `reused=` from the `SharpInference.Engine` Debug trace (`SHARPI_PREFIX_SCRATCH_TOKENS=256`):

**Baseline — single-slot (default, `SHARPI_PREFIX_SLOTS` unset):**
```
[1] LONG prompt=636 (reused=0   new=636) prefill=321ms ttft=331ms
[2] AUX  prompt=9   (reused=0   new=9)   prefill=56ms
[3] LONG prompt=636 (reused=0   new=636) prefill=263ms ttft=263ms   ← aux evicted the prefix → full re-prefill (the bug)
```

**Fixed — 2-slot (`SHARPI_PREFIX_SLOTS=2`):**
```
[1] LONG prompt=636 (reused=0   new=636) prefill=326ms ttft=337ms
[2] AUX  prompt=9   (reused=0   new=9)   prefill=56ms
[3] LONG prompt=636 (reused=624 new=12)  prefill=68ms  ttft=68ms    ← aux served from scratch, prefix preserved → reused 624/636
```

Request 3: **prefill 263ms → 68ms, TTFT 263ms → 68ms**, reuse 0 → 624 tokens. The short aux request was served from the scratch slot and left the long prefix resident in slot 0. Confirmed the slot-0 CUDA-graph path still functions (decode ~65–70 t/s in both).

### ⚠️ Interaction found: SnapKV auto-enable gates multi-slot off
On the server, SnapKV auto-enables for large contexts (`[CudaForwardPass] SnapKV auto-enabled: budget=3020 ...`). `SupportsMultiSlotPrefix` requires `_snapKvEffectiveBudget == 0`, so with SnapKV on the engine **silently runs single-slot** — my first run showed `reused=0` on turn 3 for exactly this reason. The numbers above were taken with `SHARPI_SNAPKV_BUDGET=0`.

This means that in the real #212 scenario (~23.7k system prompt → SnapKV auto-on), the fix does **not** engage out of the box; an operator must set `SHARPI_SNAPKV_BUDGET=0`. SnapKV compacts/evicts prefill KV, which doesn't trivially compose with the slot semantics, so excluding it is the safe conservative gate for this PR — but composing multi-slot with SnapKV (or making the auto-enable defer to multi-slot when `SHARPI_PREFIX_SLOTS=2`) is a worthwhile follow-up, and is the difference between this being opt-in-for-power-users vs. on-by-default-useful for agentic serving.

pekkah:
## Follow-up: multi-slot now engages by default (SnapKV auto-enable defers)

Addressed the SnapKV interaction from the previous comment in 0753016. Rather than require operators to manually set `SHARPI_SNAPKV_BUDGET=0`, the SnapKV **auto-enable** now stands down when `SHARPI_PREFIX_SLOTS=2` is set (a new branch in the auto-enable chain, alongside the existing tq/narrowed defers). Precedence: an explicit `SHARPI_SNAPKV_BUDGET>0` still wins and keeps SnapKV (multi-slot disabled) — opt-in prefix-reuse beats the auto memory-opt, but an explicit SnapKV choice beats both.

Re-verified on Server.Host (4070 Ti, Qwen3-8B-Q4_K_M, CUDA) with **only `SHARPI_PREFIX_SLOTS=2`** (no SnapKV override):
```
[CudaForwardPass] SnapKV auto-enable suppressed for the multi-slot prefix cache (SHARPI_PREFIX_SLOTS=2, issue #212)...
[1] LONG prompt=616 (reused=0   new=616) prefill=332ms ttft=342ms
[2] AUX  prompt=9   (reused=0   new=9)   prefill=58ms
[3] LONG prompt=616 (reused=608 new=8)   prefill=67ms  ttft=67ms   ← reuse, no manual SnapKV disable needed
```

Memory-safe because auto-context already sizes the context window to fit a full (uncompacted) KV cache, so dropping the auto-compaction can't OOM the base cache; the bounded scratch slot keeps its OOM→single-slot fallback. The genuine SnapKV+multi-slot *composition* (per-slot eviction state) remains a larger follow-up, but this makes #212 deliver its TTFT win out-of-box for the agentic workload.

Issue #193, opened 2026-06-10
feat(cuda): packed multi-prompt prefill for continuous batching — PrefillPackedMulti is sequential (#190 follow-up)


pekkah:
## Context

PR #192 landed the core of #190 (CUDA dense continuous batching): per-sequence GPU KV caches (`CudaSequenceKvCache`), `BatchForwardMulti` true batched decode, and the `IBatchedForwardPass` engine wiring. The one remaining item from #190's scope (item 3) is **packed multi-prompt prefill**.

Today `CudaForwardPass.PrefillPackedMulti` is a **per-sequence loop** over `PrefillWithCache` — each prompt's chunk runs as its own batched-trunk forward pass. So when several prompts are admitted concurrently, their prefill GEMMs are *not* amortized across prompts (each re-reads the weights), unlike the CPU `ForwardPass.PrefillPackedMulti` (#183 Gap 2), which packs all chunks into one forward pass.

## Goal

A `cu_seqlens`-style packed prefill: concatenate the S prompts' chunks into one packed batch so every trunk GEMM amortizes the weight read across the combined token count, with per-sequence varlen attention (each chunk attends only to its own cache, no cross-sequence attention, no padding). This mirrors:
- the CPU `ForwardPass.PrefillPackedMulti` (#183 Gap 2), and
- the existing single-sequence chunked batched-trunk (`PrefillBatchedTrunk` / `PrefillBatchChunk`, #136/#162).

## Scope / approach

- Pack chunks token-major into the `_bp*` trunk scratch (offsets per sequence), run the batched RmsNorm/QKV/O/FFN GEMMs over the packed `N = Σ chunk_len`.
- Per-sequence RoPE/QK-norm (each at its chunk's absolute positions) + KV-append into `caches[s]` at `startPos[s]` + varlen attention against `caches[s]` over `[0, startPos[s] + chunk_len[s])` (flash streams prior KV).
- Return per-sequence final-token logits where `wantLogits[s]`.

Dense-only, same gate as `SupportsContinuousBatching`.

Refs: #190 (parent, scope item 3), #183 (CPU `PrefillPackedMulti`), PR #192.

Issue #82, opened 2026-05-29
Add Gemma 4 model family support (E2B / E4B / 26B-A4B / 31B)


pekkah:
> ## Status (2026-06-14)
> **Primary target shipped.** gemma-4-E4B-it **Q8_0 text** decodes coherently on the 12 GB / 64 GB box (README "Gemma 4 E4B" rows; CUDA ~3444 prefill / ~70.5 decode t/s — within ~2× prefill / ~1.1× decode of llama.cpp, #141/#142). The Phase-1 critical-path sub-issues (#83–#87, #89) and the Gemma-3n **PLE** stack (#90) are done; the **12B** variant landed via #124. E2B falls out of the same path.
>
> **Remaining (open sub-issues):**
> - **#88** — 26B-A4B MoE: wire onto expert-offload + validate on 12 GB (deferred; Phase-1 deps now closed → unblocked).
> - **#126** — E4B multimodal (vision) — not started.
> - **#211** — E4B QAT q4_0 GGUF loader gap (`attn_k_norm` layout).
> - **#195** — Gemma 4 continuous batching (multi-user serving).
> - **#152** — remaining prefill MMQ-tiling compute lever · **#151** TC-flash test hardening · **#178** GPU draft-MTP spec decode (now unblocked).
>
> Keeping open as the family-support epic until 26B-A4B + multimodal land.
>
> ---

## Background

Gemma 4 (Google, released 2026-04-02, Apache 2.0) ships in four sizes. **Primary target: E4B at Q8_0** — best fit for the 12 GB VRAM / 64 GB RAM box; ~4.8 GB VRAM weights (PLE pinned to the RAM tier), leaves ~7 GB for KV cache and activations, and Q8 sidesteps the disproportionate Q4 quality hit small models take. 26B-A4B MoE is the **secondary target** once the E-model stack lands; 31B dense is reachable as a Phase-1 logits-parity fixture but only runs via hybrid CPU/GPU offload in practice.

| Variant | Type | Active / total | Context | Modalities | Default quant | VRAM weights | 12 GB / 64 GB fit |
|---|---|---|---|---|---|---|---|
| E2B | dense + 3n stack | 2.3B eff / 5.1B w/ embd | 128K | text, image, audio | **Q8_0** | ~2.5 GB (PLE ~2.8 GB in RAM) | ✅ trivial — falls out of the E4B path |
[...]

PR #273, opened 2026-06-17
feat(cuda): Gemma 4 continuous batching — per-layer head_dim / SWA / shared-KV / PLE (#195)

### LINKED ISSUE #195: feat(cuda): Gemma 4 continuous batching — per-layer head_dim / SWA rings / shared-KV / PLE (#190 follow-up)
## Context

PR #192 (#190) shipped CUDA continuous batching for **dense** transformers only. `CudaForwardPass.SupportsContinuousBatching` excludes Gemma-4-like models (`hp.LayerHeadDim != null`), and the batched entry points throw `NotSupportedException` for them — matching #190's "Gemma 4 per-layer head_dim / k_eq_v stays out of scope initially."

Gemma 4 is a primary target (#82/#124), so batched serving for it is worth a dedicated effort.

## Why it's deferred (the geometry challenges)

`BatchForwardMulti` / `CreateCache` assume uniform geometry and a single full-context cache per layer. Gemma 4 needs:
- **Per-layer `head_dim`** (E4B 256, 12B 256 SWA / 512 global) — the `_bp*`/per-seq view sizing (`qDim`/`kvDim`) becomes per-layer (mirror `GpuLayerBatchedTrunk`'s `layerHd`/`layerKv`).
- **SWA ring caches** — SWA layers use a `pos % SwaRingSize` window ring, not a flat full-context buffer; per-sequence caches must size + index those rings, and batched decode attention must dispatch SWA vs global per layer (`AttentionSwaKv`).
- **Shared-KV aliasing** (`KvSourceLayer`, 12B tail) — per-sequence caches must alias the same way `CudaSequenceKvCache` already supports via its `aliasedLayers` set (currently always empty for dense).
- **`k_eq_v`** (12B global) — V reuses the raw K projection; the batched decode per-seq block must replicate the `kEqV` handling.
- **PLE pre-pass**, post-attn/post-ffn norms, layer_output_scale, final-logit softcap (the softcap is currently a hard exclusion in `SupportsContinuousBatching`).

## Goal

Extend `CreateCache` (per-layer head_dim + SWA ring sizing + aliasing) and `BatchForwardMulti` (per-layer geometry, SWA/global attention dispatch, PLE + sandwich norms + softcap) so Gemma 4 E4B / 12B can batch. The single-sequence `PrefillBatchedTrunk` already does all of this per-token-block; this is the N-sequence decode analogue.

Refs: #190, #124 (Gemma 4 12B), #162/#164 (SWA ring), #136 (batched trunk), PR #192.

pekkah:
## Summary

Closes #195. **Stacked on #193** (PR #272) — base is `feat/193-cuda-packed-prefill` so this diff shows only the Gemma-4 work; rebase to `master` once #272 merges.

PR #192 (#190) shipped CUDA continuous batching for **dense** transformers only; `CudaForwardPass` excluded Gemma-4 models (`SupportsContinuousBatching` / `ThrowIfBatchingUnsupported` threw on per-layer head_dim). Gemma 4 is a primary target (#82/#124), so this adds batched serving for it.

## What changed (`src/SharpInference.Engine/CudaForwardPass.cs`)

- **`FillKvCacheArrays`** — extracts the owned-cache allocation into the single source of truth for per-layer KV geometry (per-layer `head_dim` + KV-head count, SWA ring sizing via `SwaRingSize`, shared-KV tail aliasing via `KvSourceLayer`). Used by **both** the constructor (fills the owned arrays *in place* — identity preserved so `_ownedKCache` stays valid) and **`CreateCache`**, so a per-sequence cache the batched decode / `PrefillWithCache` binds is byte-identical to the owned one. Forward-reference guard + frees partial (non-aliased) allocations on OOM.
- **`CreateCache`** — now allocates Gemma-4 per-layer geometry + SWA rings + shared-KV aliasing (was dense-only).
- **`RunBatchedTrunkGemma4` + `GpuLayerBatchedDecodeGemma4`** — the N-sequence decode analogue of the single-token `RunGemma4DeviceRegion`: batched embed (+scale) + PLE pre-pass, per layer {batched RmsNorm/QKV/QK-norm/V-norm, **per-sequence** RoPE / KV-append / SWA-or-global attention against `caches[n]` at `positions[n]`, batched O-proj + sandwich norms + FFN + PLE injection + `layer_output_scale`}, batched final norm + output GEMM + final-logit softcap into `_decodeLogitsAll`. Matmuls use the proven `GpuMatMulBatched` (cuBLAS GEMM, weight-amortized across the batch); the dense WS/MMQ decode routing for Gemma 4 is a follow-up.
- **Gates** — `SupportsContinuousBatching = dense OR Gemma4BatchedDecodeSupported` (the Gemma-4 gate does **not** exclude softcap, which the decode applies); `ThrowIfBatchingUnsupported` admits Gemma 4. Spec-verify (`SupportsBatchVerify`), the 2-slot prefix cache (`SupportsMultiSlotPrefix`), and the #193 dense packed prefill (`IsDensePackablePrefill`) stay **dense-only** — Gemma 4 prefill uses the sequential `PrefillWithCache` loop (the Gemma-4-capable single-seq batched trunk).
- **Loader** — dense CUDA full-offload already routes via `SupportsContinuousBatching`, so Gemma 4 full-offload now batches automatically. Image input + batching stays rejected; hybrid / Q4_0 fall back to single-user.

## Testing (Gemma 4 E4B Q8_0, RTX 4070 Ti)

E4B exercises per-layer head_dim (256), SWA rings, the 18-layer shared-KV tail, and PLE. New `Gemma4CudaBatchForwardMultiTests`:
- `SupportsContinuousBatching` gate is true for E4B.
- `PrefillWithCache` vs single-user `Prefill` (validates `CreateCache` geometry + aliasing).
- `BatchForwardMulti` N=2 and two decode steps vs single-user `ForwardGemma4` — argmax-stable within the fp16-GEMM tolerance.

Regressions: dense batched + #193 packed (15) and Gemma 4 single-user + prefill trunk (14) all pass — the constructor cache refactor is byte-correct for both families. The only local 12B is Q4_0 (not GEMM-N-batchable → single-user fallback), so k_eq_v's batched path isn't covered by a runnable test; it mirrors `RunGemma4DeviceRegion` exactly.

## Review

Code-reviewer + silent-failure-hunter passes: no critical/high issues. Partial-state (cache `.Length` advanced only by the caller's tail after the trunk completes; caller discards all caches on throw), double-free (aliased-layer skip), and Q4_0-slip-through (dtype gate) all verified safe. Applied the suggested forward-reference guard and fixed two stale comments.

### DIFF
--- a/src/SharpInference.Engine/CudaForwardPass.cs
+++ b/src/SharpInference.Engine/CudaForwardPass.cs
@@ -965,55 +965,15 @@ public sealed unsafe class CudaForwardPass : IForwardPass, IBatchedForwardPass, 
         }
         else
         {
-            // Per-layer KV-cache sizing for gemma4 (LayerHeadDim != null): each layer's
-            // K/V buffer takes its own head_dim, and SWA layers cap context at
-            // SlidingWindowSize. KV-share layers (KvSourceLayer[L] >= 0) alias the
-            // source layer's handle — _kvAliasedLayers tracks them so Dispose skips
-            // the double-free.
-            //
-            // Non-gemma4 path is unchanged: every layer is full-context with the
-            // model-wide head_dim.
-            bool perLayerKv = hp.LayerHeadDim is not null;
-            int swaWindow = hp.SlidingWindowSize > 0 ? hp.SlidingWindowSize : _maxSeqLen;
-            for (int i = 0; i < hp.NumLayers; i++)
-            {
-                int kvSrc = hp.KvSourceLayer is { } ksl ? ksl[i] : -1;
-                if (kvSrc >= 0)
-                {
-                    // Alias the source layer's K/V handles. Source must already be
-                    // initialised — the GGUF puts shared_kv_layers at the tail.
-                    _gpuKCache[i] = _gpuKCache[kvSrc];
-                    _gpuVCache[i] = _gpuVCache[kvSrc];
-                    _kvAliasedLayers.Add(i);
-                    continue;
-                }
-
-                int layerHd = perLayerKv ? hp.LayerHeadDim![i] : _headDim;
-                // Gemma 4 12B mixes per-layer KV head counts (8 GQA on SWA, 1 MQA on
-                // global). Size each layer's K/V by its own count so global layers
-                // don't over-allocate (and match the attention dispatch below).
-                int layerKvHeads = hp.LayerKvHeads is { } lkv ? lkv[i] : _numKvHeads;
-                int layerKvDim = layerKvHeads * layerHd;
-                // SWA layers use a window-sized ring (window + headroom for one batched
-                // append span, issue #162); everything else is full-context. The same
-                // SwaRingSize value is passed to the kernels as maxSeqLen so their
-                // pos % maxSeqLen wraps into this exact ring.
-                int layerCtx = (perLayerKv && hp.IsSwaLayer is { } swa && swa[i])
-                    ? SwaRingSize(_maxSeqLen, swaWindow)
-                    : _maxSeqLen;
-                // q8_0 KV packs 32 elements per block; the store kernels' per-warp amax
-                // reduction (and DTypeInfo.ByteSize's count/32 sizing) assume each layer's
-                // kvDim is a multiple of 32 so blocks never straddle a KV row. Every dense
-                // GGUF head_dim (64/128/256) satisfies this, but fail loud rather than
-                // silently under-allocate + corrupt if a future geometry doesn't (#179).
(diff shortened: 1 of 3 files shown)


pekkah:
Thanks @gemini-code-assist — applied the suggestion: `GpuLayerBatchedDecodeGemma4` now skips creating the `kAll`/`vAll` and per-sequence `kv`/`vv` views on shared-KV tail layers (they're never written/read there), eliding two batched + 2·N per-sequence view alloc/frees per shared layer per decode step. Bit-identical (the elided views were dead); E4B batched-decode oracles still pass. (commit 92cca20)

PR #274, opened 2026-06-17
feat(cuda): SnapKV-aware continuous batching — per-sequence eviction + prefer-batching (#196)

### LINKED ISSUE #196: feat(cuda): SnapKV-aware continuous batching — long-context batched serving silently falls back to single-user (#190 follow-up)
## Context

PR #192 (#190) ships CUDA dense continuous batching. `CudaForwardPass.SupportsContinuousBatching` returns false when SnapKV is active (`_snapKvEffectiveBudget > 0`), because SnapKV eviction physically compacts the **owned** cache during a whole-prompt prefill — there's no per-sequence eviction on the `CudaSequenceKvCache` path.

The catch: **auto-SnapKV is VRAM-scaled and engages at moderate context** (observed at ctx ≳ 2048 for Qwen3-8B on a 12 GB 4070 Ti). When it does, `InferenceEngineLoader` correctly routes the model to the single-user `InferenceEngine` — so **long-context concurrent serving on CUDA silently can't batch** and degrades to serialized single-user decode. The fallback is correct, but the limitation is real and surprising.

## Options

1. **SnapKV-aware per-sequence eviction** — run the SnapKV scoring + compaction per `CudaSequenceKvCache` at the end of each sequence's prefill, so eviction and batching compose. (Mirrors the #130/#134 SnapKV × MTP coexistence work.)
2. **Surface the tradeoff** — when `MaxBatchSize > 1`, prefer batching over auto-SnapKV (or expose a server option to choose), and/or log clearly that auto-SnapKV disabled batching so operators can pick `--kv-type bf16` (narrowed KV disables auto-SnapKV) instead.

Option 2 is the cheap mitigation; option 1 is the real feature.

## Acceptance

Long-context (e.g. 8K+) concurrent serving on CUDA either batches with per-sequence SnapKV eviction, or the operator has a clear, documented knob to trade SnapKV for batching.

Refs: #190, #130/#134 (SnapKV × MTP coexist), #185 (KV budget), PR #192.

pekkah:
## Summary

Closes #196. **Stacked on #195** (PR #273, which is stacked on #193/PR #272) — base is `feat/195-gemma4-cuda-batching`; rebase down the stack as each merges.

`CudaForwardPass.SupportsContinuousBatching` returned false when SnapKV was active, and auto-SnapKV is VRAM-scaled (engages at ctx ≳ 2048 on a 12 GB 4070 Ti) — so long-context concurrent serving on CUDA **silently** fell back to single-user decode. This implements **both** options from the issue.

## Option 1 — per-sequence eviction (the real feature)

`PrefillWithCache` runs the existing owned-cache SnapKV scoring + compaction against the **bound** per-sequence cache (`Prefill` → `ApplySnapKvEviction`), then captures the logical-minus-physical delta onto `CudaSequenceKvCache.EvictedCount` and the post-compaction physical length onto `.Length`, restoring the instance eviction state (isolated per sequence and from the owned cache). The batched decode maps a logical position `p` to the physical slot `p - EvictedCount` for KV append + attention while RoPE keeps `p` — mirroring the single-user owned-cache decode (`kvSlot = position - _kvEvictedCount`). Any evicted cache in a batch forces the per-sequence decode loop (the ragged kernels take `positions[]` for the append slot, so they can't index a compacted cache).

## Option 2 — prefer batching (the safe default)

A new constructor flag `preferBatchingOverAutoSnapKv` (loader passes `MaxBatchSize > 1`) suppresses the VRAM-scaled SnapKV **auto**-enable, so a batching server gets the ragged fast path instead of silent lossy eviction. An explicit `SHARPI_SNAPKV_BUDGET>0` still wins and composes via Option 1. Both modes log clearly.

## Gates

`SupportsContinuousBatching` drops the `budget==0` term — a budget only ever survives for **dense fp32 KV** (the constructor forces it off for Gemma 4 / narrowed KV / TQ), so per-sequence eviction is always wireable there. `ThrowIfBatchingUnsupported` drops the prefill-time SnapKV throw but keeps the decode-only guard against a compacted **owned** cache. Defense-in-depth from review: Gemma-4 decode + packed prefill assert no eviction reached them; the decode guards a negative physical slot; the loader warns when batching was preferred but the model can't actually batch (dense softcap / Q4_0) so the operator can restore KV bounds with an explicit budget or `--kv-type`.

## Testing (Qwen3-8B Q4_K, RTX 4070 Ti)

New `CudaSnapKvBatchingTests`:
- explicit-budget gate flip (`SupportsContinuousBatching` now true),
- N=1 and N=2 (distinct per-sequence deltas) batched decode of evicted caches vs single-user SnapKV decode (argmax-stable; asserts `Length + EvictedCount == prompt.Length`),
- **mixed** batch (one evicted + one un-evicted in the same call),
- Option-2 auto-suppression (with a sanity assert that auto-SnapKV would otherwise engage).

Regressions: dense batched + #193 packed (15) and single-user SnapKV (3) pass; #195 Gemma 4 batched decode (4) still green.

## Review

Two passes (code-reviewer + silent-failure-hunter). Both HIGH latent items hardened: the Gemma-4 decode `anyEvicted` bypass (now a loud assert) and the loader auto-SnapKV-suppressed-but-not-batching footgun (now warns), plus the negative-`physSlot` and packed-prefill SnapKV guards.

### DIFF
--- a/src/SharpInference.Engine/CudaForwardPass.cs
+++ b/src/SharpInference.Engine/CudaForwardPass.cs
@@ -3336,14 +3336,17 @@ public sealed unsafe class CudaForwardPass : IForwardPass, IBatchedForwardPass, 
     /// Whether the dense packed multi-prompt prefill trunk (issue #193) can run this model: the
     /// single-sequence batched-trunk prefill is supported AND none of the Gemma-4 / softcap
     /// features the dense packed trunk omits (PLE, SWA rings, per-layer head_dim, shared-KV,
-    /// softcap) are present. Gemma 4 batches its DECODE (#195) but keeps the Gemma-4-capable
-    /// single-sequence batched trunk for prefill, so it returns false here and
-    /// <see cref="PrefillPackedMulti"/> uses the sequential loop for it. This is exactly the
-    /// negation of the dense-only assertion in <see cref="PrefillPackedTrunkMulti"/>.
+    /// softcap) are present, AND SnapKV is not active. Gemma 4 batches its DECODE (#195) but keeps
+    /// the Gemma-4-capable single-sequence batched trunk for prefill; SnapKV (#196) needs the
+    /// per-token eviction the packed trunk doesn't run. Both fall back to <see cref="PrefillPacked
+    /// Multi"/>'s sequential <see cref="PrefillWithCache"/> loop — which IS SnapKV-aware — instead
+    /// of the dense-only assertion in <see cref="PrefillPackedTrunkMulti"/> (kept as a last-resort
+    /// guard). This is the negation of that assertion.
     /// </summary>
     private bool IsDensePackablePrefill() =>
         !_isGemma4Like && !_hp.HasPerLayerTokenEmbd && _hp.LayerHeadDim is null
         && _hp.SlidingWindowSize <= 0 && _hp.KvSourceLayer is null && _hp.FinalLogitSoftcap == 0f
+        && _snapKvEffectiveBudget == 0
         && IsBatchedPrefillSupported();
 
     private bool AllChunksPackable(ReadOnlyMemory<int>[] chunks, int[] startPos)
--- a/src/SharpInference.Engine/CudaForwardPass.cs
+++ b/src/SharpInference.Engine/CudaForwardPass.cs
@@ -603,7 +603,7 @@ public sealed unsafe class CudaForwardPass : IForwardPass, IBatchedForwardPass, 
     public CudaForwardPass(GgufModel model, CudaBackend gpu, ModelHyperparams hp,
         int maxContextLength = 0,
         bool enableTurboQuant = false, int tqFp32Window = 256, int tqBits = 3,
-        bool? mmqSoa = null)
+        bool? mmqSoa = null, bool preferBatchingOverAutoSnapKv = false)
     {
         _model = model;
         _gpu = gpu;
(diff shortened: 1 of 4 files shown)


pekkah:
Thanks @gemini-code-assist — addressed the review:

**Applied (CudaForwardPass.cs:3267 — packed-prefill gate vs. SnapKV):** agreed the hard throw alone is fragile. `IsDensePackablePrefill()` now also requires `_snapKvEffectiveBudget == 0`, so under an active budget `PrefillPackedMulti` falls back to its sequential `PrefillWithCache` loop (which is fully SnapKV-aware via per-sequence eviction) instead of being able to reach the assertion. The `InvalidOperationException` in `PrefillPackedTrunkMulti` stays as a last-resort guard. (commit cb2531f)

**Declined (lines 4072 / 4101 — per-element `caches[n]` null checks):** the method already does `ArgumentNullException.ThrowIfNull(caches)` on the array, and every other access in `RunBatchedTrunk` / `RunBatchedTrunkGemma4` (e.g. the long-standing `caches[n].K[layer]`) dereferences `caches[n]` directly without a per-element guard. The array is built internally by `ContinuousBatchingEngine` from active sequences whose caches are never null. Adding null checks only in the two new scan loops would be inconsistent with the surrounding code without closing a real gap, so I've left them out to match the established style.

Issue #209, opened 2026-06-11
perf(mtp): k>2 batched-verify headroom - 4-input dense MatVec, WS hybrid-trunk decode kernels, ring/refresh cleanups


pekkah:
## Context

#208 (issues #207/#30) landed MTP k-token batched verify on the GDN-hybrid path at **6.2 → 10.4 t/s (1.68×)** on 27B Q4_K_M CUDA-hybrid — but the measured optimum is the **k=2** verify batch, and the sweep shows why deeper chains don't pay yet:

| draftN (k = draftN+1) | decode t/s | acceptance |
|---|---|---|
| 1 (k=2, 1-slot ring) | **10.4** | 90% |
| 2 (k=3) | 8.5 | 95% |
| 3 (k=4) | 9.2 | 84% |

Per-step cost scaling k=2 → k=4 is ~1.94×: the GPU trunk + lm_head run the temp-free **matvec re-stream** (k weight reads, zero amortization — the MMQ/dequant-GEMM compute kernels' per-call fp16 temps only amortize at prefill N and WDDM-page at decode k, see `MatMulComputeBatchMinN`), and the CPU mmap FFN (46/64 layers, ~6 GB/token — the dominant cost) only **pair**-amortizes via `MatVec2In`. Even at 100% acceptance, k=4 roughly ties k=2.

## Work items (ordered by measured leverage)

1. **4-input dense CPU `MatVec4In`** (SimdKernels, Q4_K/Q6_K/Q5_K/Q8_0 + F32): one weight read per 4 tokens on the CPU FFN → k=4 step drops an estimated ~100 ms → **~12+ t/s (≈2×)**. Keep per-token accumulation order identical to `MatVec2In` so per-position bits stay k-parity-independent (the duplicated-input-tail idiom in both passes' BatchVerify then collapses to one helper — it's currently hand-copied at 3 sites, flagged in the #208 review).
2. **Weight-stationary batched-decode kernels for the hybrid GPU trunk** (the #194 pattern, dense-path proven 2.42× at N=8): removes the linear-in-k re-stream cost on projections + lm_head.
3. **MTP-KV-write-only refresh**: the post-verify refresh re-runs the FULL `MtpForward` (incl. the ~152K×emb lm_head matvec + a blocking 600 KB logits download) once per accepted draft just to rewrite K/V — a head-block-only variant saves ~1-1.6 ms per accepted draft.
[...]

Issue #114, opened 2026-06-03
perf(engine,cpu,cuda): remaining GDN-hybrid prefill headroom after #111/#112 (N-input MoE dots + per-position recurrence/SDPA batching)


pekkah:
@
Follow-up to **#111** (GEMM-batched trunk) and **#112** (dequant-once 2-input routed-MoE dots), both implemented on branch `perf/gdn-batched-prefill-110` (not yet merged).

## Where we are after #111 + #112

512-token prefill chunk, Carnice-Qwen3.6-35B-A3B-APEX-MTP, RTX 4070 Ti, `SHARPI_CPU_MOE=1` (warm):

| phase | #110 baseline | after #111+#112 |
|---|---:|---:|
| trunk (attn+GDN) | 6.8 s | **~1.9 s** |
| routed MoE | 4.7 s | **~3.1 s** |
| total / chunk | 11.8 s | **~5.4 s** |
| **prefill** | ~40 tok/s | **~93 tok/s (~2.3×)** |

Bit-exact (`BatchedPrefill_BitwiseMatchesSequential_Carnice` single + multi-chunk + MTP draft logits), 305/305 non-Vulkan ForwardPass tests green. Two cost centres remain.

## A. Routed MoE is now the dominant cost (~57% of wall, ~3.1 s)

#112 dots each expert-row against its tokens in **pairs** (`DotQ4K_2In` / `DotQ5K_2In` / new `DotQ3K_Q8KS_2In`), so the expensive Q3_K/Q4_K weight unpack is amortized only `decode/2`. Generalize to **N-input** (register-tiled tiles of 4–8 tokens, or true N-input) so the unpack amortizes `decode/T` where T = tokens routing to that expert (~16 avg at N=512).

- Extend the existing 2-input precedents to `DotQ4K_NIn` / `DotQ3K_Q8KS_NIn` (and optionally Q5_K). Keep N FP accumulators register-tiled per tile; each input accumulated in the identical sub-block order so it stays **bit-identical** to N single dots (the `BatchedRoutedExperts` byte-parity oracle must hold — see `feedback_q4k_q8k_no_parity_win`).
- Wire into `BatchedRoutedExperts` Phase A (gate+up) and Phase C (down), iterating the per-expert token list (`expTokI[expStart[e]..]`) in tiles instead of pairs.
- Expected: ~1.2–1.5× on routed MoE.

## B. Residual per-position trunk launches (~1.9 s)

[...]

PR #294, opened 2026-06-18
perf(mtp): fused-scan ring capture for batched verify (#290)

### LINKED ISSUE #290: perf(mtp): fused-scan ring capture for batched verify (#209 item 4)
Follow-up to #209 (work item #4). #209's #1 shipped in #287.

## Problem
`BatchVerify` captures per-token GDN recurrent state into the snapshot ring with `gdnSnapRing=true`, which forces the **per-position recurrence loop** (k×8 launches/layer) plus ~96×(k−1) `CopyDeviceRegion` launches per step (~3 ms at k=4) — instead of the #114-B fused `GdnRecurrenceScan` path the prefill trunk uses.

## Goal
Extend `GdnRecurrenceScan` to optionally dump each token's post-update (scan, conv) state into the ring slots as it scans, so the verify trunk keeps the fused path and drops the per-position relaunch + the bulk `CopyDeviceRegion` fan-out.

## Notes
- Ring slot layout is per-layer `(scan, conv)` after batch token j; the fused scan already produces these states transiently.
- Must preserve rollback correctness (`RestoreBatchSnapshot` reads slot `lengthAfter − startPos − 1`).

pekkah:
Closes #290. Follow-up to #209 (work item #4).

## Problem
`BatchVerify` set `gdnSnapRing=true`, forcing `GdnBlockBatched` into the per-position View loop (k×8 launches/layer via `GdnRecurrenceDecode`) **plus** ~2×(k−1)×numGdn `CopyDeviceRegion` launches (≈96×(k−1) on the 27B-MTP) — just to snapshot each token's GDN `(scan, conv)` state into the rollback ring, instead of the #114-B fused `GdnRecurrenceScan` path the prefill trunk uses.

## Change
Keep the fused path for verify and capture the state **during** the scan:

1. **`llm_gdn_recurrence_scan`** — optional `ring_scan`/`ring_slot_stride`/`n_capture`. In Pass B, where each thread already writes the post-update `state[off] = sij`, it also mirrors `sij` into ring slot `i` for `i < n_capture`. **Zero extra launches**; the scan arithmetic and the live-state evolution stay byte-unchanged (the ring is a disjoint `__restrict__` buffer).
2. **New `llm_gdn_conv1d_state_capture_ring`** — one launch writes every ring slot's conv state (slot `i` = `state_update` with `nTok = i+1`), run **before** the in-place conv state advance so it reads the pre-batch state.
3. **Flat ring** — the snapshot ring is now two contiguous tensors `[slots × numGdn × perLayerFloats]` so the kernels can stride across slots; OOM degrades by retrying fewer slots. `Capture/RestoreGdnRingSlot` re-point to the flat buffer (still used by the `SHARPI_BATCHED_GDN_SCAN=0` fallback + all restores).

Verify always uses the byte-exact scan (never the opt-in chunked-prefill form), so it stays in the same precision class as prefill/`Forward` — **bit-identity holds by composition** over the already-proven #114-B fused path. Rollback correctness is preserved (`RestoreBatchSnapshot` reads slot `lengthAfter − startPos − 1`).

## Tests
- New model-free bit-exactness oracles: `GdnRecurrenceScan_RingCapture_BitwiseMatchesPerSlotState`, `GdnConv1dStateCaptureRing_BitwiseMatchesPerSlotUpdate` (multi-layer ring, non-zero target layer; asserts capture doesn't perturb the scan and that untouched layers stay zero).
- All 10 `CudaGdnBatchedTrunkTests` pass.
- All 3 `CudaMtpBatchVerifyTests` (27B-MTP parity / junk-draft rollback / e2e decode) pass in **both** default and `SHARPI_BATCHED_GDN_SCAN=0` modes.
- Build clean (0 warnings, TreatWarningsAsErrors).

## Perf (RTX 4070 Ti, Qwen3.6-27B-MTP-Q4_K_M, `-g -1`, temp 0)
Warm A/B with greedy decode (identical trajectory ⇒ identical verify workload, isolating the GDN launch overhead): **11.0 vs 10.9 t/s, ~+1–2% e2e**, consistent across runs, never regresses. The e2e delta is small because this config is **CPU-FFN-bound** (44/64 dense FFN layers on CPU mmap); the structural win is **~4× fewer GDN-verify launches** (k=4, 48 layers: ~1824 → ~432), which matters more on GPU-FFN-resident GDN models or higher k.

### DIFF
--- a/src/SharpInference.Cuda/CudaBackend.cs
+++ b/src/SharpInference.Cuda/CudaBackend.cs
@@ -311,6 +311,7 @@ public sealed unsafe class CudaBackend : IComputeBackend, IImageOpsBackend, IDis
     // Issue #114-B: batched GDN trunk + batched-query SDPA kernels.
     private nint   _gdnConv1dDecodeBatchedKernel;
     private nint   _gdnConv1dStateUpdateBatchedKernel;
+    private nint   _gdnConv1dStateCaptureRingKernel;   // #290
     private nint   _gdnL2NormPerHeadBatchedKernel;
     private nint   _gdnTileHeadsBatchedKernel;
     private nint   _gdnRecurrenceScanKernel;
@@ -6003,7 +6004,8 @@ public sealed unsafe class CudaBackend : IComputeBackend, IImageOpsBackend, IDis
         Tensor alphaAll, Tensor betaAll, Tensor ssmA, Tensor dtBias,
         Tensor normWeight, Tensor zAll, Tensor outputAll,
         int numVHeads, int headDim, float normEps,
-        int qStride, int kStride, int vStride, int vHeadOff, int zStride, int oStride, int nTok)
+        int qStride, int kStride, int vStride, int vHeadOff, int zStride, int oStride, int nTok,
+        Tensor? ringScan = null, long ringScanFloatOffset = 0, int ringSlotStride = 0, int nCapture = 0)
     {
         EnsureImageKernels();
         if (!_imageKernelsAvailable)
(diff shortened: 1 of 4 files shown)


pekkah:
Thanks @gemini-code-assist. Assessed against the build + test run; no code changes needed:

- **`CudaTextKernels.cs:8300` (`#pragma unroll 4`)** — kept for consistency: the two sibling conv kernels immediately above (`llm_gdn_conv1d_decode_batched`, `llm_gdn_conv1d_state_update_batched`) use the identical directive over the same `for (r < retained)` loop. `retained = kernel_size-1` is ≤3 for our K=4 models, so NVRTC takes it as an unroll hint and compiled cleanly — the new `GdnConv1dStateCaptureRing_BitwiseMatchesPerSlotUpdate` test exercises the kernel with no warnings.
- **Tests `:320` / `:375` (claimed CS1503)** — false positives. C# array element access permits `int`, `uint`, `long`, and `ulong` index expressions, so `ring[baseOff + e]` with `long` operands is legal. The full solution builds with 0 errors / 0 warnings (TreatWarningsAsErrors) and both new tests pass, which confirms there is no CS1503.

Issue #132, opened 2026-06-04
InferenceEngine.Dispose doesn't drain in-flight generation worker → 0xC0000005 on shutdown mid-decode


pekkah:
### Summary

`InferenceEngine.Dispose` returns without waiting for the inner `Task.Run` generation worker (spawned by `GenerateAsync`) to exit. If host shutdown cancels mid-decode and then the engine's owned `IForwardPass` is disposed, the still-running worker keeps touching `PagedKvCache` while its native memory is being freed → access violation (0xC0000005).

### Repro

1. Build a host that owns an `InferenceEngine` over a `CudaForwardPass` and starts a `GenerateAsync` enumerator.
2. Ctrl-C (cancel the host) while a turn is mid-decode.
3. Host disposes engine → engine disposes forward pass → worker still in flight references freed `PagedKvCache` → 0xC0000005.

### Expected

`Dispose` (or `DisposeAsync`) drains/cancels the worker and joins it before freeing engine-owned native resources.

### Workaround in calling code

Skip `_engine.Dispose()` entirely on shutdown and let process exit reclaim the native memory. From `Ayu`'s `SharpInferenceLanguageModel`:

```csharp
public ValueTask DisposeAsync()
{
    // Intentionally NOT calling _engine.Dispose() — see this issue.
    return default;
}
```

This is obviously not a real fix; it leaks the deterministic teardown that consumers expect.

### Suggested fix

`InferenceEngine.Dispose` should:
1. Signal cancellation on the internal generation cancellation source.
2. Await/Wait the worker task to completion (with a short timeout for safety).
3. Then dispose owned resources.

Probably easier as `DisposeAsync` since `Wait()` from a finalizer/dispose path on Windows can deadlock with SynchronizationContext capture.

### Impact

Any consumer that disposes the engine while a generation is in flight (i.e. graceful Ctrl-C of an OpenAI-style server, or a Generic Host's `StopAsync` cancelling mid-stream) gets a process crash instead of a clean shutdown. Reproducible on `SharpInference 0.6.1-alpha.0.8` with `CudaForwardPass` + Qwen3-8B Q4_K_M.

pekkah:
[...]

PR #298, opened 2026-06-18
feat(server): opt-in fast-reject for overlapping requests (#109)

### LINKED ISSUE #109: Single-user InferenceEngine deadlocks on concurrent client requests (Claude Code agent loop hang)
## Summary

When the server runs with the single-user `InferenceEngine` (i.e. `SHARPI_MAX_BATCH` unset / ≤ 1) and an agentic client fires overlapping requests, a request can deadlock mid-decode: token generation freezes and `active_requests` stays pinned at 1 indefinitely.

## Repro (observed)

- Model: Carnice (Qwen3.6-35B-A3B-MTP), CUDA hybrid, `SHARPI_CPU_MOE=1`, Release build, `SHARPI_NO_THINKING=1`.
- Driver: Claude Code CLI pointed at the local server (`ANTHROPIC_BASE_URL` → our `/v1/messages`), agent loop ("write hello world to a temp file").
- A single agent turn completed fine (~18.6s). On a subsequent turn, a request hung: `tokens_generated_total` frozen at 149, `/metrics` `active_requests=1` stuck for ~24 min, no further output, file never written.

Claude Code appears to fire requests near-simultaneously; the single-user engine doesn't serialize/queue the overlap cleanly.

## Workaround

Set `SHARPI_MAX_BATCH>1` to select `ContinuousBatchingEngine`, which is built for concurrent sequences.

## Suspected area

Single-user `InferenceEngine` request handling under concurrent `GenerateChunksAsync` calls — likely a shared-state / cache reentrancy issue rather than true CPU deadlock (no progress, not a livelock). Needs a deterministic repro (two overlapping HTTP requests against `InferenceEngine`) and a guard: either serialize requests with a clear queue, or reject/409 the second concurrent request with a helpful message pointing at `SHARPI_MAX_BATCH`.

## Notes

- Pre-existing; not introduced by the Gemma 4 / v0.6.0 work. Filed as a known issue at v0.6.0.

pekkah:
## Summary

Resolves #109 by adding the issue's explicitly-requested second guard for the single-user engine: **opt-in fast-rejection of overlapping requests with HTTP 429**, pointing at `SHARPI_MAX_BATCH`.

### Background

The single-user `InferenceEngine` already serializes overlapping requests on an internal gate (the deadlock/corruption fix in `4033f0c`, plus the dispose-drain in #132 — all with passing regression tests). That is *correct*, but for an agentic client firing near-simultaneous requests (the issue's exact scenario — Claude Code), a concurrent request silently blocks for the full duration of the in-flight one, which reads as an indefinite hang with `active_requests` pinned at 1.

The issue listed two guard options — "serialize **or** reject/409." The serialize path exists; this PR adds the reject path as **opt-in**.

### Changes

- **`SharpInferenceServerOptions.MaxConcurrentRequests`** (`int?`, default `null`; env `SHARPI_MAX_CONCURRENT`). Unset = legacy serialize-and-queue (unchanged). Set to `1` to make the single-user limit explicit.
- **`RequestConcurrencyGate`** — a `SemaphoreSlim` wrapper with non-blocking `TryEnter()`, registered as a lazily-resolved DI singleton (disabled/passthrough when the option is unset).
- **`WithConcurrencyLimit()` endpoint filter** applied to the three generation endpoints (`/v1/chat/completions`, `/v1/messages`, `/v1/responses`). Overflow → `429` + `Retry-After` + a message pointing at `SHARPI_MAX_BATCH`, via `TypedResults.Json` (source-gen, AOT-safe). The filter brackets the whole request — including a streaming SSE response, since `await next()` completes only when the handler's Task finishes — and releases the slot in a `finally`. `/v1/models`, `/health`, `/metrics` stay ungated.
- Host maps `SHARPI_MAX_CONCURRENT` → the option.

### Why opt-in

Default-`null` keeps the exact current behavior, so existing deployments (and `ContinuousBatchingEngine`, which serves concurrency itself) are unaffected. An operator running single-user behind an agentic client sets one flag to turn the looks-like-hang into an immediate, actionable signal.

### Tests / verification

- New `ConcurrencyLimitTests`: 429-on-overflow (+ slot frees afterward), passthrough-when-unset (overlapping requests both succeed), and non-generation endpoints staying reachable while a slot is held.
- Full `SharpInference.Tests.Server` (121) passes; Release build clean under `TreatWarningsAsErrors`; **NativeAOT publish succeeds** (the host is AOT-published).
- The pre-existing engine-level concurrency/deadlock regression tests (`InferenceEngineConcurrencyTests`, 6) still pass — this PR is additive and doesn't touch the engine serialization.

Closes #109.

### DIFF
--- a/src/SharpInference.Server.Host/Program.cs
+++ b/src/SharpInference.Server.Host/Program.cs
@@ -24,6 +24,12 @@ builder.Services.AddSharpInference(builder.Configuration, opts =>
     if (int.TryParse(Environment.GetEnvironmentVariable("SHARPI_MAX_BATCH"), out int maxBatch) && maxBatch > 0)
         opts.MaxBatchSize = maxBatch;
 
+    // Fast-reject overlapping requests with HTTP 429 instead of serializing them on the
+    // single-user engine (issue #109). Set to 1 for a single-user server fronting an agentic
+    // client; unset keeps the legacy serialize-and-queue behaviour.
+    if (int.TryParse(Environment.GetEnvironmentVariable("SHARPI_MAX_CONCURRENT"), out int maxConcurrent) && maxConcurrent > 0)
+        opts.MaxConcurrentRequests = maxConcurrent;
+
     // Continuous-batching scheduling knobs (issue #183): prefill chunk size (Gap 1)
     // and KV admission budget in MiB (Gap 3). Same precedence as SHARPI_MAX_BATCH.
     if (int.TryParse(Environment.GetEnvironmentVariable("SHARPI_PREFILL_CHUNK"), out int prefillChunk) && prefillChunk >= 0)
--- a/src/SharpInference.Server/Endpoints/AnthropicEndpoints.cs
+++ b/src/SharpInference.Server/Endpoints/AnthropicEndpoints.cs
@@ -15,7 +15,7 @@ public static class AnthropicEndpoints
 {
     public static IEndpointRouteBuilder MapAnthropicEndpoints(this IEndpointRouteBuilder app)
     {
-        app.MapPost("/v1/messages", HandleMessages);
+        app.MapPost("/v1/messages", HandleMessages).WithConcurrencyLimit();
         return app;
     }
 
--- a/src/SharpInference.Server/Endpoints/ConcurrencyLimit.cs
+++ b/src/SharpInference.Server/Endpoints/ConcurrencyLimit.cs
@@ -12,7 +12,7 @@ namespace SharpInference.Server.Endpoints;
 /// <see cref="SharpInferenceServerOptions.MaxConcurrentRequests"/> is set, this caps the number
 /// of in-flight generation requests and the endpoints fast-reject the overflow with HTTP 429.
 /// </summary>
-internal sealed class RequestConcurrencyGate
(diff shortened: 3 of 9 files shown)


pekkah:
Addressed the review feedback:

- **IDisposable + null-safe accessors** — adopted (gate now disposes the `SemaphoreSlim` on host shutdown; `TryEnter`/`Exit` are null-safe).
- **Premature-release / slot-leak on the `IResult` path** — rather than keep the `GateReleasingResult` decorator (which, as the re-review notes, can be skipped on a client abort / pipeline short-circuit and leak the slot), I removed it. The gated generation handlers (`/v1/chat/completions`, `/v1/messages`, `/v1/responses`) write the response and return a bare `Task`, so `await next(ctx)` completes only after the streaming response is fully written, and the plain `try/finally` releases the slot exactly once on every path — normal completion, an exception out of `next()`, and a mid-stream client abort (the streaming handlers catch the disconnect and return normally). The `Response.OnCompleted` alternative is documented in a comment as the path to take if a handler ever returns an `IResult` instead.

Verified: full server suite (121) green, Release build clean under TreatWarningsAsErrors, NativeAOT publish succeeds.

PR #301, opened 2026-06-19
feat(gemma4): agentic 12B download preset + chat-template BOS fix + --thinking

pekkah:
## Summary

Adds a download preset for yuxinlu1's **Gemma 4 12B agentic finetune** and fixes two real chat-template bugs that smoke-testing it surfaced — both affecting models well beyond this one.

## Changes

**1. `gemma4-12b-agentic` download preset** (`scripts/download-model.ps1`)
Verified the repo's `/tree/main`: the placeholder `gemma4-agentic-Q4_K_M.gguf` does not exist — the real files are `gemma4-v2-<quant>.gguf`. Points `Files`/`Urls` at `gemma4-v2-Q4_K_M.gguf` (~7.4 GB), text-only (no mmproj), with a note that it's a reasoning finetune (use `--thinking`, recommended `--temp 1.0 --top-k 64 --top-p 0.95`).

**2. Missing BOS in the chat-template path** (`Core/JinjaChatTemplate.cs`, `Core/GgufTokenizer.cs`)
Chat templates that open with `{{- bos_token -}}` (Gemma, Llama, Mistral) rendered it as an empty string because the BOS string was never passed to the Jinja renderer — so every such prompt shipped with **no BOS token**. Gemma is BOS-sensitive and degenerates without it (off-topic output / leaked `<channel|>` / empty generations). `GgufTokenizer.Encode` never compensated.

Fixed centrally: seed the BOS string into `JinjaChatTemplate.BosToken` when the tokenizer builds the template, so **both the CLI and the server's `ChatTemplateRenderer`** (which share the same instance) get it. Gated on `add_bos_token`, so `add_bos_token=false` models (e.g. Qwen) are byte-identical.

**3. `--thinking` flag** (`Cli/RunCommand.cs`)
The CLI hard-forced `enable_thinking=false` for all `gemma4` (correct for Google's stock instruct models, which aren't reasoning-trained). Reasoning finetunes share the identical arch/template/tokens/sampling-hints — nothing in the GGUF metadata distinguishes them — so add an explicit `--thinking` opt-in. Default stays off; `--no-thinking` still wins.

## Result

With both fixes the agentic finetune generates correctly — clean answers in the default mode and a coherent `<|channel>` thought chain under `--thinking`. Stock `gemma-4-12b-it` regression-checked (clean, unchanged).

## Testing

- Full Release build clean (0 warnings; `TreatWarningsAsErrors`).
- `Tests.Core` 152 (incl. 4 new `bos_token` seeding tests), `Tests.Server` 124, `Tests.Cli` 32 — all green.
- Manual on RTX 4070Ti (GPU full-offload): agentic finetune + stock 12B-it, both APIs of the rendering path.

## Notes / scope

- Server's per-request thinking default is unchanged (it already has request-level control); the BOS fix covers the server via the shared template instance.
- The finetune's reasoning mode is strong in-domain (coding) but can ramble out-of-domain at temp 1.0 — default mode is the reliable path for simple Q&A.

https://claude.ai/code/session_01AseYLmrS8RaYpAT7dhMXL7

### DIFF
--- a/scripts/download-model.ps1
+++ b/scripts/download-model.ps1
@@ -5,7 +5,7 @@
     Downloads from HuggingFace to the models/ directory. Skips if already present.
     Supports: smollm2, vibethinker, qwen3-8b, olmoe-1b-7b, llama31-70b, qwen3-coder-30b-a3b, qwen36-35b-a3b,
               qwen36-27b-mtp, qwen36-27b-mtp-q5, qwen36-35b-a3b-mtp, carnice-35b-a3b-mtp,
-              gemma4-12b-qat, gemma4-12b-q4km, gemma4-e4b-qat,
+              gemma4-12b-qat, gemma4-12b-q4km, gemma4-e4b-qat, gemma4-12b-agentic,
               llama4-scout, z-image-turbo, z-image-turbo-q8, realesrgan-x4
 .PARAMETER Model
     Which model to download. Default: downloads all text models (skips large image models).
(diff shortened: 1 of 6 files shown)


pekkah:
## Review cycle 1

Ran internal review (code-reviewer + silent-failure-hunter) and Gemini.

**Addressed (bdc3e01):** `--thinking` precedence was implicit and the `--thinking`+`--no-thinking` conflict resolved silently. Extracted a pure, unit-tested `RunCommand.ResolveThinkingOff(arch, thinking, noThinking)` helper and added a warning when both flags are passed. +9 precedence tests (Tests.Cli 32→41). Behavior unchanged for every existing combination.

**Verified non-issues:** no double-BOS (the chat-template path is the only BOS source; raw-prompt path is separate and guarded); byte-identical for `add_bos_token=false` models (Qwen) and templates that don't reference `bos_token` (covered by tests); llama/mistral that *do* open with `{{ bos_token }}` now correctly get BOS (a fix, not a regression).

**Deferred as out-of-scope follow-ups** (pre-existing, not introduced here):
1. `GgufTokenizer` swallows template-parse failures in a bare `catch {}` → silent fallback to a hardcoded template. Narrow + log.
2. Contradictory GGUF metadata (`add_bos_token=true` but invalid/non-control BOS id) silently yields a BOS-less or mis-tokenized prompt. Add a load-time guard/log.
3. Server `ResponsesEndpoints` hardcodes thinking off for gemma4 with no per-request override (OpenAI/Anthropic endpoints already honor it) — needed for the new reasoning finetune across all endpoints.

Tests green: Core 152, Server 124, Cli 41. CI (GitGuardian) green.

PR #305, opened 2026-06-20
fix(gemma4): stop <|channel> thought markers leaking into text output (#304)

### LINKED ISSUE #304: fix(gemma4): `<|channel>thought<channel|>` channel header leaks into text output (not recognized as reasoning)
## Summary

When driving **Gemma 4 12B (QAT q4_0)** through an in-process **agentic tool-call loop** (record the structured assistant `tool_calls` turn + `adapter.RenderToolResult(...)`, then re-infer), the model's answer on the post-tool-result pass is wrapped in a Gemma channel header that **leaks into the user-facing `GenerateChunkKind.Text` stream**:

```
<|channel>thought<channel|>The Program.cs file reads the following environment variables: ...
```

The markers use the same asymmetric-pipe convention as the tool markers — open `<|channel>`, close `<channel|>` — with a channel label (`thought`) between them, followed by the actual content.

Observed on **every** pass that re-infers after a tool result (read_file summary, write_file confirmation, edit_file confirmation). The chat template was rendered with `enable_thinking=false`. Normal (non-tool) single-pass replies do **not** show this, so it surfaces specifically once history contains an assistant `tool_calls` turn + a `{role:tool}` result and the model is asked to continue.

## Root cause (hypothesis)

`InferenceEngine`'s reasoning split is keyed on the `<think>` / `</think>` **token IDs** (`thinkTokenId`/`endThinkTokenId`). Gemma 4 emits a `<|channel>label<channel|>` construct instead, so the channel header is never recognized as reasoning and is surfaced verbatim as `Text`. `Gemma4ToolCallAdapter.Parse` already scrubs the sibling `<|tool_response>` / `<tool_response|>` turn markers (#150) but has no knowledge of `<|channel>` / `<channel|>`.

## Suggested fix

Handle Gemma 4's channel markers in one of:
- **`Gemma4ToolCallAdapter`** — scrub `<|channel>label<channel|>` headers from `PlainText` (mirrors the existing `<|tool_response>` scrub), and expose them to the streaming state machine so the header isn't emitted as a content delta; or
- **the engine's reasoning split** — recognize Gemma 4's channel format so a `thought`/reasoning channel is routed to `GenerateChunkKind.Thinking` (and suppressed under `enable_thinking=false`), while a `final`/answer channel routes to `Text`.

In the observed output the content *after* `<|channel>thought<channel|>` was the actual final answer (no separate `final` channel was emitted), so at minimum the header markup must not reach `Text`; keep the content.

## Impact / workaround

Downstream (Ayu, a voice assistant) the leaked `thought` label + angle-bracket markup would be read aloud. Worked around there for now with a stream-safe Ayu-side scrubber that strips `<|channel>…<channel|>` headers, but the natural home is the adapter/engine since it's Gemma 4 wire-format knowledge.

## Repro sketch

1. Load Gemma 4 12B (arch `gemma4`), `enable_thinking=false`.
2. Render history: system + user + assistant(`tool_calls:[read_file]`) + `{role:tool, content, tool_call_id}` via `Gemma4ToolCallAdapter.RenderToolResult`.
3. `add_generation_prompt=true`, generate.
4. The first `Text` tokens are `<|channel>thought<channel|>` before the answer.

pekkah:
Fixes #304.

## Problem

Driving **Gemma 4 12B (QAT q4_0)** through an in-process agentic tool-call loop leaked the model's `<|channel>thought<channel|>` reasoning-channel header into the user-facing `GenerateChunkKind.Text` stream. The issue + its comment surface three coupled problems in the same scenario:

1. **Channel markers leak into `Text`** (the titled bug) — and a *bare* `<channel|>` close can appear with no preceding open.
2. **Generation doesn't stop** at the tool-call boundary — this QAT line opens a `<|tool_response>` turn (not in the EOG set) instead of `<end_of_turn>`, so it runs on and hallucinates the channel tail.
3. **`StopTokenIds` is a replace-not-augment footgun** — `sp.StopTokenIds ?? EogTokenIds` silently drops EOG when a consumer adds a stop.

## Root cause of (1)

The reasoning split is keyed on token IDs the engine is *told* about. `<|channel>` / `<channel|>` **are** single special tokens in the Gemma 4 vocab (verified: ids **100 / 101**, type 4 USER_DEFINED), and the server + CLI each resolved them — but `InferenceEngine`'s convenience constructor hardcoded the split **off** (`thinkTokenId: -1`), so an in-process host (e.g. Ayu) never got it. The boundary handling also *emitted* an orphan/bare close as literal text.

## Fix

**Core**
- `ITokenizer.ReasoningTokens` (default `(-1,-1)`); `GgufTokenizer` resolves `<think>`/`</think>` then `<|channel>`/`<channel|>` — one source of truth for engine, loader, and CLI.
- `IToolCallAdapter.ToolBoundaryStopMarkers` (default `[]`); Gemma 4 returns `<|tool_response>`.
- `Gemma4ToolCallAdapter.Parse` scrubs `<|channel>label … <channel|>` blocks (incl. bare close / unterminated open) from `PlainText`, mirroring the GGUF chat template's own history scrubber.

**Engine**
- Convenience constructor auto-resolves the split from `tokenizer.ReasoningTokens` (fixes the in-process leak; explicit `-1` still forces it off).
- Reasoning boundary tokens are **always consumed, never emitted** — a bare `<channel|>` close (which the post-tool generation prompt can prime) is swallowed, not leaked. Mirrored in `ContinuousBatchingEngine`.
- `SamplingParams.AdditionalStopTokenIds` — always unioned with the effective stop set (footgun-free way to add a stop); `StopTokenIds` replace semantics documented.

**Server**
- Loader resolves the adapter's tool-boundary markers → ids, stashed on `ChatTemplateRenderer`; the OpenAI + Anthropic chat endpoints add them to the stop set on tool-active turns so generation halts the instant the tool calls finish.

## Compatibility

All default paths are byte-identical: no reasoning tokens → `(-1,-1)`; null `AdditionalStopTokenIds` → base stop set unchanged; non-Gemma adapters → no tool-boundary markers.

## Tests

- **Core** — channel scrub shapes (header-only, with reasoning content, bare close, unterminated open, after a tool call, untouched inside a tool arg) + tool-boundary markers; model-free `ResolveReasoningTokens` (think/channel/precedence/zero-id/partial-pair).
- **Engine** — bare-close suppression, convenience-ctor auto-resolve, additive-stop union + EOG retention.
- **Server** — renderer tool-boundary stash/clear.

Build clean (warnings-as-errors). Model-free suites green (Core 169, Server 125, Cli 41) + mock-based engine tests (50). The heavy GPU/CPU model matrix was not run locally; channel behavior is unit-tested against the real Gemma 4 vocab ids (100/101).

https://claude.ai/code/session_01KDR5BLSrhvCk7URYqrp6zA

### DIFF
--- a/src/SharpInference.Cli/RunCommand.cs
+++ b/src/SharpInference.Cli/RunCommand.cs
@@ -1545,18 +1545,28 @@ public sealed class RunCommand : Command<RunCommand.Settings>
     /// </summary>
     private static bool EmitToken(int next, GgufTokenizer tok, Utf8StreamDecoder streamDec, ref bool inThinking, bool hideThinking = false)
     {
+        // Reasoning boundary tokens are always consumed (never printed), with the state flip
+        // gated on the current mode so a malformed double-open or a bare close — e.g. Gemma 4's
+        // post-tool prompt primes <|channel>, so the answer pass emits a lone <channel|> with no
+        // open — is swallowed rather than rendered as a literal marker (issue #304).
         if (next == s_thinkTokenId)
         {
-            inThinking = true;
-            // No trailing \n: the model often emits its own leading newline inside the block,
-            // and a double break before the reasoning starts looks noisy.
-            Console.Write("\x1b[2m[Thinking...] ");
+            if (!inThinking)
+            {
+                inThinking = true;
+                // No trailing \n: the model often emits its own leading newline inside the block,
+                // and a double break before the reasoning starts looks noisy.
+                Console.Write("\x1b[2m[Thinking...] ");
+            }
             return false;
         }
-        if (next == s_endThinkTokenId && inThinking)
+        if (next == s_endThinkTokenId)
         {
-            inThinking = false;
-            Console.Write("\x1b[0m\n");
+            if (inThinking)
+            {
+                inThinking = false;
+                Console.Write("\x1b[0m\n");
+            }
             return false;
         }
         // Stream through the same UTF-8 decoder regardless of mode so multibyte
--- a/src/SharpInference.Cli/RunCommand.cs
+++ b/src/SharpInference.Cli/RunCommand.cs
@@ -347,24 +347,10 @@ public sealed class RunCommand : Command<RunCommand.Settings>
         var tokenizer = GgufTokenizer.FromGgufModel(model);
         s_jinja = tokenizer.ChatTemplate;
 
-        // Reasoning models (Qwen3, DeepSeek-R1, SmolLM3, ...) register <think>/</think>
-        // as control tokens in their GGUF. The decode loops no-op when these IDs are -1.
-        if (tokenizer.SpecialTokens.TryGetValue("<think>", out int thinkId)
-            && tokenizer.SpecialTokens.TryGetValue("</think>", out int endThinkId)
-            && thinkId > 0 && endThinkId > 0)
-        {
-            s_thinkTokenId = thinkId;
-            s_endThinkTokenId = endThinkId;
-        }
-        // Gemma 4 brackets its reasoning in <|channel>thought … <channel|> instead. Route it
-        // through the same think/end-think machinery so the markers don't leak into output.
-        else if (tokenizer.SpecialTokens.TryGetValue("<|channel>", out int channelId)
-            && tokenizer.SpecialTokens.TryGetValue("<channel|>", out int endChannelId)
-            && channelId > 0 && endChannelId > 0)
-        {
-            s_thinkTokenId = channelId;
-            s_endThinkTokenId = endChannelId;
-        }
+        // Reasoning boundary tokens: ChatML <think>/</think> (Qwen3, DeepSeek-R1, SmolLM3, ...)
+        // or Gemma 4's <|channel>thought … <channel|>. Resolved once on the tokenizer so the CLI,
+        // server, and engine share one definition. The decode loops no-op when these IDs are -1.
+        (s_thinkTokenId, s_endThinkTokenId) = tokenizer.ReasoningTokens;
 
         // Resolve reasoning on/off (see ResolveThinkingOff for the full precedence). --no-thinking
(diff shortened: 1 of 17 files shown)


pekkah:
Addressed the review feedback across two cycles (internal review agents + Gemini Code Assist):

**Internal review (correctness / silent-failure):**
- `ContinuousBatchingEngine.AdmitOne` routed the **first** sampled token through the old guarded boundary form — a bare `<channel|>` first token (or a double-open) still leaked on the batched path. Migrated to the same always-consume contract as the decode loop.
- CLI `EmitToken` had the same orphan-close gap (a bare close while `!inThinking` printed the literal marker). Now always-consumed; double-open is idempotent.
- Loader now warns (stderr, mirroring the image-input diagnostic) when an adapter declares tool-boundary markers that don't resolve to a vocab id, instead of silently re-opening the run-on symptom.
- Documented the OpenAI-vs-Anthropic tool-boundary-stop gate difference (each is scoped to that endpoint's own tool-aware rendering condition).

**Gemini Code Assist:**
- Added `?? throw new ArgumentNullException` guards to the `InferenceEngine` convenience ctor (it now dereferences `tokenizer.ReasoningTokens` in the base initializer).
- Simplified `ResolveStopSet`'s `HashSet` seeding to drop the `IEnumerable<int>` cast.

Build clean (warnings-as-errors); model-free suites green (Core 169, Server 125, Cli 41) + mock-based engine tests. The heavy GPU/CPU model matrix wasn't run locally; channel behavior is unit-tested against the real Gemma 4 vocab ids (`<|channel>`=100 / `<channel|>`=101). The `ContinuousBatchingEngine` + CLI boundary fixes are 1:1 mirrors of the engine logic covered by `GenerateChunksAsync_BareCloseWithNoOpen_IsSwallowedNotLeaked`.

PR #306, opened 2026-06-20
perf(cli): partial top-k for --verbose-prompt instead of full-vocab sort (#155)

### LINKED ISSUE #155: cli: --verbose-prompt does a full-vocab LINQ sort per decode token
## Problem

`RunCommand.DecodeLoop` (`src/SharpInference.Cli/RunCommand.cs`, ~line 1025) logs the top-5 next-token candidates every decode step when `--verbose-prompt` is set:

```csharp
var logitsArr = logits.ToArray();                       // ~1 MB alloc + copy (Gemma 4 vocab = 262144)
var top5 = Enumerable.Range(0, logitsArr.Length)
    .OrderByDescending(j => logitsArr[j]).Take(5) ...    // full-vocab sort, every token
```

For a 262144-token vocab this is a full O(V·logV) LINQ sort plus a 1 MB allocation **per token** — ~1.5–2.5 ms/token of CPU overhead. It badly skews decode t/s for anyone benchmarking or generating with `--verbose-prompt` on.

Surfaced while diagnosing #142 — `scripts/bench-textgen.ps1` hard-coded `--verbose-prompt`, understating decode by 10–25%. The bench was fixed in #154 (flag removed), but the underlying logging path is still O(V·logV)/token for direct CLI users.

## Fix

Replace the `OrderByDescending().Take(5)` with a partial top-5 selection (single O(V) pass keeping the 5 best, like `Sampler` does for top-k) and drop the `logits.ToArray()` copy — read the span directly. This makes `--verbose-prompt` a near-free debug aid instead of a per-token full sort.

Low priority (debug flag only), but cheap and removes a foot-gun for benchmarking.

pekkah:
Fixes #155.

## Problem

`RunCommand.DecodeLoop` logs the top-5 next-token candidates every decode step when `--verbose-prompt` is set, via:

```csharp
var logitsArr = logits.ToArray();                                  // ~1 MB copy (V=262144)
var top5 = Enumerable.Range(0, logitsArr.Length)
    .OrderByDescending(j => logitsArr[j]).Take(5) ...              // full O(V·logV) sort, per token
```

That's a ~1 MB allocation plus a full-vocab sort **per decode token** (~1.5–2.5 ms/token on Gemma 4's 262144-token vocab), badly skewing decode t/s for anyone generating/benchmarking with the flag.

## Fix

Replace it with `FormatTopLogits(ReadOnlySpan<float> logits, int k)` — a single O(V·k) pass over the logits span keeping the k best (descending value, ties broken by lower index), with no `ToArray()` copy and no full sort. Extracted as an `internal static` helper so it's unit-testable.

## Tests

New `VerbosePromptTopKTests` (Tests.Cli) prove **exact equivalence** to the old LINQ output: basic descending order, tie-breaking by lower index, k > length, empty/zero-k, negative values, and a seeded 4096-element pseudo-random vocab compared against the original `OrderByDescending` oracle across k=1/5/8.

Debug-flag-only path; no change to normal decode. Build clean (warnings-as-errors); Tests.Cli 49 green.

https://claude.ai/code/session_01KDR5BLSrhvCk7URYqrp6zA

### DIFF
--- a/src/SharpInference.Cli/RunCommand.cs
+++ b/src/SharpInference.Cli/RunCommand.cs
@@ -1511,10 +1511,7 @@ public sealed class RunCommand : Command<RunCommand.Settings>
             }
             if (verbosePromptLogging)
             {
-                var logitsArr = logits.ToArray();
-                var top5 = Enumerable.Range(0, logitsArr.Length).OrderByDescending(j => logitsArr[j]).Take(5)
-                    .Select(j => $"{j}({logitsArr[j]:F2})");
-                Console.Error.WriteLine($"[DBG] tok={i} next={next}('{tok.Decode([next])}') stop={sp.StopTokenIds.Contains(next)} top5:{string.Join(" ", top5)}");
+                Console.Error.WriteLine($"[DBG] tok={i} next={next}('{tok.Decode([next])}') stop={sp.StopTokenIds.Contains(next)} top5:{FormatTopLogits(logits, 5)}");
             }
             if (sp.StopTokenIds.Contains(next)) break;
             // Counter resets on each <think> open (in case the model opens multiple blocks)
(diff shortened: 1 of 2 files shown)


pekkah:
Addressed review feedback:

**Internal review:** the partial top-k silently dropped `-Infinity` entries (the value-sentinel seed made a real `-Infinity` never outrank the placeholder), diverging from the old LINQ output — which matters because `--verbose-prompt` is exactly what you use when a model emits non-finite garbage. Rewrote with count-tracked insertion + an explicit `SortsBefore` predicate (higher first, NaN last, ties → lower index), so output is now **byte-identical** to `OrderByDescending` for `±Infinity` and `NaN` too. Added oracle tests over the non-finite domain.

**Gemini:** guarded the `stackalloc[k]` with a heap fallback for `k > 256` (matches `Sampler.FindKthLargest`) so a future caller passing a large `k` can't stack-overflow; the decode loop's k=5 stays on the stack.

Tests.Cli 51 green; build clean.

Issue #73, opened 2026-05-29
Vulkan: add MatVecQ5K shader so Q5_K experts can be cached raw (currently dequantized to F32)


pekkah:
## Background

Cached experts should stay in their native quant format. `CudaExpertSlotManager` already does this for Q4_K / Q5_K / Q6_K via `UploadRaw` (`CudaExpertSlotManager.cs:162`). But two other paths **dequantize Q5_K to F32** on upload:

- **Vulkan `ExpertSlotManager`** — `ExpertSlotManager.cs:146` keeps Q4_K/Q6_K raw, but Q5_K falls through to the `Dequantize.ToFloat32` branch (`:156`), so it's cached at **4 bytes/element**.
- **Non-GDN `CudaHybridForwardPass`** resident upload — same pattern: Q4_K/Q6_K raw (`:1418`), Q5_K → F32 (`:1428`).

This matters concretely for Q5_K_M-quantized MoE models (and qwen35moe's `ffn_down_exps`, though that runs on the CUDA-GDN path, not Vulkan): every cached Q5_K expert is 4× its on-disk size, cutting SLRU residency.

## ⚠️ Correction (after source audit)

This is **not the one-line fix** the original issue implied. The Q5_K→F32 dequant on Vulkan is **deliberately correct today**, because:

- `VulkanBackend.MatMul(output, matrix, vector, weightDType)` only has cases for **`Float32`, `Q6_K`, and default→`Q4_K`** (`VulkanBackend.cs:1178`). **There is no Q5_K matvec pipeline.**
- There is a standalone Q5_K *dequant* shader (`Shaders.DequantQ5KM`, `:1935`) but **no fused Q5_K matvec** equivalent to `MatVecQ6K`.
- So if we uploaded Q5_K raw and tagged it `Q5_K` in `_gpuWeightDTypes`, `MatMul` would hit the `default` branch and decode it as **Q4_K** — wrong block layout (144 B vs 176 B per 256-elem block) → garbage output.

## Real scope

This is now a **Vulkan shader work item**, not a slot-manager tweak:

1. **Write a `MatVecQ5K` GLSL compute shader** mirroring `Shaders.MatVecQ6K` (per-block Q5_K dequant-dot; reuse the block-layout logic already in `DequantQ5KM`).
2. Add `_matVecQ5KPipeline` + a `case DType.Q5_K` in `VulkanBackend.MatMul` (`:1178`); dispose it alongside the others (`:1686`).
[...]

PR #317, opened 2026-06-21
feat(vulkan): Q8_0 and Q4_0 weight matvec shaders (#310)

### LINKED ISSUE #310: Vulkan: missing Q8_0 and Q4_0 weight matvec shaders (Q8_0 models can't run on GPU)
## Summary

`VulkanBackend.MatMul` only dispatches weight matvecs for `Float32`, `Q6_K`, and `Q4_K`. There are no shaders for `Q8_0` or `Q4_0` weights, so models quantized in those formats cannot run on the Vulkan GPU path.

## Evidence

`src/SharpInference.Vulkan/VulkanBackend.cs:1197-1205` — `MatMul` switches over only:
- `DType.Float32` → `MatVecF32`
- `DType.Q6_K` → `MatVecQ6K`
- `DType.Q4_K` → `MatVecQ4K` (default)

`Shaders.cs` contains `MatVecF32`, `MatVecQ6K`, `MatVecQ4K` only — no `MatVecQ8_0` / `MatVecQ4_0`.

## What CUDA has

`CudaBackend` matvecs cover `Q4_0`, `Q4_K`, `Q5_K`, `Q6_K`, `Q8_0`, `F32` (dp4a on `Q4_0`/`Q4_K`/`Q8_0`).

## Impact

Concrete user-facing gap: VibeThinker-1.5B ships **Q8_0** as the recommended near-lossless default (see CLAUDE.md), and Q8_0/Q4_0 GGUFs are common. These currently cannot be offloaded to a Vulkan GPU.

## Proposed work

Add `MatVecQ8_0` and `MatVecQ4_0` GLSL shaders and wire them into the `MatMul` dispatch.

## Related

- #73 (add `MatVecQ5K` shader so Q5_K experts can be cached raw) — same class of gap for Q5_K.

pekkah:
Closes #310.

## Problem
`VulkanBackend.MatMul` only dispatched weight matvecs for `Float32`, `Q6_K`, and `Q4_K`. Q8_0 and Q4_0 GGUFs had no shader, so `GpuForwardPass.UploadWeight` silently **dequantized them to F32** on upload. They ran on the Vulkan GPU, but at ~4× the weight VRAM — defeating the point of the quantization. This is a concrete user-facing gap: VibeThinker-1.5B ships **Q8_0** as its recommended near-lossless default (per CLAUDE.md), and Q8_0/Q4_0 GGUFs are common.

## Change
- **New GLSL shaders** `MatVecQ8_0` / `MatVecQ4_0` (`Shaders.cs`), modelled on the existing `MatVecQ6K`: byte-addressed `uint[]` storage buffer, 8 rows × 32 lanes per workgroup, one element per lane per 32-wide block, `subgroupAdd` reduction. The dequant math mirrors the CPU `DequantQ8_0`/`DequantQ4_0` and the CUDA `llm_matvec_q8_0`/`llm_matvec_q4_0` kernels:
  - Q8_0 (34 B/block): `value = d * int8`
  - Q4_0 (18 B/block): `value = (nibble - 8) * d`; low nibble = element `j`, high nibble = element `j+16`
- **Dispatch wiring**: `DType.Q8_0` / `DType.Q4_0` cases in `MatMul`, lazy pipelines, disposed in `DisposeCore`.
- **Keep weights raw on GPU**: new `IsRawGpuQuant()` extends the raw-upload gate to Q8_0/Q4_0 in `UploadWeight`, `UploadExpertWeight`, and `EstimateGpuTensorBytes` (was Q4_K/Q6_K only). The raw byte→float reinterpret now rounds up (`(len+3)/4`) so the non-4-multiple block strides of Q8_0/Q4_0 don't truncate/throw — this also hardens the latent Q6_K (210 B) edge case.

## Verification
- **End-to-end** (`Qwen3-0.6B-Q8_0` on Vulkan, `-g -1 --temp 0`): **byte-identical greedy output** vs the prior F32-dequant path, KV auto-context grew **26155 → 33788** (freed weight VRAM), decode **113 → 151 t/s**.
- **Unit parity tests** (new, synthetic quantized bytes — no GGUF fixture): `MatVecQ8_0MatchesCpu` / `MatVecQ4_0MatchesCpu` dequantize identical bytes via the codebase's own `Dequantize.ToFloat32` (CPU ref) vs the shader (GPU); `matRows=131` (not ÷8) exercises the partial-workgroup guard. **All 18 MatVec tests pass.**
- Reviewed by an internal code-review pass: no findings ≥ Low; the one cosmetic doc nit was fixed.

## Scope / notes
- Embeddings stay on the existing path: a Q8_0/Q4_0 `token_embd` is dequantized to F32 for the embed lookup (a separate quantized-gather path that only supports Q4_K-raw). Orthogonal pre-existing limitation; not regressed.
- CUDA already supported these dtypes; this brings the **Vulkan dense path** to parity. Q5_K remains a gap (tracked separately by #73).

### DIFF
--- a/src/SharpInference.Engine/GpuForwardPass.cs
+++ b/src/SharpInference.Engine/GpuForwardPass.cs
@@ -945,10 +945,11 @@ public sealed unsafe class GpuForwardPass : IForwardPass
             result = _gpu.Upload(floats, TensorShape.D1(floats.Length));
             _weightDTypes[result.Handle] = DType.Float32;
         }
-        else if (info.DType == DType.Q4_K || info.DType == DType.Q6_K)
+        else if (IsRawGpuQuant(info.DType))
         {
-            // Upload raw quantized bytes (reinterpret as floats for storage buffer)
-            int floatCount = data.Length / 4;
+            // Upload raw quantized bytes (reinterpret as floats for storage buffer).
+            // Round up: Q8_0 (34B) / Q4_0 (18B) blocks can make a non-4-multiple total.
+            int floatCount = (data.Length + 3) / 4;
             var rawFloats = new float[floatCount];
             data.CopyTo(MemoryMarshal.AsBytes(rawFloats.AsSpan()));
             result = _gpu.Upload(rawFloats, TensorShape.D1(floatCount));
@@ -1007,9 +1008,9 @@ public sealed unsafe class GpuForwardPass : IForwardPass
         int byteOffset = expertIdx * expertBytes;
         var expertData = data.Slice(byteOffset, expertBytes);
 
-        if (info.DType == DType.Q4_K || info.DType == DType.Q6_K)
+        if (IsRawGpuQuant(info.DType))
         {
-            int floatCount = expertData.Length / 4;
+            int floatCount = (expertData.Length + 3) / 4;
             var rawFloats = new float[floatCount];
             expertData.CopyTo(MemoryMarshal.AsBytes(rawFloats.AsSpan()));
             var result = _gpu.Upload(rawFloats, TensorShape.D1(floatCount));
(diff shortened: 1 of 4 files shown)


pekkah:

pekkah:
Thanks for the review. Re: the `subgroupAdd`/`subgroupElect` subgroup-size-32 assumption (Wave64) — this is a valid concern, but it's a **pre-existing, codebase-wide** convention rather than something introduced here. Every existing Vulkan matvec/reduction shader (`MatVecQ6K`, `MatVecF32`, `MatVecQ4K`, and others) uses the identical `subgroupAdd(acc)` + `if (subgroupElect()) output_data[row] = result;` idiom with 8 rows × 32 lanes per 256-thread workgroup. The new `MatVecQ8_0`/`MatVecQ4_0` shaders deliberately follow that same convention.

Switching only these two shaders to a manual `subgroupShuffleDown` reduction would:
1. Diverge from the rest of the shader suite, and
2. Still not enable Wave64 in practice — the rest of the pipeline (RmsNorm, attention, the other matvecs) shares the same size-32 assumption.

The cleaner fix is a single-point one: pin `requiredSubgroupSize = 32` via `VK_EXT_subgroup_size_control` (already detected as `HasSubgroupSizeControl` but currently unused by `ComputePipeline`), which makes the **entire** existing suite Wave64-correct with zero shader edits. I've filed #318 to track this across all reduction shaders (it also needs validation on real AMD Wave64 hardware, which isn't available in the NVIDIA-only dev environment here).

Correctness on the target hardware is verified: `Qwen3-0.6B-Q8_0` on Vulkan produces byte-identical greedy output vs the prior F32-dequant path, and the new synthetic-bytes parity tests pass.

PR #319, opened 2026-06-21
fix(vulkan,cuda): implement IComputeBackend.SiLU contract (#314)

### LINKED ISSUE #314: Vulkan: VulkanBackend.SiLU throws NotImplementedException (IComputeBackend contract gap)
## Summary

`VulkanBackend.SiLU(Tensor)` is unimplemented and throws, diverging from the `IComputeBackend` contract. Only the fused `SiLuMul` path is available. CPU and CUDA both implement standalone SiLU, so any code path (current or future) that calls `backend.SiLU(...)` will crash on Vulkan instead of working.

## Evidence

- `src/SharpInference.Vulkan/VulkanBackend.cs:1084`
  ```csharp
  public void SiLU(Tensor x) => throw new NotImplementedException("Use SiLuMul for fused SiLU*gate");
  ```
- CUDA implements `SiLU` and `SiLUInPlace` (`CudaBackend.cs`); CPU has `GdnKernels.SiLu`.

## Proposed work

Add a small `SiLU` GLSL shader (or reuse the SiLU half of `SiLuMul`) so `VulkanBackend.SiLU` honors the interface. Low effort; closes a latent crash for any non-fused caller (e.g. a future GDN-on-Vulkan path).

## Severity

Low — no current Vulkan code path calls standalone `SiLU`, but it is an unmet interface contract and a crash waiting for the first caller.

pekkah:
Closes #314.

## Problem
`IComputeBackend` declares `void SiLU(Tensor x)` (in-place SiLU/Swish), and the CPU backend implements it (`GdnKernels.SiLu`). But:
- `VulkanBackend.SiLU` threw `NotImplementedException("Use SiLuMul for fused SiLU*gate")`
- `CudaBackend.SiLU` threw `NotSupportedException("Use SiLuMul(gate, up) for fused SwiGLU on CUDA.")`

So any current or future caller of standalone `backend.SiLU(...)` (e.g. a GDN-on-GPU path that needs the SiLU output independent of a multiplicand) would crash on Vulkan/CUDA instead of working — an unmet interface contract and a latent crash.

## Change
- **Vulkan**: new standalone `SiLU` GLSL shader — in-place `x[i] = x[i] / (1 + exp(-x[i]))`, the unfused counterpart to the existing `SiLuMul`. Wired into `VulkanBackend.SiLU` with a lazy pipeline (single binding, `CountParams`) and disposal, exactly mirroring `SiLuMul`.
- **CUDA**: `SiLU(Tensor)` now delegates to the existing, already-tested `SiLUInPlace(Tensor)` NVRTC kernel (identical formula) instead of throwing.
- **Tests**:
  - `VulkanShaderTests.SiLUMatchesCpu` — N=1031 (not a multiple of 256, exercises the `if (i >= n) return;` bounds guard), compared to the CPU formula within 1e-5.
  - `CudaGdnKernelsTests.SiLU_HonorsInterfaceContract` — calls the `IComputeBackend.SiLU` method directly to guard against re-introducing the throw.

## Verification
All 10 SiLU-related tests pass (Vulkan + CUDA) on a 4070 Ti. Build clean (TreatWarningsAsErrors). Reviewed by an internal code-review pass: no findings ≥ Low on the Vulkan diff; the CUDA delegation (the one suggestion) was incorporated so the contract is now honored uniformly across CPU, Vulkan, and CUDA.

### DIFF
--- a/src/SharpInference.Cuda/CudaBackend.cs
+++ b/src/SharpInference.Cuda/CudaBackend.cs
@@ -3687,8 +3687,10 @@ public sealed unsafe class CudaBackend : IComputeBackend, IImageOpsBackend, IDis
         return result;
     }
 
-    public void SiLU(Tensor x) =>
-        throw new NotSupportedException("Use SiLuMul(gate, up) for fused SwiGLU on CUDA.");
+    // IComputeBackend contract (#314): standalone in-place SiLU. The fused SwiGLU
+    // path still prefers SiLuMul(gate, up); this delegates to the existing NVRTC
+    // in-place kernel so the interface is honored uniformly across backends.
+    public void SiLU(Tensor x) => SiLUInPlace(x);
 
     public void RoPE(Tensor x, int position, int headDim, float ropeTheta = 10000f, bool neox = false)
     {
--- a/src/SharpInference.Vulkan/Shaders.cs
+++ b/src/SharpInference.Vulkan/Shaders.cs
@@ -84,6 +84,29 @@ internal static class Shaders
         """;
 
     /// <summary>
+    /// SiLU (Swish) activation in-place: x[i] = x[i] * sigmoid(x[i]) = x[i] / (1 + exp(-x[i])).
+    /// Push constants: { uint n }.
+    /// Bindings: 0=x (in/out).
+    /// Standalone (unfused) counterpart to <see cref="SiLuMul"/>; matches the CPU
+    /// GdnKernels.SiLu / CUDA SiLUInPlace formula.
+    /// </summary>
+    internal const string SiLU = """
+        #version 450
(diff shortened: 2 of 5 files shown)


pekkah:

pekkah:
Applied the 64-bit dispatch-count arithmetic fix in c3d84fa — `(uint)((x.ElementCount + 255) / 256)`. (The overflow is unreachable for SiLU activation buffers, which are at most embed/intermediate-dim sized, but it's correct hygiene for the new line.) Note this `((uint)ElementCount + 255) / 256` idiom is used throughout VulkanBackend's elementwise dispatches; a codebase-wide sweep would be a separate cleanup.

PR #320, opened 2026-06-21
fix(vulkan): harden + centralize MoE FFN scratch sizing (#315)

### LINKED ISSUE #315: Vulkan: fragile MoE FFN scratch sizing ("garble chased since #2") needs hardening
## Summary

`GpuForwardPass` carries a known-fragile workaround for MoE FFN scratch-buffer sizing. The comment documents a recurring corruption bug ("the MoE-on-Vulkan garble that's been chased since #2") that is currently avoided only by hand-picking the scratch dimension. This is latent-bug territory: the correct size depends on whether the model is MoE, and getting it wrong corrupts expert outputs.

## Evidence

- `src/SharpInference.Engine/GpuForwardPass.cs:225-228`
  ```csharp
  // max(_intermDim, _expertDim) would make an expert MatMul write _intermDim rows
  // when only _expertDim are valid — the MoE-on-Vulkan garble that's been chased
  // since #2 ...
  int ffnScratchDim = _isMoE ? _expertDim : _intermDim;
  ```

## Why file it

- The fix is a manual invariant rather than an enforced one; a future refactor that reintroduces `max(_intermDim, _expertDim)` (or a new expert MatMul that writes `_intermDim` rows) silently re-corrupts output.
- There is no test asserting expert-output correctness against this sizing.

## Proposed work

- Add an assertion/guard tying expert MatMul output rows to `_expertDim`, and a forward-pass test that would catch the garble regression.
- Consider centralizing FFN/expert scratch sizing so the MoE vs dense distinction can't drift.

pekkah:
Closes #315.

## Problem
`GpuForwardPass` sized its FFN scratch buffers (`_ffnGate`/`_ffnUp`) with a fragile manual invariant documented only by a comment — the *"MoE-on-Vulkan garble that's been chased since #2"*:
```csharp
int ffnScratchDim = _isMoE ? _expertDim : _intermDim;
```
`VulkanBackend.MatMul` derives the output row count from `output.ElementCount`, so a future refactor that reintroduces `max(_intermDim, _expertDim)` — or a new expert MatMul writing `_intermDim` rows — would silently corrupt expert output again, with no test to catch it.

## Change
- **Centralize** the sizing in `internal static ComputeFfnScratchDim(isMoE, intermDim, expertDim)` and **enforce** it with `internal static ValidateFfnScratchDim(...)`, called right at the scratch-allocation site. The guard throws `InvalidOperationException` when:
  - a MoE model is mis-sized (`scratch != expertDim`), i.e. the exact `max(intermDim,expertDim)` regression,
  - a MoE model has `expertDim <= 0` (bad GGUF metadata), or
  - a dense model is mis-sized (`scratch != intermDim`).
  Fail loudly instead of garbling. `internal static` keeps it unit-testable via the existing `InternalsVisibleTo`.
- **Tests** (`GpuFfnScratchGuardTests.cs`):
  - 7 GPU-free unit tests pin the invariant in both directions — including the exact regression (`ValidateFfnScratchDim(isMoE:true, scratchDim:8192, intermDim:8192, expertDim:2816)` → throws). This is the deterministic regression detector.
  - A GPU OLMoE forward-pass test confirms the guard does **not** false-positive on a real MoE layout and that MoE inference produces well-formed logits on Vulkan (finite, non-degenerate spread, ≥2 distinct greedy tokens).

## Notes / scope
- Behavior is unchanged for all valid models — `ComputeFfnScratchDim` is byte-identical to the old inline logic; the guard only adds a throw on the corrupting cases.
- OLMoE has `expertDim == intermDim == 1024` (its GGUF lacks `expert_feed_forward_length`), so it can't itself distinguish the regression — the **unit test** covers that direction; the GPU test is the no-false-positive + well-formedness check. (The only local model with `expertDim < intermDim` is Qwen3-Coder-30B, which is too large to fully offload to a 12 GB GPU on the pure Vulkan path, so it isn't used as a GPU fixture.) These facts are documented honestly in the test.
- **Follow-up**: `CudaForwardPass` carries the same `_isMoE ? _expertDim : _intermDim` pattern with no guard; it can adopt these backend-agnostic statics. Also, the shared-expert FFN reuses the `_expertDim`-sized scratch (fine for current models, where `expert_shared_feed_forward_length` is 0) — worth a guard if a model with a divergent shared-expert dim is ever added.

Verified on a 4070 Ti: build clean (TreatWarningsAsErrors), 8/8 new tests pass (the GPU OLMoE test runs real inference). Reviewed by an internal code-review pass; the one substantive finding (misleading OLMoE docstrings) was corrected.

### DIFF
--- a/src/SharpInference.Engine/GpuForwardPass.cs
+++ b/src/SharpInference.Engine/GpuForwardPass.cs
@@ -225,7 +225,10 @@ public sealed unsafe class GpuForwardPass : IForwardPass
         // max(_intermDim, _expertDim) would make an expert MatMul write _intermDim rows
         // when only _expertDim are valid — the MoE-on-Vulkan garble that's been chased
         // since #2. Pure-MoE and pure-dense models both fall out correctly from this.
-        int ffnScratchDim = _isMoE ? _expertDim : _intermDim;
+        // Sizing is centralized in ComputeFfnScratchDim and enforced by
+        // ValidateFfnScratchDim so the MoE-vs-dense distinction can't drift (issue #315).
+        int ffnScratchDim = ComputeFfnScratchDim(_isMoE, _intermDim, _expertDim);
+        ValidateFfnScratchDim(_isMoE, ffnScratchDim, _intermDim, _expertDim);
         _ffnGate = gpu.Allocate(TensorShape.D1(ffnScratchDim));
         _ffnUp = gpu.Allocate(TensorShape.D1(ffnScratchDim));
         _logits = gpu.Allocate(TensorShape.D1(hp.VocabSize));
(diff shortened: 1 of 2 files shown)


pekkah:

pekkah:
Thanks. On the backslash path separator: this matches the established `FindModelPath` convention used across the test project (`VulkanShaderTests`, `CudaHybridForwardPassCpuMoeTests`, etc. all use `models\…`). These are Windows-local model fixtures — on Linux/macOS the files aren't present and the test skips regardless, so the separator doesn't change behavior there. Keeping it consistent with the existing helpers rather than diverging in one file; a cross-platform sweep of all of them could be a separate cleanup if the suite ever runs with fixtures on non-Windows CI.

PR #322, opened 2026-06-21
feat(vulkan): Q5_K weight matvec shader (#73)

### LINKED ISSUE #73: Vulkan: add MatVecQ5K shader so Q5_K experts can be cached raw (currently dequantized to F32)

pekkah:
Closes #73.

## Problem
`VulkanBackend.MatMul` had no `Q5_K` case, so `GpuForwardPass`/`HybridForwardPass` dequantized Q5_K weights to F32 on upload (~5.8× the VRAM). A 4B Q5_K model (e.g. the Z-Image Qwen3-4B text encoder, `Z-Image-AbliteratedV1.Q5_K_M.gguf`) **OOM'd on a 12 GB GPU** on Vulkan — it couldn't even fit as a 29/36-layer hybrid split. CUDA already keeps Q5_K raw (`_matvecQ5KKernel`, `EmbedLookupQ5K`); this brings the Vulkan path to parity. Follow-on to #310 (Q8_0/Q4_0).

## Change
- **New `MatVecQ5K` GLSL shader** (`Shaders.cs`): reuses `MatVecQ4K`'s proven 6-bit scale/min unpack and adds the Q5_K `qh` high-bit (`+16`) per quant. 8 rows × 32 lanes, `gByte` byte-gather, 176-byte/256-element blocks. Mirrors CUDA `llm_matvec_q5k` and CPU `DequantQ5K`.
- **Dispatch**: `DType.Q5_K` case in `MatMul` (lazy pipeline + disposal).
- **Keep Q5_K raw on upload**: `GpuForwardPass.IsRawGpuQuant`, `HybridForwardPass` (3 gates), and `TierPlanner.EstimateGpuTensorBytes` (weights — correct for *both* backends; also fixes the stale "Q5_K dequants to F32" comment).
- **`TierPlanner.MeasureGpuEmbeddingBytes` deliberately keeps Q5_K on the F32 estimate.** TierPlanner is backend-agnostic and the Vulkan path has **no** `EmbedLookupQ5K` (a Q5_K `token_embd` dequantizes to F32), so counting it raw would under-budget VRAM on Vulkan (OOM). Conservative for CUDA, safe for Vulkan. (A future Vulkan `EmbedLookupQ5K` could let embeddings go raw too — out of scope; #73 targets the dominant matrix-weight cost.)

## Verification
- **End-to-end (4070 Ti)**: the Z-Image Q5_K 4B model that **OOM'd on master** now loads **all 36 layers on the Vulkan GPU** and decodes coherently (~56 t/s). Before: `ErrorOutOfDeviceMemory` (F32-dequantized weights ~16 GB > 12 GB).
- **Parity test**: `MatVecQ5KMatchesCpu` (synthetic 176-byte blocks, `matRows=131` partial-workgroup, `matCols=512` = 2 blocks) passes bit-clean vs CPU `Dequantize.ToFloat32`; all 19 MatVec tests pass. Build clean (TreatWarningsAsErrors).
- Internal code-review pass: no findings at any severity; shader bit-math independently verified against CPU `DequantQ5K` and CUDA `llm_matvec_q5k`, and the embedding-gate exclusion confirmed correct.

## Related
- #310 (Q8_0/Q4_0 matvec) — same pattern. Q5_K was the remaining common K-quant gap on Vulkan.

### DIFF
--- a/src/SharpInference.Engine/GpuForwardPass.cs
+++ b/src/SharpInference.Engine/GpuForwardPass.cs
@@ -1275,6 +1275,6 @@ public sealed unsafe class GpuForwardPass : IForwardPass
     /// matching <c>VulkanBackend.MatMul</c> matvec dispatch) rather than expanded to F32 on
     /// the CPU. Keeping these quantized is the whole point of the GPU matvec shaders.
     /// </summary>
-    private static bool IsRawGpuQuant(DType dtype) =>
+    internal static bool IsRawGpuQuant(DType dtype) =>
         dtype is DType.Q4_K or DType.Q5_K or DType.Q6_K or DType.Q8_0 or DType.Q4_0;
 }
--- a/src/SharpInference.Engine/GpuForwardPass.cs
+++ b/src/SharpInference.Engine/GpuForwardPass.cs
@@ -1276,5 +1276,5 @@ public sealed unsafe class GpuForwardPass : IForwardPass
     /// the CPU. Keeping these quantized is the whole point of the GPU matvec shaders.
     /// </summary>
     private static bool IsRawGpuQuant(DType dtype) =>
-        dtype is DType.Q4_K or DType.Q6_K or DType.Q8_0 or DType.Q4_0;
+        dtype is DType.Q4_K or DType.Q5_K or DType.Q6_K or DType.Q8_0 or DType.Q4_0;
 }
--- a/src/SharpInference.Engine/HybridForwardPass.cs
+++ b/src/SharpInference.Engine/HybridForwardPass.cs
@@ -1267,7 +1267,7 @@ public sealed unsafe class HybridForwardPass : IForwardPass
 
     private static long EstimateGpuTensorBytes(GgufTensorInfo tensor)
     {
-        if (tensor.DType == DType.Float32 || tensor.DType == DType.Q4_K || tensor.DType == DType.Q6_K)
+        if (tensor.DType == DType.Float32 || tensor.DType == DType.Q4_K || tensor.DType == DType.Q5_K || tensor.DType == DType.Q6_K)
             return (tensor.ByteSize + 3) & ~3L;
 
         return tensor.ElementCount * sizeof(float);
@@ -1328,7 +1328,7 @@ public sealed unsafe class HybridForwardPass : IForwardPass
             result = _gpu.Upload(floats, TensorShape.D1(floats.Length));
             _gpuWeightDTypes[result.Handle] = DType.Float32;
         }
-        else if (info.DType == DType.Q4_K || info.DType == DType.Q6_K)
+        else if (info.DType == DType.Q4_K || info.DType == DType.Q5_K || info.DType == DType.Q6_K)
         {
             int floatCount = data.Length / 4;
             var rawFloats = new float[floatCount];
@@ -1584,7 +1584,7 @@ public sealed unsafe class HybridForwardPass : IForwardPass
         int byteOffset = expertIdx * expertBytes;
         var expertData = data.Slice(byteOffset, expertBytes);
 
-        if (info.DType == DType.Q4_K || info.DType == DType.Q6_K)
+        if (info.DType == DType.Q4_K || info.DType == DType.Q5_K || info.DType == DType.Q6_K)
         {
             int floatCount = expertData.Length / 4;
             var rawFloats = new float[floatCount];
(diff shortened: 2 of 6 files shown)


pekkah:

pekkah:
Applied all review findings in the latest commit:
- **HybridForwardPass (high)**: all three gates now route through the now-`internal` `GpuForwardPass.IsRawGpuQuant` (Q4_K/Q5_K/Q6_K/Q8_0/Q4_0) — single source of truth shared by the dense pass, hybrid pass, and TierPlanner, so they can't drift. This also closes the pre-existing #310 gap where the hybrid path dequantized Q8_0/Q4_0 to F32 while TierPlanner budgeted them packed (under-budget OOM risk).
- **floatCount truncation (high)**: `data.Length / 4` → `(data.Length + 3) / 4` in both hybrid upload paths.
- **MatVecQ5K aligned reads (medium)**: d/dmin + the 12 scale bytes now read as four aligned uint words (block start is 176-aligned), matching MatVecQ4K; qh/ql stay byte-gathered. Dequant unchanged; parity test still bit-clean.

Re-verified e2e: the Q5_K 4B model still loads all 36 layers on Vulkan and decodes coherently.

Issue #312, opened 2026-06-21
Vulkan: attention limited by 4096-slot shared-memory fast path — no flash-attention / split-KV


pekkah:
## Summary

Vulkan attention uses a shared-memory fast path capped at sequence length 4096; beyond that it spills per-position softmax scores to a VRAM scratch buffer (`numHeads × maxSeqLen` floats), which is markedly slower. There is no flash-attention (online-softmax, tiled) kernel and no split-KV / flash-decoding for long contexts, both of which CUDA has.

## Evidence

- `VulkanBackend.cs:1242` — VRAM scratch is used "when `seqLen > 4096`".
- `VulkanBackend.cs:1269,1379` — `promptLen > 4096` / `tqSeqLen + fp16SeqLen > 4096` exceed the shared-memory fast-path cap.
- `GpuForwardPass.cs:289-293` — allocates `numHeads × maxSeqLen` softmax spill scratch once `maxSeqLen > 4096`.

## What CUDA has

- `FlashAttentionPrefill` + tensor-core variants `FlashAttentionPrefillTc` / `FlashAttentionPrefillTc2` (issue #147).
- `AttentionSplitKv` flash-decoding split-KV for long contexts (issues #235/#237/#238), gated by `SHARPI_SPLIT_DECODE`.

## Proposed work

- Implement an online-softmax (flash-attention) attention shader on Vulkan that does not require an O(seqLen) shared-memory or VRAM score buffer.
- Add split-KV decode for long-context throughput.

## Impact

Long-context prefill/decode on Vulkan degrades sharply past 4096 tokens and wastes VRAM on the score spill buffer.

pekkah:
### Attempted (single-query online-softmax) — measured a ~3× regression, NOT shipping that design

I implemented the natural flash-decode design — one workgroup per head, **thread `d` owns output dim `d`**, with a per-KV-position cooperative Q·K reduction (shared-memory tree) + online-softmax recurrence, deleting the `scores[]`/`scores_scratch[]` buffer entirely. It's **correct** (passed the `AttentionShader_*` parity tests at `<1e-3`, including a new >4096 case, plus bf16/q8_0 KV parity), but it's a **large performance regression** because of the per-position `barrier()` cost (~log(head_dim) barriers per KV position):

Qwen3-8B, `--backend vulkan -g -1 --kv-type fp32`, ~4.4K-token context:
| | Prefill | Decode |
|---|---|---|
[...]

PR #323, opened 2026-06-21
feat(vulkan): bf16 (fp16-packed) KV cache (#311)

### LINKED ISSUE #311: Vulkan: KV cache is FP32-only — no BF16 or Q8_0 KV cache (2× memory vs CUDA)
## Summary

The Vulkan KV cache stores keys/values in FP32. There is no half-width (BF16) or Q8_0 KV cache option, so Vulkan uses roughly 2× the KV memory of CUDA for the same context, capping achievable context length on a given VRAM budget. (TurboQuant 3-bit exists, but it is a separate, more aggressive path with constraints — see below — not a drop-in half-width cache.)

## Evidence

- `Shaders.cs` provides `KvAppend` (FP32) and `TqKvAppend` (TurboQuant 3-bit) only — no `KvAppendBf16` / `KvAppendQ8_0` and no matching attention variants.
- TurboQuant on Vulkan is gated: `GpuForwardPass.cs:178` restricts it to `head_dim ∈ {128, 256}`, and `GpuForwardPass.cs:188` blocks SnapKV+TQ composition (issue #60).

## What CUDA has

- BF16 KV cache (`KvAppendBf16` / `KvAppendBatchedBf16`, issue #27).
- Q8_0 KV cache at ~1.06 B/elem, ~half of BF16 (`KvAppendQ8_0` / `KvAppendBatchedQ8_0`, issue #179), with matching `AttentionQ8_0` and `SnapKvScoreBf16` paths.
- `SHARPI_KV_DTYPE` to select fp32/bf16.

## Proposed work

Add `KvAppendBf16` (and optionally `KvAppendQ8_0`) shaders plus matching attention kernels on Vulkan, and a KV-dtype selection knob mirroring `SHARPI_KV_DTYPE`.

## Impact

Lower max context / higher memory pressure on non-NVIDIA GPUs vs CUDA for identical models.

pekkah:
Closes #311.

## Problem
The Vulkan KV cache was **FP32-only** (the CUDA path already has bf16/q8_0 KV). At fp32, KV dominates VRAM at long context — e.g. Qwen3-8B on a 12 GB GPU auto-caps context at ~11.4K and auto-enables SnapKV to fit.

## Change
Add a half-width KV option selected by `--kv-type bf16`. **Storage is IEEE fp16 packed two-per-`uint`** via core-GLSL `packHalf2x16`/`unpackHalf2x16` (no device extension); attention arithmetic stays fp32 — only the stored K/V mantissa is narrowed. The user-facing flag stays `bf16` (= "half-width KV"); fp16 is in fact *more* precise than bf16 for the small-magnitude KV values, and needs no extension gating.

- **New `KvAppendBf16` / `AttentionBf16` GLSL shaders** — structurally identical to the fp32 `KvAppend`/`Attention`, indexing the cache the same way (**no ring modulo**); only the K/V store/load is fp16-packed. `KvAppendBf16` runs one thread per 2 elements (`kv_dim` is always even); `AttentionBf16` reads `unpackHalf2x16(cache[(off+d)>>1])[(off+d)&1]` and keeps scores/softmax/accumulation + the `1/sqrt(headDim)` scale in fp32.
- **`VulkanBackend.KvAppendBf16`/`AttentionBf16`** (+ lazy pipelines, disposal), reusing the existing push-constant structs.
- **`GpuForwardPass`**: new `kvDtype` ctor arg (default `Float32` → byte-identical fp32 path); bf16 cache allocated as `DType.BFloat16`; per-layer dispatch branch; `EstimateMaxContext` halves the KV byte estimate under bf16 (buys ~2× auto-context); `[KV bf16]` banner. **Guards**: bf16 rejects q8_0, TurboQuant, and explicit SnapKV, and forces auto-SnapKV off (the `SnapKvScore`/`KvCompact` shaders read fp32 KV); plus an even-KV-row-width guard so a future odd-`head_dim` model fails loudly instead of corrupting the packed cache.
- **Plumbing**: `--kv-type` now reaches the Vulkan ctor on **both** the CLI and the server (`InferenceEngineLoader`).

## Verification (4070 Ti)
- **2× context, identical output**: Qwen3-8B `--backend vulkan -g -1` — fp32 caps at `Context size: 11399` (SnapKV auto-on, ~3206 MiB cache); bf16 reaches **22798** `[KV bf16]` (SnapKV correctly auto-off) with **byte-identical greedy output** and unchanged decode t/s.
- **Long-context (>4096) scratch-spill attention path**: a **5149-token prefill** with bf16 decodes coherently with no NaN — exercising the `AttentionBf16` `seq_len > 4096` branch for thousands of positions.
- **Tests**: new `GpuForwardPassKvDtypeTests` — teacher-forced bf16-vs-fp32 argmax parity + greedy coherence. fp32 path is byte-identical (default ctor arg); existing Vulkan SnapKV/scratch-guard tests still pass.
- Internal code-review pass: the one High finding (server path silently ignored the flag) and a kv_dim-even guard were applied; the long-context path it flagged was then validated by the 5149-token e2e run above.

## Scope / follow-ups
- **bf16 only**; Q8_0 KV on Vulkan is a natural follow-up (needs per-block quant on append + dequant in attention).
- bf16 + SnapKV on Vulkan is guarded off (would need fp16 variants of the SnapKV score/compact shaders) — follow-up.

### DIFF
--- a/src/SharpInference.Cli/RunCommand.cs
+++ b/src/SharpInference.Cli/RunCommand.cs
@@ -739,9 +739,13 @@ public sealed class RunCommand : Command<RunCommand.Settings>
 
                 if (nGpuLayers >= hp.NumLayers)
                 {
-                    // All layers on GPU
+                    // All layers on GPU. Pass the configured KV dtype (issue #311): fp32 default,
+                    // bf16 = half-width KV. The Vulkan ctor rejects q8_0 with a clear message
+                    // (q8_0 KV is CUDA-only for now). Reuses the same --kv-type/SHARPI_KV_DTYPE
+                    // parser the CUDA path uses.
                     var gfwd = new GpuForwardPass(model, gpu, hp, ctxSize,
-                        enableTurboQuant: settings.TurboQuant);
+                        enableTurboQuant: settings.TurboQuant,
+                        kvDtype: CudaForwardPass.ResolveConfiguredKvDType());
                     if (settings.TurboQuant)
                         AnsiConsole.MarkupLine($"[dim]TurboQuant: [green]enabled[/] (3-bit, context: {gfwd.MaxSeqLen})[/]");
                     gpuFwd = gfwd;
--- a/src/SharpInference.Engine/GpuForwardPass.cs
+++ b/src/SharpInference.Engine/GpuForwardPass.cs
@@ -128,8 +128,16 @@ public sealed unsafe class GpuForwardPass : IForwardPass
     private bool _snapKvScoreScratchOwned; // false if aliased to _attnScoresScratch
     private int _snapKvCaptureSlot = -1; // 0..W-1 for tokens in the capture window; -1 otherwise
 
+    // KV-cache store dtype (issue #311). Float32 (default) keeps the original fp32 cache;
+    // BFloat16 selects the half-width KV path. On Vulkan "bf16" means "half-width KV" but the
+    // value is stored as IEEE fp16 packed two-per-uint (more precise than bf16 for the small KV
+    // magnitudes, and packHalf2x16 is core GLSL so no device extension is needed). Arithmetic
+    // stays fp32 — only the stored value is narrowed. q8_0 is CUDA-only for now (rejected below).
+    private readonly DType _kvDType;
+
     public GpuForwardPass(GgufModel model, VulkanBackend gpu, ModelHyperparams hp,
-        int maxContextLength = 0, bool enableTurboQuant = false, int tqFp32Window = 256, int tqBits = 3)
+        int maxContextLength = 0, bool enableTurboQuant = false, int tqFp32Window = 256, int tqBits = 3,
+        DType kvDtype = DType.Float32)
     {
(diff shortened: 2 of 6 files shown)


pekkah:

pekkah:
Thanks — all three are good perf suggestions. One correction on the alignment finding: the **current** `AttentionBf16` reads by absolute index (`unpackHalf2x16(cache[(off+d)>>1])[(off+d)&1]`), which is correct for any per-head offset within an even-`kv_dim` row, so there's no data-corruption risk today and the `kv_dim`-even ctor guard is the precise/sufficient invariant. The `head_dim`-even check is specifically a **prerequisite for the two-at-a-time optimization** (Phase 1 `k_off>>1` word-stepping), not for the current code.

Since the two shader optimizations + the tightened guard form one coherent change (and overlap with the #312 attention rework), and this PR is already verified end-to-end (bf16-vs-fp32 parity + a 5149-token long-context run through the scratch-spill path), I've filed **#324** to do the read-each-word-once optimization (with the `head_dim`-even guard) rather than rewrite the attention hot path in this already-large PR. Merging the correct, verified version now.

Issue #325, opened 2026-06-21
Vulkan: Q8_0 KV cache (4× vs fp32) — #311 follow-up (bf16 shipped, q8_0 remains)


pekkah:
## Summary

#311 added a **bf16** (fp16-packed) KV cache to the Vulkan path (PR #323), but its original scope was "no BF16 **or Q8_0** KV cache". Q8_0 KV remains: it stores K/V as 8-bit (~4× smaller than fp32, ~2× smaller than bf16), letting even longer contexts fit on the GPU. CUDA already has this.

## CUDA reference to mirror
- `CudaBackend.KvAppendQ8_0` / `AttentionQ8_0` (`src/SharpInference.Cuda/CudaBackend.cs` ~5052/5085).
- `llm_kv_append_q8_0` (`src/SharpInference.Cuda/CudaTextKernels.cs` ~827) — the store kernel quantizes **per 32-lane warp** (one block per 32 elements); read-side recovery is `sharpi_kvload(const block_q8_0*, ...)`. The same templated attention bodies serve fp32/bf16/q8_0 via the load overload. Relies on **kv_dim being a multiple of 32**.

## Proposed work
- Vulkan `KvAppendQ8_0` shader: per 32-element block of the KV row, `subgroupMax(|x|)` → `d = amax/127` (fp16), quantize 32 int8, store the block (fp16 d + 32 int8). Requires `kv_dim % 32 == 0` (guard).
- Vulkan `AttentionQ8_0` shader: dequantize K/V blocks (`int8 * d`) in the Q·K dot and V accumulate; arithmetic stays fp32. Mirror the existing fp32/bf16 `Attention` structure (shared-mem fast path + scratch spill).
- Reuse the #311 plumbing: extend the `GpuForwardPass` `kvDtype` path + dispatch + `DType.Q8_0` allocation (34 or 36 bytes/32 elems — confirm block_q8_0 vs block_q8_1 from the CUDA layout) + `EstimateMaxContext` (4× smaller). Drop the current "q8_0 is CUDA-only" rejection in the Vulkan ctor.
- Guards: q8_0 requires `kv_dim % 32 == 0`; mutually exclusive with TurboQuant and SnapKV (same as bf16).

## Verification
[...]

PR #327, opened 2026-06-21
perf(vulkan): optimize narrowed-KV attention reads (#324)

### LINKED ISSUE #324: perf(vulkan): optimize narrowed-KV attention reads (bf16 two-at-a-time + q8_0 single-word scale/bitfieldExtract) — #311/#325 follow-up
## Summary

`AttentionBf16` (added in #311 / PR #323) reads the fp16-packed KV cache by absolute element index:
```glsl
unpackHalf2x16(k_cache[(k_off + d) >> 1])[(k_off + d) & 1u]
```
This is correct for any offset (it needs only `kv_dim` even, which the ctor guards), but for consecutive `d` it loads the **same `uint` word twice** (once for the low half, once for the high half), unpacking both each time and discarding one. Attention KV reads are the bandwidth-bound hot path — exactly what bf16 KV exists to stretch (long context).

## Proposed work (from a Gemini review of PR #323)

1. **Phase 1 (Q·K dot)** — process elements two-at-a-time so each packed word is read once and both halves used:
   ```glsl
   uint k_off_half = k_off >> 1;
   for (uint d_half = 0; d_half < (head_dim >> 1); d_half++) {
       vec2 kv = unpackHalf2x16(k_cache[k_off_half + d_half]);
       uint d = d_half << 1;
       dot += q_data[q_off + d] * kv.x + q_data[q_off + d + 1u] * kv.y;
   }
   ```
   Halves the K global loads + unpacks in the dot product. Apply to both the shared-memory and scratch-spill branches.
2. **Phase 3 (V accumulate)** — loop-invariant code motion / strength reduction: precompute `d_half`/`component`, init `v_off_half` once, increment by `kv_dim>>1` per `t` (removes the per-iteration multiply). (Each thread owns one `d`, so this doesn't halve V loads, just removes arithmetic.)

## Prerequisite

These optimizations assume each **head** starts on a word boundary, i.e. `head_dim` itself even (so `kv_head * head_dim` is always even), not just `kv_dim` even. Tighten the `GpuForwardPass` bf16 guard from `(numKvHeads*headDim) even` to `headDim even` when adopting this. (All currently-supported head dims — 64/128/256 — are even, so this rejects nothing real.)

## Notes
- Pure perf; the current shader is correct. Verify with the existing `Bf16Kv_ArgmaxStable_VsFp32` parity test (shared path) **plus** a long-context (>4096 tokens) e2e run to cover the scratch-spill branch.
- Overlaps with #312 (Vulkan attention rework / flash-attention) — could be folded in there.

pekkah:
Closes #324. Follow-up to #311 (bf16 KV) and #325 (q8_0 KV).

## Change
Pure-efficiency, **output-identical** read optimizations to the two narrowed-KV Vulkan attention shaders. No math/output change — verified bit-identical by the existing `Bf16Kv`/`Q8Kv` parity tests (27 pass, no bound change).

- **`AttentionBf16` Q·K dot** — read each fp16-packed `uint` word **once** (two K elements at a time) instead of loading the same word twice for consecutive `d`. Halves K-cache word loads in the attention hot path. Valid because `head_dim` is even (guard tightened below), so `k_off>>1` is the exact word base.
- **`AttentionBf16` V accumulate** — loop-invariant strength reduction (hoist `d_half`/`component`, walk `v_off_half += kv_dim>>1`); no load-halving possible since each thread owns one `d`.
- **`AttentionQ8_0` `loadK`/`loadV`** — load the fp16 scale word **once** (`b0` and `b0+1` always share a `uint` word since `b0=34·blk` is even) and sign-extend the int8 with `bitfieldExtract` (single BFE, no ternary). Drops the now-unused `gByteK`/`gByteV` helpers.
- **Guard** — tighten the bf16 KV guard from even-kv-row-width to `head_dim`-even (the two-at-a-time read needs each head word-aligned). Rejects nothing real (head dims 64/128/256 are all even).

Strictly fewer loads/instructions with identical output — **cannot regress**.

## Verification (4070 Ti)
- **Correctness**: all **27** `Bf16Kv`/`Q8Kv` parity + coherence tests pass (bit-identical reads; no tolerance change).
- **Perf** (Qwen3-8B, `--backend vulkan -g -1 --kv-type bf16`, ~4.5K context, decode 100 tokens): **26.2 → 28.7 t/s = +9.5%** decode. Above noise, and the win grows with context (attention K-load is a larger fraction at the 22K–40K contexts bf16/q8_0 KV enables).

### DIFF
--- a/src/SharpInference.Engine/GpuForwardPass.cs
+++ b/src/SharpInference.Engine/GpuForwardPass.cs
@@ -161,13 +161,15 @@ public sealed unsafe class GpuForwardPass : IForwardPass
         if (kvNarrowed && _kvDType is not (DType.BFloat16 or DType.Q8_0))
             throw new NotSupportedException(
                 "Vulkan KV cache supports fp32, bf16, and q8_0 only (issue #325).");
-        // The bf16 path packs two fp16 per uint, so a KV row (numKvHeads*headDim) must be
-        // even for rows to stay word-aligned. True for all supported head dims (64/128/256),
-        // but enforce it so a future odd-head_dim model fails loudly instead of corrupting KV.
-        if (_kvDType == DType.BFloat16 && ((hp.NumKvHeads * hp.HeadDim) & 1) != 0)
+        // The bf16 path packs two fp16 per uint, so each KV head must START on a uint-word
+        // boundary for the AttentionBf16 two-at-a-time reads (issue #324) to address words
+        // correctly: a head begins at kv_head*head_dim, which is word-even iff head_dim itself
+        // is even. True for all supported head dims (64/128/256), but enforce it so a future
+        // odd-head_dim model fails loudly instead of corrupting KV reads.
+        if (_kvDType == DType.BFloat16 && (hp.HeadDim & 1) != 0)
             throw new NotSupportedException(
-                $"bf16 KV requires an even KV-row width (numKvHeads*headDim); got " +
-                $"{hp.NumKvHeads}*{hp.HeadDim}. Use fp32 KV for this model (issue #311).");
+                $"bf16 KV requires an even head dimension; got {hp.HeadDim}. " +
+                "Use fp32 KV (issue #324).");
         // The q8_0 path quantizes per 32-element block (one thread per block), so a KV row must
         // be a multiple of 32 for blocks to align to row boundaries (no straddling).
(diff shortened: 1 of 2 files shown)


pekkah:

PR #328, opened 2026-06-21
feat(vulkan): precompile shaders to committed SPIR-V for NativeAOT (#316)

### LINKED ISSUE #316: Vulkan: runtime glslc (Vulkan SDK) dependency breaks NativeAOT single-binary deployment
## Summary

Vulkan shaders are compiled from embedded GLSL to SPIR-V **at runtime** by shelling out to `glslc`, which requires the Vulkan SDK to be installed and on PATH. This conflicts with the project's NativeAOT single-binary deployment goal — a published binary will throw `FileNotFoundException` on any machine without the Vulkan SDK, even though the GPU/driver supports Vulkan.

## Evidence

- `src/SharpInference.Vulkan/ShaderCompiler.cs` — locates `glslc` and otherwise throws `FileNotFoundException("glslc not found. Install the Vulkan SDK.")`.
- All 40 shaders in `Shaders.cs` are GLSL strings compiled on demand via `glslc`; there is no pre-compiled SPIR-V fallback.

## Contrast

CUDA also compiles kernels at runtime (NVRTC), but NVRTC ships inside the CUDA runtime that a CUDA app already requires — there is no extra developer-SDK install for end users. The Vulkan path requires a separate SDK that end users typically do not have.

## Proposed work

- Pre-compile the GLSL shaders to SPIR-V at build time (e.g. MSBuild step invoking `glslc`/`shaderc`) and embed the resulting SPIR-V as resources.
- Load embedded SPIR-V at runtime; keep `glslc` strictly as a dev-time/build-time tool.

## Impact

Today, shipping a self-contained Vulkan-enabled build is not possible without instructing end users to install the Vulkan SDK.

pekkah:
Closes #316.

## Problem
`ShaderCompiler.Compile` shelled out to `glslc.exe` at **runtime**, so the NativeAOT single-binary (a stated project goal) depended on the Vulkan SDK being installed on the *target* machine. 48 shaders, all compiled on first use.

## Change
Precompile every shipped shader to SPIR-V at **dev time** and commit the table; the published binary serves shaders without glslc.

- **`ShaderCompiler`**: new deterministic **FNV-1a 64-bit `StableHash`** over UTF-8 (`String.GetHashCode` is per-process randomized → unusable for a build-time table). `Compile()` returns precompiled SPIR-V from the committed `ShadersPrecompiled` table on a stable-hash hit; the in-process cache is rekeyed to the stable hash; glslc is invoked **only on a miss** (ad-hoc/test shaders), bumping `GlslcFallbackCount`.
- **`Shaders.Precompiled.g.cs`** (committed, generated): trim/AOT-safe `static readonly byte[]` table + `switch`-based `TryGet`. **46/48** shaders precompiled. `SgemmBf16`/`SgemmFp8` are skipped — glslc 1.4.341 can't compile their bf16/fp8 extensions; `VulkanBackend.Sgemm`'s `catch (Exception)` already degrades those to fp16/fp32 (the `FileNotFoundException` on a no-glslc binary is caught the same as a driver lacking the feature → **no crash, no regression**).
- **`tools/SpirvGen`** (dev-only; not in the solution/pack, referenced by no shipping project): reflects the `Shaders` consts, calls `ShaderCompiler.Compile` (glslc) for each, emits the committed table. Skips **only** the known unsupported-extension case; **hard-fails on any other glslc error** so a real shader can't be silently dropped. `scripts/gen-spirv.ps1` regenerates; CLAUDE.md documents the step.
- Committing the table keeps the **glslc-less `dotnet pack` CI** (ubuntu) building unchanged — glslc is needed only to *regenerate*, never to *build*.

## Verification
- `VulkanPrecompiledShaderTests`: **(a)** `AllShaderConstantsArePrecompiled` recomputes the runtime stable hash for each const and asserts a table hit + SPIR-V magic `0x07230203` — the linchpin that catches drift/hash-mismatch; **(b)** count reconciliation (`Count + Skipped == 48`); **(c)** GPU `GlslcFallbackCount == 0` after a real `RmsNorm` (proves the precompiled path served the shader). All pass.
- Full `VulkanShaderTests` (24) pass via the precompiled path; full solution + meta-package build 0 warnings (trim/AOT analyzers + TreatWarningsAsErrors).
- Internal code-review pass: clean (hash determinism + collision-checked at generation, AOT/CI-safe table, graceful skipped-shader fallback verified end-to-end).

## Scope / notes
- Full `dotnet publish -r win-x64 /p:PublishAot=true` end-to-end confirmation is **deferred** (slow); the mechanism is proven by the completeness + no-fallback tests without it.
- The committed `.g.cs` is large (~0.7 MB of SPIR-V byte data) — that's the cost of an in-binary, glslc-free shader table; could move to an EmbeddedResource blob later if repo size matters.
- Drift is guarded: editing a shader without regenerating fails the completeness test with a message pointing at `scripts/gen-spirv.ps1`.


pekkah:

pekkah:
Applied all review findings in the latest commit:
- **High (temp-file leak)**: the glslc fallback now uses `Path.GetRandomFileName()` instead of `Path.GetTempFileName() + ext` (the latter leaks an empty `.tmp` per call — hit 48× per SpirvGen run). Fixed.
- **Medium (BypassPrecompiled)**: added `ShaderCompiler.BypassPrecompiled`; SpirvGen sets it at startup so regeneration always recompiles fresh from GLSL — correctly picking up glslc-flag changes (not just source edits, which already changed the hash). Good catch.
- **Medium×2 (hex lookup)**: SpirvGen now precomputes the 256 byte→hex strings once.

Re-verified: regenerated table is **byte-identical** (idempotent, confirming the fresh BypassPrecompiled compile matches), precompiled completeness/count/no-fallback + functional shader tests all pass.

PR #329, opened 2026-06-21
fix(vulkan): pin requiredSubgroupSize=32 for reduction shaders (#318)

### LINKED ISSUE #318: Vulkan: reduction shaders assume subgroup size 32 — pin requiredSubgroupSize=32 (or manual shuffle reduce) for Wave64 safety
## Summary

Every Vulkan compute shader that does a per-row reduction uses `subgroupAdd(acc)` followed by `if (subgroupElect()) output_data[row] = result;`, assuming the subgroup size is exactly 32 (== `THREADS_PER_ROW`, with 8 rows packed per 256-thread workgroup).

On a GPU with subgroup size 64 (e.g. AMD running Wave64), this is incorrect:
- `subgroupAdd` sums across all 64 lanes → combines two adjacent rows.
- `subgroupElect()` is true only for the first lane of the 64-wide subgroup → the second row in each subgroup never writes its output, and the first row's sum is wrong.

This is a **pre-existing, codebase-wide** assumption (the Vulkan path is developed and tested on NVIDIA, where the subgroup/warp size is always 32). It is not specific to any one shader.

## Affected shaders (`src/SharpInference.Vulkan/Shaders.cs`)

All matvec + reduction kernels using the `subgroupAdd` + `subgroupElect` idiom, including:
- `MatVecQ6K`, `MatVecF32`, `MatVecQ4K`
- `MatVecQ8_0`, `MatVecQ4_0` (added in #310 / PR #317, following the same established convention)
- any other reduction shaders sharing the pattern (grep: `subgroupElect`)

## Options

1. **Pin `requiredSubgroupSize = 32` at pipeline creation** via `VK_EXT_subgroup_size_control` (already detected as `HasSubgroupSizeControl` in `VulkanBackend.cs:312`, but currently unused by `ComputePipeline`). This is a single-point fix that makes the **entire** existing shader suite correct on Wave64 with **zero shader edits** — preferred if the extension is reliably available.
2. **Manual 32-lane reduction** with `subgroupShuffleDown(acc, 16/8/4/2/1)` + `if (lane == 0)` write, applied uniformly across all reduction shaders. Works regardless of subgroup size but touches every shader.

## Notes

- Raised by an automated review on PR #317. Fixing only the two new Q8_0/Q4_0 shaders was intentionally deferred here so the fix can be applied consistently across the whole shader suite rather than diverging in two kernels (which would not actually enable Wave64, since the rest of the pipeline — RmsNorm, attention, etc. — shares the same assumption).
- Needs validation on actual AMD Wave64 hardware (not available in the current dev environment, which is NVIDIA-only).

pekkah:
Closes #318.

## Problem
The Vulkan reduction shaders (matvec / attention / etc.) pack "8 rows × 32 lanes" in a 256/128-thread workgroup and use `subgroupAdd`/`subgroupElect` **assuming a 32-wide subgroup**. On **AMD Wave64** (subgroup size 64), a subgroup spans two 32-lane row groups → `subgroupAdd` sums across two rows and `subgroupElect` writes only half the rows → **garbage output**. The Vulkan backend explicitly targets cross-vendor (AMD / Intel / NVIDIA), so this is a real correctness gap for the stated platforms.

## Change
Pin `requiredSubgroupSize = 32` via `VK_EXT_subgroup_size_control` so every subgroup is exactly one 32-lane row group on all hardware:
- **`VulkanBackend`**: enable the `VK_EXT_subgroup_size_control` extension + the `subgroupSizeControl` device feature, and query min/max subgroup size (`VkPhysicalDeviceSubgroupSizeControlProperties`) at init.
- **`ComputePipeline`**: chain `VkPipelineShaderStageRequiredSubgroupSizeCreateInfo{requiredSubgroupSize=32}` into the shader stage, **gated** to shaders whose `local_size_x` is a multiple of 32 (the 43 reduction shaders; the 5 sixteen-thread image-op shaders are excluded — pinning 32 on a 16-thread workgroup is invalid, and they don't assume 32-lane subgroups), and only when 32 ∈ [min, max] **and** the device isn't already locked to 32 (`min==max==32`, e.g. NVIDIA → pin is a no-op, skipped).

## Verification
- **No-regression on the dev NVIDIA HW**: subgroup is already 32, so the pin never engages — all 43 Vulkan tests pass with byte-identical output; device + pipelines create cleanly with the feature/extension enabled, and the properties query returns 32/32. New `VulkanSubgroupSizeTests` (10) cover the query + the pin-gate logic.
- **Wiring verified by construction**: an internal code-review pass confirmed (incl. decompiling the Vortice.Vulkan metadata + a runtime `sType` check) that the feature struct is correctly prepended to the device pNext chain and genuinely enabled, the extension-array math has no off-by-one, and the pipeline pNext/pointer-lifetime are correct — i.e. the two test-invisible failure modes (silent feature-enable failure, extension off-by-one) are clean. So the fix will engage correctly on a Wave64 device (min=32/max=64).
- Shader **sources are untouched** → the precompiled SPIR-V table needs no regeneration.

## ⚠️ Verification caveat (please read)
The **Wave64 effect cannot be verified on the available NVIDIA-only hardware** — the pin path engages only on devices that can choose a subgroup size ≠ 32 (AMD Wave64, possibly Intel), which aren't present here. On NVIDIA the change is a verified no-op. Correctness on Wave64 rests on the standard `VK_EXT_subgroup_size_control` mechanism + the wiring review, not a live AMD run.

### DIFF
--- a/src/SharpInference.Vulkan/ComputePipeline.cs
+++ b/src/SharpInference.Vulkan/ComputePipeline.cs
@@ -34,10 +34,11 @@ public sealed unsafe class ComputePipeline : IDisposable
     // Public for tests/diagnostics that need to assert per-recording DS isolation.
     public int CurrentRecordingDispatchCount => _recordingIdx;
 
-    // Matches "local_size_x = <N>" in a GLSL layout qualifier. Every compute shader in this
-    // backend declares it as a literal; used to decide subgroup-size pinning (issue #318).
+    // Matches "local_size_x = <N>" inside a GLSL layout(...) qualifier (scoped to the
+    // qualifier so a stray mention in a comment can't false-match). Every compute shader in
+    // this backend declares it as a literal; used to decide subgroup-size pinning (issue #318).
     private static readonly Regex LocalSizeXRegex =
-        new(@"local_size_x\s*=\s*(\d+)", RegexOptions.Compiled);
+        new(@"layout\s*\([^)]*local_size_x\s*=\s*(\d+)", RegexOptions.Compiled);
 
     /// <summary>
     /// Parse the <c>local_size_x</c> workgroup dimension from GLSL source.
--- a/src/SharpInference.Vulkan/ComputePipeline.cs
+++ b/src/SharpInference.Vulkan/ComputePipeline.cs
@@ -1,3 +1,4 @@
+using System.Text.RegularExpressions;
 using Vortice.Vulkan;
 
 namespace SharpInference.Vulkan;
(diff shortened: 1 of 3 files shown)


pekkah:

pekkah:
Applied — scoped the regex to `layout\s*\([^)]*local_size_x\s*=\s*(\d+)` so a mention outside a `layout(...)` qualifier (e.g. a comment) can't false-match. Updated the whitespace test to a layout-qualified string and added a negative case (comment mention → 0). All 10 subgroup tests pass.

PR #330, opened 2026-06-21
fix(vulkan): dispose TurboQuant pipelines — fixes flaky 0xC0000005 (#153)

### LINKED ISSUE #153: Vulkan: 0xC0000005 access violation in GpuForwardPass.CopyBuffer (TurboQuant past fp32 window) aborts the ForwardPass test suite
The full `SharpInference.Tests.ForwardPass` suite **aborts non-deterministically** with a hard native crash — `0xC0000005` (access violation) — in the Vulkan forward pass. It does not surface as a test failure (`Failed: 0`); the test host process dies, so the run reports "Test Run Aborted" after a variable number of passing tests (observed abort points: 58, 141, 382 across runs).

### Crash
```
[GpuForwardPass] Uploading 36 layers to VRAM....... emb... done.
Fatal error.
0xC0000005
   at SharpInference.Engine.GpuForwardPass.CopyBuffer(SharpInference.Core.Tensor, SharpInference.Core.Tensor)
   at SharpInference.Engine.GpuForwardPass.Forward(Int32, Int32)
   at SharpInference.Tests.ForwardPass.VulkanShaderTests.GpuForwardPassTurboQuantRunsPastFp32Window()
```
- **Test:** `VulkanShaderTests.GpuForwardPassTurboQuantRunsPastFp32Window` (TurboQuant TQ3, context size 32, decoding past the fp32 window).
- **Site:** `GpuForwardPass.CopyBuffer(Tensor, Tensor)` (the **Vulkan** backend), reached from `Forward`.
- **Flaky:** the abort point varies run-to-run → likely a buffer/lifetime hazard (use-after-free / out-of-bounds Vulkan buffer copy), possibly only triggered when prior tests have left the Vulkan device/buffer-pool in a particular state, or specifically when the TQ ring crosses the fp32 window and `CopyBuffer` is handed a stale/too-small buffer.

### Impact
- Blocks a clean full-suite green: the CUDA-only path is unaffected (`--filter "FullyQualifiedName~Cuda"` → 131/131 pass cleanly), but `dotnet test tests/SharpInference.Tests.ForwardPass` cannot complete.
- Pre-existing — surfaced while gating PR #148 (CUDA-only; `git diff` confirms it touches **zero** Vulkan / `GpuForwardPass` files). Should be confirmed reproducible on `master` independent of any branch.

### Next steps
- Repro in isolation: `dotnet test ... --filter "FullyQualifiedName~GpuForwardPassTurboQuantRunsPastFp32Window"` (may need to run after other Vulkan tests to hit the state).
- Audit `GpuForwardPass.CopyBuffer` for buffer sizing / lifetime around the TQ fp32-window boundary; check the Vulkan buffer pool isn''t handing back a recycled/undersized buffer.

pekkah:
Closes #153.

## Root cause
`VulkanBackend.Dispose()` disposed every lazily-created `ComputePipeline` **except the three TurboQuant ones** — `_tqRotateQueryPipeline`, `_tqKvAppendPipeline`, `_tqAttentionPipeline` were declared but missing from the disposal list. Each leaked pipeline also leaks its descriptor pools.

Across repeated backend lifecycles (the full `~Vulkan` / ForwardPass test churn), destroying a `VkDevice` with live child objects is undefined behavior; on this driver it corrupts internal descriptor/pipeline state, which later surfaces as the flaky native **`0xC0000005`** access violation inside a recorded compute dispatch (`CopyBuffer` → `vkCmdBindDescriptorSets`) in a **subsequent** backend — exactly matching #153's variable abort points (58/141/382...) and the prior triage's `vkCmdBindDescriptorSets` observation. Only the TurboQuant path creates these pipelines, so only `GpuForwardPassTurboQuantRunsPastFp32Window` was the trigger.

## Fix
Add the three missing `?.Dispose()` calls. The disposal list is now **complete** (47 pipeline fields ↔ 47 disposals, verified 1:1 in review).

## Diagnosis tooling (kept; opt-in, default-off)
This was diagnosed by adding **opt-in Vulkan validation layers** (`SHARPI_VULKAN_VALIDATION=1`): enables `VK_LAYER_KHRONOS_validation` + `VK_EXT_debug_utils` with an AOT-safe `[UnmanagedCallersOnly]` debug callback writing `[VK-VALIDATION]` messages to stderr. Validation pinpointed `VUID-vkDestroyDevice-device-05137` ("VkPipeline/... has not been destroyed"). The **default path (env var unset) is byte-identical** — no layers, no extension, no messenger, zero overhead; the messenger is destroyed before the instance. Useful future Vulkan-debugging infra.

## Verification (deterministic)
- **With the leak** (fix reverted): **5/5** churn runs crash with `0xC0000005` + the `VUID-vkDestroyDevice` validation error.
- **After the fix**: validation-clean (the VUID error is gone — deterministic proof the leak is closed) and the `~Vulkan` filter passes **47/47**, repeatedly, with no AV.
- Build clean (TreatWarningsAsErrors + AOT analyzers); internal code-review pass confirmed the disposal list is complete and the default-off path is unchanged.

### DIFF
--- a/src/SharpInference.Vulkan/VulkanBackend.cs
+++ b/src/SharpInference.Vulkan/VulkanBackend.cs
@@ -39,6 +39,12 @@ public sealed unsafe class VulkanBackend : IComputeBackend, IImageOpsBackend, ID
     // Serializes vkQueueSubmit from both main and background threads.
     private readonly object _queueLock = new();
 
+    // Opt-in Vulkan validation layers (SHARPI_VULKAN_VALIDATION=1). Used to turn
+    // flaky native access-violations into deterministic validation diagnostics
+    // (issue #153). Null when validation is off (the default).
+    private VkDebugUtilsMessengerEXT _debugMessenger;
+    private readonly bool _validationEnabled;
+
     private bool _disposed;
 
     public string Name { get; }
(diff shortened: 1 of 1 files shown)


pekkah:

pekkah:
Applied all four review suggestions: wrapped the `[UnmanagedCallersOnly]` callback body in try/catch (a managed exception crossing into the driver would fail-fast), switched native-string decoding to `Marshal.PtrToStringUTF8` (correct UTF-8 vs ANSI), used `"..."u8` literals for the layer/extension names (no heap alloc), and made the env-var check case-insensitive. Re-verified: build clean, ~Vulkan 47/47 (no AV), validation-on run clean (no `VUID-vkDestroyDevice`).

PR #331, opened 2026-06-21
feat(vulkan): split-KV flash-decoding attention for long context (#312)

### LINKED ISSUE #312: Vulkan: attention limited by 4096-slot shared-memory fast path — no flash-attention / split-KV

pekkah:
Closes #312.

## Problem
Vulkan attention spilled per-position softmax scores to a `numHeads × maxSeqLen` VRAM buffer beyond seq_len 4096 — slow at long context (it round-trips all scores through VRAM, and a single workgroup per head serialises the KV scan). Two earlier single-workgroup online-softmax attempts **regressed ~3-4×** for exactly that reason (they removed the spill but kept the serial scan — see the #312 history).

## Change
Ports CUDA's split-KV / flash-decoding (#235): the KV range is split into 512-position chunks across **`numHeads × nSplits` workgroups** (restoring seq_len parallelism, scores kept in shared memory — no VRAM spill), each emitting an un-normalised weighted-V partial + `(m, l)`; a **combine pass** LSE-merges the partials per head. **fp32 only** for now (bf16/q8_0 keep the spill — a follow-up).

- New GLSL: `AttentionSplitKvPartial` (2D dispatch `numHeads × nSplits`, online softmax over each chunk) + `AttentionSplitKvCombine` (LSE merge). Both `local_size_x=256` (subgroup pinned to 32 by #318); plain shared-mem reductions, no subgroup ops.
- `VulkanBackend.AttentionSplitKv` (2-pass dispatch + barrier; guards `nSplits ≤ 256`). `GpuForwardPass` allocates two small partial buffers (`numHeads × maxSplits × (headDim+2)` floats ≈ 0.4 MB — *smaller* than the spill it replaces) and gates the fp32 decode attention to split-KV for `seq_len>4096 && headDim%32==0`.
- **Default-ON**; `SHARPI_VULKAN_SPLIT_DECODE=0` reverts to the spill (kill switch), mirroring CUDA's `SHARPI_SPLIT_DECODE`. The ≤4096 / `headDim%32!=0` / fp32-spill paths are **untouched**.

## Verification
- **Perf** (Qwen3-8B, `--backend vulkan -g -1 --kv-type fp32`, ~10.5K context): decode **15.3 → 30.4 t/s (~2×)**, prefill **23.7 → 34.6 (+46%)**. Above noise; the win grows with context (CUDA #235 saw 2-4× at 32K). [Why moderate context shows nothing: at ≤8K attention is only ~5% of per-token decode; the win materialises as attention's `O(ctx)` share grows.]
- **Parity**: 3 `AttentionSplitKv` tests match the CPU SDPA+softmax+GQA reference `<1e-3` (multi-split, GQA, tiny tail). The combine LSE handles empty/short splits without NaN.
- **No regression**: full `~Vulkan` suite **50/50** with split-KV default-on; ≤4096 byte-identical; internal code-review pass clean (combine LSE / no-NaN-on-empty-split / default safety verified).

## Scope / follow-ups
- fp32 only. Extending split-KV to **bf16/q8_0 KV** (which is what enables the 20K–40K contexts where the win is largest) is the natural follow-up — the partial/combine kernels just need the narrowed K/V reads.
- Split-KV engages only for fp32 KV >4096 with `headDim%32==0`; everything else keeps the existing path.


pekkah:

pekkah:
Applied all three: bounded the split-count math against overflow by checking the seqLen/maxSeqLen bound directly (≤131072 = 256×512) before the `+511` — `GpuForwardPass` allocates only when `maxSeqLen<=131072`, and `AttentionSplitKv` throws when `seqLen>131072`. Fixed the TinyTail test doc comment (4097 → 9 splits). Parity 3/3 still pass, build clean.

PR #333, opened 2026-06-21
feat(vulkan): extend split-KV flash-decoding to bf16/q8_0 KV (#332)

### LINKED ISSUE #332: perf(vulkan): extend split-KV flash-decoding to bf16/q8_0 KV (#312 follow-up)
## Summary

#312 (PR #331) added split-KV flash-decoding to the Vulkan path for **fp32 KV only**, with a measured ~2× decode win at ~10.5K context (`Qwen3-8B`, 15.3→30.4 t/s). But fp32 KV auto-caps at ~11K context on a 12 GB GPU — the contexts where split-KV's win is **largest** (20K–40K) are reached via **bf16/q8_0 KV** (#311/#325), which currently still use the slow VRAM score-spill at >4096.

## Proposed work

Extend the split-KV path (`AttentionSplitKvPartial` / `AttentionSplitKvCombine` + `VulkanBackend.AttentionSplitKv` + the `GpuForwardPass` gate) to bf16 and q8_0 KV. The partial/combine kernels are dtype-agnostic except the K/V reads — mirror the existing narrowed-KV read idioms:
- bf16: `unpackHalf2x16(k_cache[(off+d)>>1])[(off+d)&1]`
- q8_0: the `block_q8_0` dequant (`gByte` byte-gather, `dsc*int8`) used by `AttentionQ8_0`.

So: add bf16/q8_0 variants of `AttentionSplitKvPartial` (combine is dtype-agnostic — it only reads the fp32 partial buffers), gate them in `GpuForwardPass`'s bf16/q8_0 dispatch branches (currently they call `AttentionBf16`/`AttentionQ8_0`), and dispatch split-KV for `seq_len>4096 && headDim%32==0` like the fp32 path.

## Why it's high value
- bf16 KV reaches ~22K context, q8_0 ~40K (full model context) on a 12 GB GPU. At those lengths attention dominates per-token decode and the spill's `numHeads×maxSeqLen` VRAM round-trip is the bottleneck — split-KV should give the 3-4× the CUDA #235 work saw at 32K (vs the ~2× already measured for fp32 at the fp32-capped 10.5K).
- Lets the perf win be benchmarked at the contexts where it actually matters (the fp32 PR could only reach ~10.5K).

## Notes
- Reuses everything from #312 (the partial buffers, the combine, the 2-pass dispatch, the kill-switch `SHARPI_VULKAN_SPLIT_DECODE=0`); only the partial kernel's K/V read changes per dtype.
- Verify: parity (bf16/q8_0 split-KV vs CPU <1e-3 / coherence) + a 20K-40K decode benchmark (split-KV vs spill) — where the big win lives.

pekkah:
Closes #332. Follow-up to #312 (split-KV for fp32 KV).

## Why
#312 added split-KV flash-decoding for **fp32 KV** (~2× decode at 10.5K). But fp32 KV auto-caps at ~11K context on 12 GB — the 20K-40K contexts where split-KV's win is largest are reached via **bf16/q8_0 KV** (#311/#325), which still used the slow VRAM score-spill at >4096. This closes that gap.

## Change
The split-KV algorithm is unchanged; **only the partial kernel's K/V read narrows per dtype** (the combine kernel + the fp32 partial buffers are dtype-agnostic and reused):
- `AttentionSplitKvPartialBf16` — fp16-packed read (`unpackHalf2x16`), mirroring `AttentionBf16`.
- `AttentionSplitKvPartialQ8` — `block_q8_0` dequant (`gByte` + `bitfieldExtract`), mirroring `AttentionQ8_0`.
- `VulkanBackend.AttentionSplitKvBf16`/`AttentionSplitKvQ8` reuse the shared combine pipeline; `GpuForwardPass` gates split-KV in the bf16/q8_0 decode branches with the same condition as fp32 (`seq_len>4096 && headDim%32==0`).
- Default-ON; `SHARPI_VULKAN_SPLIT_DECODE=0` kill switch. ≤4096 / split-disabled paths byte-identical.

## Verification
- **Parity**: 4 new direct-shader tests (bf16 + q8_0, each GQA-multi-split + tiny-tail) build the packed KV cache (fp16-packed / `block_q8_0` via the `KvAppendQ8_0` recipe) and match a CPU SDPA reference dequantized from the **same** bytes `<1e-2` (isolating split-KV from quant loss). Full `~Vulkan` suite **54/54** (no regression).
- **Perf** (Qwen3-8B, `--backend vulkan -g -1 --kv-type bf16`, ~13.6K ctx): split-KV decodes **28.2 t/s and completes the run**; the spill path (`SHARPI_VULKAN_SPLIT_DECODE=0`) **can't finish the same run in 560s** — it collapses at this context. The algorithm is identical to fp32 split-KV, which measured a clean **~2×** at 10.5K (#312). q8_0 uses the same path and reaches ~40K context.
- Code-review pass clean (the narrowed bf16/q8_0 reads verified bit-for-bit against the cache writers; everything else identical to the proven fp32 split-KV).


pekkah:

pekkah:
Applied all four perf suggestions: the bf16/q8_0 partial shaders now read each packed word / block scale once (bf16 paired `unpackHalf2x16`; q8_0 reads the fp16 block scale once per 32-elem block instead of 32× redundantly) and hoist the address math out of the Phase-3 inner loop — mirroring the existing `AttentionBf16`/`AttentionQ8_0`. Removed the now-unused `loadK`/`loadV` helpers. Perf-only: split-KV parity 7/7 still pass <1e-2, bf16 e2e coherent. SPIR-V regenerated.

Issue #309, opened 2026-06-21
Vulkan: Gemma 2/3/4 unsupported — missing SWA, logit softcap, GELU-tanh, per-layer head_dim kernels


pekkah:
## Summary

`GpuForwardPass` explicitly rejects Gemma 4 (and the family of features it shares with Gemma 2/3) at construction time, and `VulkanBackend` is missing the underlying kernels. These models run on CPU and CUDA but not on the Vulkan GPU path.

## Evidence

- `src/SharpInference.Engine/GpuForwardPass.cs:137-145` — throws `NotSupportedException` when `hp.LayerHeadDim is not null` (the Gemma 4 master switch): "no SWA / PLE / per-layer head_dim / softcap kernels".
- `VulkanBackend` default-throws the required ops (none are implemented):
  - `GeluTanhMul()` — Gemma fused GELU-tanh × up.
  - `SoftcapInPlace()` — attention/final-logit softcap (Gemma 2/3).
  - `AttentionSwa()` — sliding-window attention.
- No shader exists in `Shaders.cs` for SWA, softcap, or GELU-tanh, and there is no per-layer `head_dim` handling or PLE (Progressive Layer Embedding) injection seam.

## What CUDA has

`CudaBackend` implements `GeluTanhMul`/`GeluTanhMulStrided`, `SoftcapInPlace`, `AttentionSwa`/`AttentionSwaBatched`, per-layer head_dim, and PLE pre-pass dequant (`DequantRowsQ8_0`, issue #247). Gemma-4 tied-embedding lookups (`EmbedLookupQ6K`, issue #124) are present too.

## Proposed work

- Add GLSL shaders + `VulkanBackend` methods for `GeluTanhMul`, `SoftcapInPlace`, and `AttentionSwa`.
- Support per-layer `head_dim` and a PLE injection seam in `GpuForwardPass`, then lift the init-time `NotSupportedException`.

## Related

- #82 (add Gemma 4 model family support — architecture-level)
- #252 (Gemma 4 vision input on Vulkan / ForwardEmbedding seam)

This issue tracks the Vulkan compute-kernel gap specifically.

pekkah:
### Progress: PR1 merged (#334) — GELU-tanh + final-logit softcap Vulkan kernels

[...]

PR #334, opened 2026-06-21
feat(vulkan): GELU-tanh + final-logit softcap kernels (toward Gemma, #309)

pekkah:
First bounded step toward #309 (Gemma on Vulkan).

## Context
Investigating #309 showed its "Gemma 2/3/4" framing is broader than the code: only the **gemma4** arch is wired anywhere (CPU/CUDA/loader), and end-to-end Gemma-on-Vulkan is multi-session (the big risks are SWA masking across all attention variants + per-layer head_dim trunk surgery). This PR lands the two **independently unit-testable** kernel gaps first.

## Change
The Vulkan backend threw `NotSupportedException` on two ops Gemma needs; this implements both as compute shaders:
- **`GeluTanhMul`**: `gate[i] = 0.5·g·(1+tanh(0.7978845608·(g + 0.044715·g³)))·up[i]` (g=gate[i], in place) — a `SiLuMul` clone with the activation swapped. Matches `SimdKernels.GeluTanhMul` / CUDA `llm_gelu_tanh_mul`.
- **`SoftcapInPlace`**: `x[i] = tanh(x[i]/cap)·cap` — a `ScaleInPlace`-shaped elementwise op. Matches `SimdKernels.SoftcapInPlace` / CUDA `llm_softcap_inplace`.

`VulkanBackend` overrides the two `IComputeBackend` methods (both pipelines lazy-created + disposed). Precompiled SPIR-V regenerated (52 entries; both new shaders precompile cleanly).

## Verification
- **Unit parity vs CPU** (`<1e-3`): `GeluTanhMulMatchesCpu` (gate ∈ [-4,4]), `SoftcapMatchesCpu` (x ∈ [-100,100], cap=30, N=2049 to exercise the bounds guard). No model needed — fully deterministic.
- Full `~Vulkan` suite **56/56**, no regression. Pure new kernels — no behavior change to any existing path.
- Internal code-review: formulas/constants confirmed against CPU+CUDA, cap field correct, no-regression.

## Follow-ups (toward end-to-end Gemma on Vulkan)
Sandwich-norm + embedding-scale + query-scale-override wiring; per-layer head_dim + dual RoPE; SWA window masking across all attention variants; then end-to-end gemma4 enablement + a teacher-forced parity test vs CPU.


pekkah:

pekkah:
Thanks — considered all three; declining them to stay consistent with the established backend conventions (the PR is correct + parity-verified as-is):

1. **gate/up size guard & 2. cap>0 guard** — the sibling kernel-dispatch methods on `VulkanBackend` (`SiLuMul`, `ScaleInPlace`, `AddInPlace`, `AddScaledInPlace`) deliberately carry **no** argument guards: they're internal hot-path dispatchers with caller-guaranteed invariants (`gate`/`up` are always the same FFN tensor pair; `cap` is always the model's `final_logit_softcapping` constant, 30/50, never ≤0). Adding guards to only the two new methods would be inconsistent. If guard-on-dispatch were the desired convention it'd belong on all of them, as a separate change.

3. **Factored GELU formula** — `0.7978845608·(g + 0.044715·g³)` is kept intentionally to mirror the CPU (`SimdKernels.GeluTanhMul`) and CUDA (`llm_gelu_tanh_mul`) structure for cross-backend parity-intent. The factored `g·(0.7978845608 + 0.0356774·g²)` is math-equivalent but changes the FP rounding order (diverging from the reference at the ULP level) to save a single multiply on a path dominated by the FFN matvecs, not the activation. This repo only ships micro-opts that are bit-identical or measured wins; this is neither.

Merging — parity tests pass <1e-3 vs the CPU references, full ~Vulkan suite 56/56.

PR #335, opened 2026-06-21
feat(vulkan): sliding-window attention (SWA) masking (toward Gemma, #309)

pekkah:
Toward #309 (Gemma on Vulkan). Follow-up to #334 (GELU-tanh + softcap).

## Change
Gemma SWA layers attend only to the last `window` KV positions. This adds a `window` push-constant to the Vulkan attention shaders, restricting score / softmax / V-aggregation to `[startSeq, seqLen)` where:
```
startSeq = (window != 0 && window < seqLen) ? seqLen - window : 0
```
(uint-underflow-safe; mirrors `ForwardPass.Attention`'s `windowSize` logic). **`window == 0` is full attention — byte-identical to before.**

Covers all **six** attention shaders:
- **Single-workgroup** (`Attention`, `AttentionBf16`, `AttentionQ8_0`): every phase (score / max / exp-sum / normalize / V-agg) loops from `startSeq`.
- **Split-KV partials** (fp32/bf16/q8): a chunk fully below the window writes empty meta (`m=-inf, l=0`) and bails (the combine skips it via `exp(-inf)=0`); a straddling chunk clamps `t0 = max(t0, startSeq)`. q8 block alignment preserved (`kv_base` stays a multiple of 32). Combine shader unchanged.

`window` appended to `AttentionParams` + `SplitKvPartialParams`, threaded through the six `VulkanBackend` dispatch methods (default 0). **All `GpuForwardPass`/`HybridForwardPass` call sites pass `window:0` → no behavior change** (per-layer SWA wiring is a later PR).

## Verification
- **Unit parity vs a CPU windowed SDPA reference** (`<1e-3`): fp32/bf16/q8 masked (window<seqLen), window≥seqLen→full, **window=0→full (regression guard)**, GQA, and split-KV chunk-skip (`seqLen=5000, window=1024` → `startSeq=3976`, forcing whole chunks skipped + one clamped). 10 new tests.
- Existing full-attention parity **unchanged** (confirms window=0 byte-identical); full `~Vulkan` suite **64/64**.
- Internal code-review (live GPU): no-regression, uint-underflow, split-KV chunk-skip / q8-alignment / combine-NaN-safety all verified PASS.

## Why now
[...]

PR #336, opened 2026-06-22
feat(vulkan): end-to-end Gemma 4 inference (#309)

### LINKED ISSUE #309: Vulkan: Gemma 2/3/4 unsupported — missing SWA, logit softcap, GELU-tanh, per-layer head_dim kernels

pekkah:
Closes #309. The finale — builds on the merged Vulkan kernels (#334 GELU-tanh + softcap, #335 SWA window masking).

## Change
Wires the full gemma4 trunk into `GpuForwardPass` (new `ForwardGemma4`, gated on `hp.LayerHeadDim`), mirroring `CudaForwardPass.RunGemma4DeviceRegion`:
- **per-layer head_dim** (512 global / 256 SWA) + **per-layer KV heads** via Tensor views into max-sized scratch; per-layer KV cache allocation
- embedding scale (√d); **Q/K per-head norm + V-norm** (`HeadNormPure`); **dual RoPE theta** (global/SWA); **`rope_freqs` on global layers** (new `RoPEWithFactors` shader)
- **attnScale=1.0** via Q pre-scale; **SWA window** per layer; **k_eq_v** (V = raw K copy where `attn_v` absent); **sandwich norm** (post-attn/post-ffw before residual); **GELU-tanh** FFN; per-layer output scale; **final-logit softcap**
- **SnapKV force-disabled for gemma4** (per-layer dims would mis-index the score/compaction shaders — mirrors CUDA)

New raw **`EmbedLookupQ6K`**: gemma4-v2's Q6_K `token_embd` (787 MiB) was dequantizing to ~4 GB F32, OOMing the 12 GB GPU; now it stays packed. Q4_K + F32 embedding paths unchanged.

## No-regression
Every gemma4 change is behind `_isGemma4`; for non-gemma4 models `_maxHeadDim == _headDim` and the dense path is **byte-identical** — the only change is a Q6_K embedding newly taking the raw path (verified equivalent). Full `~Vulkan` suite **66/66**.

## Verification
- **gemma4-v2-Q4_K_M greedy output on Vulkan is BYTE-IDENTICAL to CUDA** (cross-backend, argmax-stable) — e.g. `"The capital of France is Paris"` and a 40-token continuation matching exactly. This is the load-bearing correctness signal (the full trunk: per-layer head_dim, SWA, dual RoPE, qk/V-norm, sandwich norm, softcap).
- `EmbedLookupQ6K` parity `<1e-2` vs CPU Q6_K dequant; build clean; internal code-review found + fixed the SnapKV-vs-gemma4 guard.
- Skippable `Gemma4GpuForwardPassTests` (CPU-vs-Vulkan logit parity; silent-skips without the model / Vulkan / on OOM).

## Scope
PLE (per-layer embeddings) and shared-KV stay out of scope — both disabled in gemma4-v2. Run with an explicit `-c` (e.g. 4096) for now; the per-layer full-context KV (no SWA ring on Vulkan) makes the model-max auto-context large.


pekkah:

pekkah:
Thanks — declining for this PR, with rationale:

This per-layer view pattern (`new Tensor(TensorShape.D1(len), DType.Float32, _q.Handle)`) **mirrors `CudaForwardPass.RunGemma4DeviceRegion` exactly** (CudaForwardPass.cs:1926-1928), which is the proven, shipped reference this PR ports — and which produces the byte-identical greedy output the PR verifies against. The views also have **per-layer-varying lengths** (head_dim 256 SWA / 512 global, per-layer KV heads), so they aren't a single cacheable view-per-role; eliminating the alloc would mean 48×4 per-layer cached views.

The allocation is ~hundreds of tiny gen0 objects per token, dwarfed by the 48 layers of matvecs that dominate each gemma4 decode step — not a measured bottleneck. If view-allocation GC ever shows up in a profile, it should be addressed **uniformly across the CUDA and Vulkan gemma4 paths** (so they don't diverge), as a dedicated optimization — not a one-off in this enablement PR.

Merging: gemma4-v2 greedy output is byte-identical to CUDA; full ~Vulkan suite 66/66; non-gemma4 path byte-identical.

Issue #252, opened 2026-06-15
feat(vision): Gemma 4 image input on partial-offload hybrids + Vulkan (ForwardEmbedding seam)


pekkah:
Follow-up to #250 / PR #251 (Gemma 4 12B encoder-free `gemma4uv` image input).

Today `IForwardPass.ForwardEmbedding` (the multimodal soft-token splice seam) is implemented only on:
- `ForwardPass` (CPU), and
- `CudaForwardPass` (CUDA **full** offload, `-g -1` of a model that fits VRAM).

`SupportsEmbeddingInput` is `false` (interface default-throw) on the partial-offload hybrids and Vulkan, so `--image` errors there with a backend-aware message.

## Scope
Implement `SupportsEmbeddingInput`/`ForwardEmbedding` (faithful mirrors of the CPU/CUDA seam — skip the sqrt(d) embedding scale, build PLE from padding token 0, run the normal trunk + KV append) on:
- [ ] `CudaHybridForwardPass` (dense/MoE partial offload)
- [ ] `GpuForwardPass` (Vulkan, full offload)
- [ ] (if practical) `HybridForwardPass` (Vulkan partial offload)

Then drop the CLI's implicit restriction so `--image` works under `-g N` / `--backend vulkan` once the chosen pass reports `SupportsEmbeddingInput`.

## Notes
- The contract is documented on `IForwardPass.ForwardEmbedding` (skip sqrt(d) scale + PLE-token-0).
- CPU/CUDA are the reference implementations; verify argmax parity against the CPU path on a fixed image+prompt.

pekkah:
**Vulkan full-offload vision: DONE (#337).** `GpuForwardPass.ForwardEmbedding` + `SupportsEmbeddingInput` implemented; gemma4 image input now works on `--backend vulkan -g -1`. Verified RED→"Red" / BLUE→not-red on GPU, and a real-photo CLI A/B matches CUDA near-identically. Review caught + fixed a transfer→compute barrier.

Remaining for this issue: **partial-offload hybrids** (`HybridForwardPass`/`CudaHybridForwardPass` have no `ForwardEmbedding`/`SupportsEmbeddingInput`) — the embedding seam must thread the CPU-resident + GPU-offloaded layers + the hybrid trunk, a separate/larger effort. Leaving open for that part.

pekkah:
## CUDA partial-offload vision: DONE (#349 / master 9b46157)

[...]

PR #337, opened 2026-06-22
feat(vulkan): Gemma 4 image input (ForwardEmbedding seam) (#252)

pekkah:
Toward #252 — Gemma 4 image input on the Vulkan full-offload path. Follow-on to #309 (gemma4 text on Vulkan).

## Change
Gemma 4 encoder-free vision injects image soft-token embeddings into the prefill via `IForwardPass.ForwardEmbedding`. CPU + CUDA implement it; Vulkan didn't. Everything above the forward pass — the CPU `GemmaUvVisionEmbedder`, `InferenceEngine.SplicePrefillImages`, the CLI `--image` path — is backend-agnostic and already routes the active forward pass. So the only missing piece was the Vulkan seam:
- **`GpuForwardPass.SupportsEmbeddingInput => _isGemma4`** (was the `false` default)
- **`GpuForwardPass.ForwardEmbedding`**: mirrors `ForwardGemma4` with the embedding-lookup prologue replaced by uploading the provided soft-token embedding into `_hidden` with **NO √d scale** (soft tokens arrive already-final, per the contract + CPU/CUDA), via `UploadToExisting` *before* `BeginRecord`; then the identical `RunGemma4Layers` + final-norm/output/softcap/download tail. Guards: gemma4-only + `embedding.Length == embDim`. No PLE pre-pass (Vulkan gemma4 rejects PLE at construction).
- CLI `--image` help text: drop the stale "CPU only (-g 0)" note.

**Purely additive** — non-gemma4 models report `SupportsEmbeddingInput=false` (unchanged); the dense/MoE paths are byte-identical.

## Verification
- New model-gated `Gemma4VisionVulkanE2ETests` (gemma4-v2 + the 12B gemma4uv mmproj) **runs on GPU**: solid-RED image → "Red", solid-BLUE → not "red" (image-dependence + correctness).
- **CLI A/B on a real photo**: Vulkan → *"A misty lake sits beneath a pale sky, with pine trees and mountains rising through the fog"* vs CUDA → *"...and a distant mountain rising through the fog"* — near-identical (minor late divergence = expected cross-backend float variance in the vision soft-tokens).
- Full `~Vulkan` suite **67/67**; build clean; internal code-review clean (no-scale + upload-before-record + no-regression verified).

## Scope
Full-offload `GpuForwardPass`. Partial-offload `HybridForwardPass` vision is a follow-up (separate forward pass, no embedding seam yet).

### DIFF
--- a/src/SharpInference.Cli/RunCommand.cs
+++ b/src/SharpInference.Cli/RunCommand.cs
@@ -32,7 +32,7 @@ public sealed class RunCommand : Command<RunCommand.Settings>
         public string? PromptFile { get; init; }
 
         [CommandOption("--image <PATH>")]
-        [Description("Path to a PNG image for multimodal input (Gemma 4 encoder-free vision). Repeatable for multiple images; reference each with an <image> marker in -p (left-to-right), or omit markers to prepend them. Requires --mmproj and a text prompt (-p). CPU only for now (-g 0).")]
+        [Description("Path to a PNG image for multimodal input (Gemma 4 encoder-free vision). Repeatable for multiple images; reference each with an <image> marker in -p (left-to-right), or omit markers to prepend them. Requires --mmproj and a text prompt (-p). Runs on CPU, CUDA, and Vulkan (full offload).")]
         public string[]? ImagePaths { get; init; }
 
         [CommandOption("--mmproj")]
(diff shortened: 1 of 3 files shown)


pekkah:

pekkah:
Good catch — fixed. `UploadToExisting` is a device transfer in a separate fence-waited submission, and a host fence-wait doesn't make that write visible to the next submission's compute reads. Added `RecordTransferBarrier()` (TransferWrite→ShaderRead, Transfer→ComputeShader) right after `BeginRecord`; its first sync scope includes the earlier transfer submission in queue order, so the first layer's `_hidden` read is correctly ordered after the upload. Spec-correct/portable (it happened to work on NVIDIA without it). Re-verified: gemma4 vision output unchanged.

Issue #308, opened 2026-06-21
Vulkan: no speculative decoding / MTP batched verify (SupportsBatchVerify)


pekkah:
## Summary

The Vulkan forward pass cannot accelerate speculative decoding or self-speculative Multi-Token Prediction (MTP). `GpuForwardPass` does not expose `SupportsBatchVerify`, so `SpeculativeDecoder`, `MtpDecoder`, and `PromptLookupDraft` cannot batch-verify draft tokens against a Vulkan target — they fall back to one-token-at-a-time verification.

## Evidence

- `IForwardPass.SupportsBatchVerify` is implemented by CPU (`ForwardPass.cs:1183`), CUDA dense (`CudaForwardPass.cs:4659`), CUDA hybrid GDN (`CudaHybridGdnForwardPass.cs:3334`), and CPU hybrid GDN (`HybridGdnForwardPass.cs:1241`) — but **not** by `GpuForwardPass`.
- `SpeculativeDecoder.cs:475` gates the fast path on `_target.SupportsBatchVerify`.
- `MtpDecoder.cs:197,348` gate batched MTP verify on `_fwd.SupportsBatchVerify`.

## What CUDA has

- `BatchForward2` / `MatMulN2` dual-output matvec amortizes weight reads across two draft tokens (MTP, issue #43).
- Folded k-token batched verify (`CudaForwardPass.cs:4693`, issue #207) plus on-device argmax (`SupportsBatchVerifyArgmax`, issue #219).

## Proposed work

- Add a batched/folded verify matvec path on Vulkan (process k draft tokens against one weight read) and implement `SupportsBatchVerify` on `GpuForwardPass`.
- Optionally add a GPU argmax shader so draft acceptance avoids full-vocab download.

## Impact

`--mtp` and `--draft-model` provide no speedup on the Vulkan backend today.

pekkah:
## Vulkan speculative decoding: shipped (7 PRs). Summary + honest perf.

Prompt-lookup speculative decoding now works end-to-end on the Vulkan backend (full-offload dense Q4_K/Q6_K), opt-in via `--draft-lookup`, lossless (bit-exact greedy parity).

**The arc (each bit-exact + parity-gated):**
- #338 BatchVerify foundation (K-loop reference + rollback + harness)
- #339 / #340 batched weight-stationary matvecs (Q4_K / Q6_K) — the weight amortization
[...]

PR #338, opened 2026-06-22
feat(vulkan): BatchVerify foundation for speculative decoding (#308 PR1)

pekkah:
First PR toward #308 (Vulkan speculative decoding). The user committed to this multi-session feature; this is the verifiable **foundation + test harness**.

## What this is (and isn't)
The core of spec decode is a *batched-K-verify forward* that amortizes weight reads across K draft tokens. `GpuForwardPass` is single-query today, so that batched forward is built incrementally. **This PR ships the `BatchVerify` interface + rollback semantics + the parity-test harness** as a **K-loop reference** (loops the existing `Forward` K times). It is correct and bit-identical to K sequential forwards, but does **not yet amortize weight reads** — so no CLI/spec gate is flipped and default behavior is unchanged. The weight-amortizing batched quantized matvec (the real speedup, the biggest risk) and the spec-loop wiring are the next PRs.

## Change
- **`SupportsBatchVerify => !_isGemma4 && !_tqEnabled && !_kvEvicted`** — conservative gate (the K-loop is correct for all of these, but the names are reserved for the future packed batched forward, which per-layer-head_dim gemma4 / the TQ ring / SnapKV-compacted caches would break). Mirrors CUDA's `_kvEvictedCount==0`.
- **`BatchVerify`**: `TruncateTo(startPos)` then loop `Forward(tokens[i], startPos+i)`, returning K logit arrays (each `.ToArray()`'d since `Forward` reuses its logits buffer); leaves the cache at `startPos+k`, per the contract.
- New minimal **`_kvEvicted`** flag: set only on an actual SnapKV compaction, reset in `ResetCache`; `TruncateTo` deliberately leaves it (mirrors CUDA) so verify stays off on a compacted cache until reset.

## Verification
- New `VulkanSpecBatchVerifyTests` (Qwen3-0.6B, on GPU): **BatchVerify == K sequential Forward** (exact argmax, `<1e-4`), **truncate-and-commit rollback** (rejected K/V tail correctly ignored after `TruncateTo`), gate assertion. Silent-skips without Vulkan/model.
- Full `~Vulkan` suite **71/71**, no regression; build clean; review clean (`.ToArray()` aliasing, contiguous positions, `_kvEvicted` lifecycle all verified).

## Roadmap (#308)
PR1 (this): BatchVerify foundation + harness. → PR2: batched quantized matvec shader (Q4_K, the weight-amortizing kernel — biggest risk, model-free unit-tested). → PR3: re-impl BatchVerify on the batched trunk (real speedup) + micro-bench. → PR4: flip the CLI gate for PromptLookupDraft spec, E2E greedy parity (spec is lossless). → draft-model spec, argmax fast path, gemma4.

### DIFF
--- a/src/SharpInference.Engine/GpuForwardPass.cs
+++ b/src/SharpInference.Engine/GpuForwardPass.cs
@@ -90,6 +90,60 @@ public sealed unsafe class GpuForwardPass : IForwardPass
     /// <inheritdoc />
     public bool SupportsPartialRewind => true;
 
+    // ── Speculative-decode batched verify (issue #308) ───────────────────────────────────
+
+    /// <summary>
+    /// Whether <see cref="BatchVerify"/> can run on this Vulkan dense full-offload pass. Gated to
+    /// the dense path (non-Gemma-4 — its per-layer head_dim / SWA rings / shared-KV / softcap need
+    /// a separate batched path, a later PR — and non-TurboQuant, whose ring bookkeeping breaks the
+    /// contiguous [startPos, startPos+k) layout) with an uncompacted cache. Mirrors
+    /// <see cref="CudaForwardPass.SupportsBatchVerify"/>: a CONFIGURED SnapKV budget does not
+    /// disable verify — only an actual prefill-time eviction does (then physical slot != logical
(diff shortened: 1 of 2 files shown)


pekkah:

PR #339, opened 2026-06-22
feat(vulkan): batched weight-stationary Q4_K matvec (#308 PR1b)

pekkah:
#308 PR1b — the weight-amortizing core for Vulkan speculative decoding. Follows #338 (BatchVerify foundation).

## Change
The single-row `MatVecQ4K` reads+dequantizes each weight block once per output element and multiplies by ONE input vector. The batched (M=K) version computes the dequantized weight value **once** and multiplies it into **K** accumulators (one per draft token), so the weight is read from VRAM once for all K tokens — the spec-decode win.

- **`Shaders.MatVecBatchedQ4K`**: verbatim `MatVecQ4K` (8 rows × 32 lanes, same block/element order, same d/dmin + 8 (scale,min) unpack) generalized to K inputs. `float acc[8]` (nTok ≤ 8); per element compute `w = dsc[si]*nibble - dmn[si]` once, then `acc[k] += w * input[k*cols+idx]`; per-k `subgroupAdd`. Input `[nTok][cols]`, output `[nTok][rows]`. Per-(row,k) accumulation order is identical to the single-row shader → **bit-identical** to K separate `MatVecQ4K` calls (the E2E spec greedy-parity contract). `local_size_x=256` → subgroup pinned to 32 (#318).
- **`VulkanBackend.MatMulBatched`**: validates `nTok∈[1,8]` + divisibility; Q4_K → batched shader; other dtypes → a correct K-loop fallback (no amortization yet — Q5_K/Q6_K/Q8_0/F32 batched shaders are follow-ups).

Shader + dispatch + unit test only — **not yet wired into `BatchVerify`** (next PR), like the GELU/softcap kernels (#334) before the gemma4 trunk (#336).

## Verification
- `MatVecBatchedQ4KMatchesSingleRow` (rows=64, cols=512, nTok=1/4/6): **BIT-EXACT** (`==`, no tolerance) vs K single-row `MatMul` calls — the load-bearing property the spec greedy-parity depends on. Model-free.
- Full `~Vulkan` suite **74/74**, no regression; build clean; SPIR-V regenerated; review clean (bit-exactness + subgroup uniformity verified).

## Note for PR1c
The non-Q4_K fallback emits no barriers inside a recording session (unreachable today — no caller). When PR1c wires `BatchVerify`, either add `RecordBarrier()` in the fallback under recording, or use the batched path with Q4_K-only weights. Tracked.

## Roadmap (#308)
PR1 ✓ (BatchVerify foundation, #338) · **PR1b ✓ (this)** · PR1c (wire BatchVerify on the batched trunk + micro-bench) · PR2 (flip CLI gate for prompt-lookup spec, E2E greedy parity) · then Q5_K/Q6_K/Q8_0 batched shaders, draft-model spec, gemma4.


pekkah:

pekkah:
Thanks — applied one, deferred two with rationale:

- **cols%256 validation (applied):** added a guard throwing when `cols` isn't a multiple of 256 for the Q4_K path (the shader derives `num_blocks = cols >> 8`, so a non-multiple would silently drop elements).
- **Fallback barriers (deferred to PR1c):** the non-Q4_K fallback reuses `tmpIn`/`tmpOut` across iterations, which is only hazard-free on the immediate (fence-serialized) dispatch path it currently uses. It has **no caller yet** — it gets wired in the BatchVerify PR (#308 PR1c), where the recording-session path is actually exercised and testable. I documented the constraint in-code and will add the per-iteration barriers (or restrict the batched path to Q4_K) there rather than ship untested barrier code on an unreachable path now.
- **nTok unroll (deferred to PR1c):** unrolling to a fixed 8 with masking keeps `acc[]` register-resident, but it also runs 8 masked iterations regardless of nTok (vs `nTok` today) in the hottest inner loop, so for small nTok (2–4, the common spec lookahead) it can regress. Whether the register-residency win beats the extra iterations is compiler/nTok-dependent — exactly what PR1c's micro-bench is for. Applying it unmeasured here would violate the repo's measure-before-shipping-perf norm.

Parity test still bit-exact (3/3); build clean.

Issue #313, opened 2026-06-21
Vulkan: VK_KHR_cooperative_matrix detected but never used (no tensor-core matvec/GEMM)


pekkah:
## Summary

`VulkanBackend` probes for `VK_KHR_cooperative_matrix` at init and records the capability, but no pipeline ever uses cooperative matrices. All matvecs and SGEMMs run as scalar/subgroup kernels, leaving tensor-core throughput on the table. CUDA, by contrast, leans heavily on int8 tensor-core MMQ.

## Evidence

- `VulkanBackend.cs:311` — `HasCooperativeMatrix = extNames.Contains("VK_KHR_cooperative_matrix");`
- `VulkanBackend.cs:466` / `494` — exposed as a property and added to a "found extensions" list, but never consulted when selecting a matvec/GEMM pipeline.
- Matvec dispatch (`VulkanBackend.cs:1197-1205`) and SGEMM kernels (`SgemmF32/F16/Bf16/Fp8`) are all standard tiled/subgroup shaders.

## What CUDA has

- Int8 tensor-core MMQ for decode and prefill: `MatMulBatchedDecodeMmq`, `MatMulBatchedMmq`, `MatMulBatchedMmqSoa` over Q4_K/Q6_K/Q8_0/Q4_0 (issues #156/#201/#204/#205), plus SoA weight repacking.

## Proposed work

Add cooperative-matrix GLSL shader variants for the hot matvec/GEMM paths and dispatch them when `HasCooperativeMatrix` is true (with a scalar fallback).

## Impact

Vulkan leaves significant matmul throughput unrealized on hardware that supports cooperative matrices.

pekkah:
## Investigated + measured: implemented coop-matrix Sgemm, but it doesn't pay off (bandwidth-bound). Closing not-planned.

I implemented a full `VK_KHR_cooperative_matrix` (tensor-core) fast path for `VulkanBackend.Sgemm` and benchmarked it end-to-end. Summary: **it works and engages, but gives no measurable speedup** — the Vulkan workloads that use Sgemm are bandwidth-bound, not FLOP-bound, so tensor cores can't help.

### What was built (and verified correct)
- `SgemmCoopMat` GLSL: subgroup-scoped, fp32→fp16 16×16 tile staging through shared memory, `coopMatMulAdd` with an **fp32 accumulator**, B-transpose handled (`C = A × B^T`, B stored `[N,K]`).
[...]

PR #341, opened 2026-06-22
feat(vulkan): batched-trunk BatchVerify for speculative decoding (#308 PR1c)

pekkah:
#308 PR1c — the payoff integration. Wires the merged batched matvecs (#339/#340) into a batched-trunk `BatchVerify`.

## Change
Re-implements `GpuForwardPass.BatchVerify` on a batched trunk that processes all K draft tokens in **one command buffer**, using `MatMulBatched` (Q4_K/Q6_K weight-stationary) for the weight-bound matmuls — each weight matrix read from VRAM once for all K tokens. The previous K-loop body is preserved as `BatchVerifyKLoop` (fallback + parity oracle).

- **`_canBatchedTrunk` gate**: dense (`!_isMoE`), `!_isGemma4`, **no QKV/output bias**, every trunk matmul weight ∈ {Q4_K, Q6_K} (so the `MatMulBatched` non-Q4K/Q6K fallback's recording-hazard is never hit). Qwen3-8B-Q4_K_M qualifies; Q8_0/bias/MoE/gemma4 use the unchanged K-loop. K==1 → single `Forward`.
- **Batched trunk**: matmuls (Q/K/V/O, gate/up/down, lm-head) via `MatMulBatched` over `[K][dim]` buffers; elementwise (residual add/copy, SiLuMul) whole-buffer; per-token reductions/position ops (RmsNorm, HeadNorm, RoPE@startPos+i, KvAppend@startPos+i, Attention seqLen=startPos+i+1 for causal-among-K) via gather/scatter to single-token scratch. One `BeginRecord`/`EndRecordAndSubmit`, one K×vocab download then host-split.

## Verification
- **BIT-EXACT**: Qwen3-8B-Q4_K_M `BatchVerify(k=4,6)` == k sequential `Forward`, **maxAbs = 0.0** (truly bit-identical); Q8_0 still passes via the K-loop fallback; truncate-commit rollback holds. Full `~Vulkan` suite **82/82**, no regression; review clean (no-regression + bit-exactness + gate verified).
- **PERF (honest micro-bench, the PR2 arbiter)**, batched vs K-loop: **~0.90–1.08×** across k=4..8 / 512..2048 ctx — essentially **break-even** in the spec regime. The matmul amortization (≈4× less weight traffic at k=4) is real but masked by per-token gather/scatter + dispatch overhead.

## Why land it (gate-off, no user change)
The **CLI spec gate is NOT flipped** — default behavior is unchanged. This lands the bit-exact batched-trunk **foundation** (like the #338 K-loop foundation). Unlike a bandwidth-bound dead-end (cf. #313 coop-matrix), the break-even here is from **optimizable overhead**: the next step is **batched cheap-ops** (RmsNorm/RoPE/HeadNorm over [K][dim] in one dispatch, eliminating the gather/scatter) and/or batched attention. PR2 flips the gate only once the micro-bench shows a clear win.

## Follow-ups
- Batched cheap-ops to remove the gather/scatter overhead (the main break-even cause) → re-measure.
- Bias-model coverage (a Q4_K Qwen2 like VibeThinker-Q4 would exercise the currently-gated-out bias path).
- Then PR2: flip the CLI gate for prompt-lookup spec + E2E greedy parity.

### DIFF
--- a/src/SharpInference.Engine/GpuForwardPass.cs
+++ b/src/SharpInference.Engine/GpuForwardPass.cs
@@ -1833,26 +1833,43 @@ public sealed unsafe class GpuForwardPass : IForwardPass
     private void EnsureBatchVerifyScratch(int k)
     {
         if (_bvK >= k) return;
-        if (_bvK > 0) FreeBatchVerifyScratch();
+        if (_bvK > 0) FreeBatchVerifyScratch(); // free the prior (fully-allocated) generation
 
         int qDim = _numHeads * _headDim;
         int kvDim = _numKvHeads * _headDim;
         int ffnDim = ComputeFfnScratchDim(_isMoE, _intermDim, _expertDim);
 
-        _hiddenK = _gpu.Allocate(TensorShape.D1((long)k * _embDim));
-        _residualK = _gpu.Allocate(TensorShape.D1((long)k * _embDim));
-        _normK = _gpu.Allocate(TensorShape.D1((long)k * _embDim));
-        _qK = _gpu.Allocate(TensorShape.D1((long)k * qDim));
-        _kK = _gpu.Allocate(TensorShape.D1((long)k * kvDim));
-        _vK = _gpu.Allocate(TensorShape.D1((long)k * kvDim));
-        _attnOutK = _gpu.Allocate(TensorShape.D1((long)k * qDim));
-        _ffnGateK = _gpu.Allocate(TensorShape.D1((long)k * ffnDim));
-        _ffnUpK = _gpu.Allocate(TensorShape.D1((long)k * ffnDim));
-        _logitsK = _gpu.Allocate(TensorShape.D1((long)k * _hp.VocabSize));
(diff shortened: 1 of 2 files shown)


pekkah:

pekkah:
Good catch — fixed. If `EnsureBatchVerifyScratch` threw mid-allocation (OOM on a later [K][dim] buffer), the earlier buffers leaked (`_bvK` stayed 0, so the `_bvK>0`-guarded frees in both the next call and `Dispose` skipped them). Now each allocation is tracked in a list and the partial set is freed in a catch; switched the host logits buffer to `int` length.

Note: `FreeBatchVerifyScratch` can't be made unconditionally-called — `VulkanBackend.Free` dereferences `tensor.Handle` and `Tensor` is a reference type, so an unallocated field is null and would NRE (the test caught exactly this on my first attempt). It stays guarded to the fully-allocated case; partial cleanup goes through the catch's tracked list. Spec parity 8/8 still pass.

PR #342, opened 2026-06-22
perf(vulkan): batched cheap-ops for spec-decode verify (#308 PR1c-opt)

pekkah:
#308 PR1c-opt — turns the break-even batched-trunk `BatchVerify` (#341) into a measurable win.

## Change
PR1c was break-even: the per-token RmsNorm/HeadNorm/RoPE used K-loop gather/scatter (3 dispatches × K per op), dominating each layer and cancelling the matmul weight-amortization. This adds **batched** variants that process all K rows of a `[K][dim]` buffer in **one dispatch** each.

- New shaders **`RmsNormBatched` / `HeadNormBatched` / `RoPEBatched` / `RoPENeoxBatched`** — each processes K independent rows (row r = token r), bit-identical to the single-row op on token r. RoPE uses **per-row position = basePos + row** (token r → startPos+r) — the load-bearing detail for parity.
- `BatchVerifyBatched` rewired: attn/ffn/output RmsNorm, Q/K HeadNorm, Q/K RoPE → single batched calls on the `[K][dim]` buffers (#157 QK-norm→RoPE order preserved). Attention + KvAppend stay the per-token K-loop (batched attention is a separate larger lever).

## Verification
- **BIT-EXACT preserved**: Qwen3-8B-Q4_K_M `BatchVerify == k sequential Forward`, maxAbs=0.0; **14 new model-free** batched-vs-single-row unit tests (k=1/4/6/8, HeadNorm shared+per-channel, RoPE interleaved+NEOX, per-row position) all bit-exact. Full `~Vulkan` suite **96/96**, no regression; review clean (per-row bit-exactness + the in-place-norm WAR-safety verified).
- **MICRO-BENCH** (Qwen3-8B-Q4_K_M, batched vs K-loop) — break-even → a measurable win growing with k/context:

  | prefill | k=4 | k=6 | k=8 |
  |---|---|---|---|
  | 512  | 0.96× | 1.05× | 1.09× |
  | 2048 | 1.05× | 1.14× | **1.18×** |

## Honest status / next
Not yet a *strong* win (>1.2×). The remaining cap is the per-token **KvAppend + Attention K-loop** + the q/k/v/attnOut gather — i.e. **batched attention** is the next, larger lever. The **CLI spec gate is still NOT flipped** (PR2 flips it only on a clear end-to-end win). This PR is a correct, bit-exact improvement that takes the foundation from break-even to a real (if modest) speedup.


pekkah:

pekkah:
Good catch — fixed. Added `!_hp.UseL2QkNorm` to `ComputeCanBatchedTrunk` so L2 QK-norm (HeadNormPure post-RoPE, Llama-4) is excluded explicitly rather than relying on the `Debug.Assert` (stripped in Release) + the indirect L2⟹Llama4⟹MoE coupling. A future non-MoE L2 model now falls back to the verified K-loop instead of silently skipping the norm. Spec parity 8/8 still pass.

PR #343, opened 2026-06-22
perf(vulkan): batched attention for spec-decode verify (#308)

pekkah:
#308 — batched attention, the big lever that turns the modest spec-verify speedup into a context-scaling win.

## Change
The batched-trunk verify still did attention as a per-token K-loop (gather q_i → KvAppend → Attention(seqLen=startPos+i+1) → scatter attnOut_i) — ~4K dispatches+barriers/layer, the dominant remaining overhead. Replace it (fp32 KV, startPos+k≤4096) with **one batched KvAppend + one batched attention dispatch** reading `_qK` / writing `_attnOutK` directly.

- **`Shaders.AttentionBatched`**: 2D grid numHeads × K queries; query qi uses **per-row `seq_len_i = base_pos+qi+1`** (causal-among-K: attends `[0, base_pos+qi]`, not the full batch), `-inf` tail pad on the per-row bound. Score/softmax/aggregate verbatim from the single-query fast path → each (h,qi) workgroup is an independent copy → **bit-identical** to K single-query Attention calls.
- **`Shaders.KvAppendBatched`**: append all k K/V in one 2D dispatch.
- **Rewire**: `fp32 && startPos+k≤4096` → batched path; else the existing per-token K-loop (bf16/q8_0 KV or >4096) **unchanged**.

## Verification
- **BIT-EXACT**: Qwen3-8B-Q4_K_M `BatchVerify == k sequential Forward`, maxAbs=0.0; new model-free unit tests `AttentionBatched == k× single-query` (GQA 8/2, **distinct per-query causal ranges**) and `KvAppendBatched == k× KvAppend` both bit-exact. Full `~Vulkan` **101/101**, no regression; review clean (per-row seq_len/tail-pad causal correctness + gate/fallback verified).
- **MICRO-BENCH** (batched vs K-loop), the win now scales with context:

  | prefill | k=4 | k=6 | k=8 |
  |---|---|---|---|
  | 512  | 1.01× | 1.12× | 1.16× |
  | 2048 | **1.17×** | **1.28×** | **1.34×** |

  Isolated A/B: batched attention shaves ~12-14% off the batched-trunk wall-clock at 2048.

## Status / next
The spec-decode **verify** is now a solid context-scaling win (up from the 1.05-1.18× band). CLI spec gate still **not flipped** — **PR2** wires PromptLookupDraft spec on Vulkan + measures the real end-to-end greedy-parity + speedup (the final arbiter for flipping the gate). bf16/q8_0 KV batched attention = follow-up (same shader + narrowed K/V read).


pekkah:

pekkah:
Good call — removed. Confirmed the `-inf` tail-pad of `scores[seq_len_i..4096)` is dead: Phase 2 (max scan / exp-sum / scale) and Phase 3 (V-aggregate) all iterate `[0, seq_len_i)`, so the padded slots are never read (the single-query shader padded because its scan structure differed). Dropped the pad loop + its barrier — saves up to ~4096 shared writes/workgroup on the hot path. Parity stays bit-exact (maxAbs=0.0); 17/17 tests pass; SPIR-V regenerated.

PR #345, opened 2026-06-22
perf(vulkan): bf16/q8_0 batched attention for spec verify (#308 follow-up)

pekkah:
#308 follow-up — extend the fp32 batched spec-verify path to **bf16 and q8_0 KV**, the narrowed dtypes that reach the long contexts where speculative decoding wins most (previously they fell back to the per-token attention K-loop).

## Change
- **4 new shaders**: `AttentionBatchedBf16`/`Q8_0` (the fp32 `AttentionBatched` skeleton — per-query `seq_len_i = base_pos+qi+1`, causal-among-K, no tail-pad — with K/V reads swapped to the proven bf16 `unpackHalf2x16` / q8_0 block-dequant idioms) and `KvAppendBatchedBf16`/`Q8_0` (K-row append; bf16 `packHalf2x16` word store; q8_0 amax→quant + masked-atomic byte store).
- **Gate**: `BatchVerifyBatched` switched from fp32-only to a switch over `{Float32, BFloat16, Q8_0}` at `startPos+k≤4096`, each dispatching its batched pair; `>4096`/other dtypes → the unchanged per-token K-loop fallback. fp32 path byte-identical.

## Verification
- **BIT-EXACT** (batched == k sequential Forward, same dtype): bf16 maxAbs=0.0, q8_0 maxAbs=0.0. The q8_0 masked-atomic batched append is byte-identical to k per-token appends — review **exhaustively verified** the cross-row disjoint-byte safety (zero collisions across all `(blocks_per_row, position, k)`, masked atomics order-independent). Full `~Vulkan` suite **108/108**, no regression; build clean; review clean.
- **MICRO-BENCH** (Qwen3-8B-Q4_K_M, prefill 3072, batched vs K-loop) — *exceeds* fp32 (these dtypes reach longer ctx):

  | k | bf16 | q8_0 |
  |---|------|------|
  | 4 | 1.32× | 1.26× |
  | 6 | 1.46× | 1.36× |
  | 8 | 1.53× | 1.41× |

Spec stays opt-in. With this, all three KV dtypes (fp32/bf16/q8_0) take the weight-amortized batched verify; `--draft-model` on Vulkan remains the open follow-up.


pekkah:

PR #344, opened 2026-06-22
feat(vulkan): enable prompt-lookup speculative decoding (#308 PR2)

pekkah:
#308 PR2 — the finale: wire `--draft-lookup` (prompt-lookup) speculative decoding on the Vulkan backend.

## Change
The `SpeculativeDecoder` + `PromptLookupDraft` are backend-agnostic; this extends the CLI spec gate to admit `GpuForwardPass { SupportsBatchVerify: true }` (the merged weight-amortized + batched-attention `BatchVerify`) for the lookup path. `--draft-model` on Vulkan (needs a 2nd GpuForwardPass + VRAM mgmt) is a CUDA-only follow-up — **guarded to warn + fall back** to normal generation (never hits the CUDA-specific cast). CUDA/CPU spec paths unchanged; opt-in (default off).

## Verification
- **LOSSLESS**: `VulkanSpecDecodeE2ETests` greedy-parity — spec-on tokens == spec-off greedy tokens **EXACTLY** (Qwen3-8B-Q4_K_M, 40 tokens, **75% acceptance** on a repetitive prompt — exercises the amortized `BatchVerifyBatched` path + accept/reject + `TruncateTo` rollback). 102/102 Vulkan + 51/51 Cli, build clean; review clean (CUDA/CPU no-regression + the Vulkan `--draft-model` guard verified on the disjoint sealed-type hierarchy).

## Honest E2E speedup (the arbiter)
The end-to-end spec speedup is **acceptance- and context-gated** (prompt-lookup's nature, same as CUDA):
- **Win** on high-acceptance / long-context workloads — repetitive, code, copy/edit, structured extraction (output echoes prompt n-grams), riding the verify step's **1.17-1.34×** weight-amortized speedup (at prefill 2048).
- **~baseline-to-slight-slowdown** on low-acceptance *novel* generation at short context (measured Qwen3-8B, novel/thinking output, short ctx: ~8% acceptance → 44.5 vs 50.2 t/s).

The verify STEP is the #308 engineering win; the net E2E benefit scales with prompt-lookup acceptance. `--draft-lookup` is the right tool for lookup-friendly workloads.

## #308 arc complete (7 PRs)
[...]

PR #346, opened 2026-06-22
fix(vulkan): exact-size batched-verify scratch for variable k (#308 correctness)

pekkah:
**Shipped correctness fix** for the Vulkan spec-decode path (#341/#344).

## The bug
`EnsureBatchVerifyScratch` sized the batched-verify scratch **grow-only** (`if (_bvK >= k) return;`). But `BatchVerifyBatched` passes the **whole** `[k][dim]` buffers to `MatMulBatched`, which derives `rows/cols = ElementCount/nTok`. When a later step has a **smaller k** than a prior one — which happens **every generation** (k shrinks to 2/3 at the tail as `remaining < lookahead`, and on partial prompt-lookup matches) — the oversized buffer made rows/cols wrong:
- k doesn't divide the oversized count → `ArgumentException` (**hard crash**), or
- k divides evenly → wrong slices → **garbage logits, argmax flips, divergent output** (measured maxAbs ~36 vs the K-loop oracle).

So the shipped `--draft-lookup` (#344) **crashes** at some decode lengths (tail k=3) and **silently diverges** at others (tail k=2). The 40-token e2e test passed only because that acceptance pattern never landed the tail on k∈{2,3}.

## Fix
Exact sizing — reallocate on **any** size change (`_bvK == k`), mirroring CUDA's `EnsureDecodeLogits` `== n` (CUDA was never affected). Reallocs occur only when k actually changes (k is the constant lookahead in steady state; shrinks only at the tail).

## Diagnosis (decisive, evidence-based)
- **K-loop vs batched under the same churn**: forcing `BatchVerifyKLoop` is lossless (firstDiv=-1 at 40 & 120 tokens) while the batched path crashes/diverges → the bug is in `BatchVerifyBatched`, not the (backend-agnostic) decoder/churn.
- **Churned-KV vs fresh-Prefill** byte compare at committed positions: maxAbs=**0.0** → the truncate+re-append churn is byte-perfect (refutes the "KV-churn corrupts the cache" hypothesis).
- Affects fp32 and q8_0 identically (upstream of KV dtype). CUDA unaffected.

## Verification
- New regression test `Qwen3_8B_Q4K_BatchVerify_VariableK_MatchesSequentialForward` drives **k = 4→2→3→6→1 on ONE instance** vs the K-loop oracle (crashes/diverges pre-fix, passes post-fix) — the gap the fixed-k tests missed (each used a fresh instance + single k).
- Full Vulkan spec suite **16/16**; build clean. This also unblocks `--draft-model` on Vulkan (lossless once the scratch is exact).

### DIFF
--- a/src/SharpInference.Engine/GpuForwardPass.cs
+++ b/src/SharpInference.Engine/GpuForwardPass.cs
@@ -1848,8 +1848,14 @@ public sealed unsafe class GpuForwardPass : IForwardPass
     /// </summary>
     private void EnsureBatchVerifyScratch(int k)
     {
-        if (_bvK >= k) return;
-        if (_bvK > 0) FreeBatchVerifyScratch(); // free the prior (fully-allocated) generation
+        // EXACT sizing (not grow-only): BatchVerifyBatched passes the WHOLE [k][dim] buffers to
+        // MatMulBatched, which derives rows/cols = ElementCount/nTok. A grow-only guard (>=) would
+        // reuse an OVERSIZED buffer when a later step has smaller k (k shrinks to 2/3 at the
+        // generation tail, or on a partial prompt-lookup match) → wrong rows/cols → garbage logits
+        // (or an ArgumentException if k doesn't divide the oversized count). Mirror CUDA's exact
+        // `_decodeLogitsCapacity == n` sizing. (Was `>=` — a latent crash/divergence in #308.)
+        if (_bvK == k) return;
+        if (_bvK > 0) FreeBatchVerifyScratch(); // free the prior generation (any size)
 
         int qDim = _numHeads * _headDim;
         int kvDim = _numKvHeads * _headDim;
--- a/tests/SharpInference.Tests.ForwardPass/VulkanSpecBatchVerifyTests.cs
+++ b/tests/SharpInference.Tests.ForwardPass/VulkanSpecBatchVerifyTests.cs
@@ -250,6 +250,61 @@ public sealed class VulkanSpecBatchVerifyTests
     }
 
     /// <summary>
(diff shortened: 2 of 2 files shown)


pekkah:

pekkah:
Thanks — considered the grow-only + exact-view approach, but I'm keeping exact sizing here, for a reason specific to this path:

1. **It keeps the dominant transfer minimal.** The biggest per-step cost in `BatchVerifyBatched` is the logits download — `k × vocab × 4B` (vocab 151936 → ~0.6 MB/token, up to ~4.6 MB at k=8). With exact sizing `_logitsKBuf` is `k×vocab`, so the download is exactly k-sized. A grow-only buffer would over-download `max_k×vocab` on **every** batched step (e.g. _bvK=4, k=2 → 2× the download) unless the views are also paired with a k-sized download path — i.e. the grow-only approach needs *more* code to avoid making the dominant cost worse.

2. **Reallocs are infrequent, not per-step.** k=1 (no n-gram match) takes the single-token `Forward` path (no batched scratch at all); full-lookahead matches reuse the existing buffer (`_bvK == k`, no realloc). Only partial-match steps (k∈{2,3}) and the generation tail trigger a realloc — and the scratch buffers are small relative to the vocab-sized logits transfer the exact path saves.

3. Mirrors CUDA's `EnsureDecodeLogits` `== n` exact sizing (which is why CUDA was never affected).

So the realloc Gemini flags is bounded and cheaper than the over-download it would otherwise introduce. If profiling later shows the realloc dominates, a grow-only + exact-view + k-sized-download refinement is a clean follow-up — but it's strictly more code for this path, not less.

PR #347, opened 2026-06-22
perf(vulkan): int8-DP4A batched Q4_K matvec — Vulkan spec decode now wins (#308)

pekkah:
**The #308 breakthrough: Vulkan spec decode now delivers a real E2E speedup.**

A diagnostic proved the FP batched verify matvec (`MatVecBatchedQ4K`) didn't amortize weights (k=4 ≈ 2.0-2.2× a single matvec → verify ≈ 3.4× a decode → spec ~2× SLOWER despite good acceptance). This rebuilds it around **int8-activation quantization + the `GL_EXT_integer_dot_product` DP4A intrinsic**, mirroring CUDA's WS kernel — the per-token cost collapses from 8 FP32 loads+FMAs/weight-word to ~2 packed-int loads + 2 dp4a.

## Change
- **`Shaders.QuantizeQ8_1`**: FP32 → Q8_1 (per-32 amax→int8 + `{fp16 d, fp16 d·Σq}`). `subgroupMax/Add`, uniform-flow `subgroupShuffle` pack.
- **`Shaders.MatVecBatchedQ4KInt8`**: Q4_K 4-bit × int8 via `dotPacked4x8AccSatEXT`. Hoists token-invariant `super_d·sc`/`super_dmin·mn` once (weight read amortizes); per token `acc += (sd_sc·d8)·dp4a(wq,act) − (sm_mn·d8)·dp4a(0x01010101,act)`. Identity + layout verbatim from CUDA `llm_matvec_q4k_ws_n`.
- **Gate**: `Q4_K && HasShaderIntegerDotProduct` → int8 path; else FP fallback (try/catch). Q6_K + other dtypes unchanged (FP) — Q6_K int8 is the next phase.
- **`_q81BatchBuf`** scratch with **deferred free** (fixes a device-lost UAF on mid-recording grow).
- **Device**: chain `VkPhysicalDeviceShaderIntegerDotProductFeatures` (the intrinsic needs the feature enabled, not just the extension — VUID-06279).

## Correctness
Spec verify needs **argmax-stability**, not bit-exactness — int8 is lossy but argmax-stable, which is lossless for greedy spec (same contract as CUDA). Parity relaxed to exact-argmax + maxAbs<1.0 + top-5 overlap for the int8 path (FP fallback keeps bit-exact). Measured: BatchVerify vs FP K-loop **argmax-exact, maxAbs ~0.067** (k=4); **E2E greedy byte-identical** (lossless). 30/30 spec/matvec/precompiled tests pass; build clean; both shaders precompiled (Count 68); review clean (the int8 identity, the Q8_1 quant + uniform-flow shuffle, and the deferred-free UAF fix all verified).

## The win (Qwen3-8B-Q4_K_M, Vulkan, 4070 Ti)
- **Verify step** (int8 batched vs K-loop, prefill 2048): k4 **1.93×** / k6 **2.13×** / k8 **2.24×** faster (the FP batched path was break-even).
- **E2E `--draft-lookup`**, high acceptance (71%, `/no_think` repetition): **88.7 vs 50.7 t/s spec-off = 1.75× FASTER** — the FP path was break-even-to-slower.

This is the lever the FP batched verify lacked. Spec stays opt-in. **Follow-ons**: Q6_K int8 (P2, remaining trunk weights → even faster), `--draft-model` rewiring (now worth it).


pekkah:

pekkah:
Addressed:

1. **Q8_1 scratch leak on fallback** — fixed (de4c39d). The catch that disables the int8 path now frees `_q81BatchBuf` (a prior successful call may have allocated it), using the same deferred-free guard as the grow path: defer if mid-recording (a recorded-but-unsubmitted dispatch could still reference it — an immediate `Free` would risk the UAF the deferred mechanism fixes), else free immediately. Note: `EnsureQ81BatchBuf` runs *after* pipeline creation, so a pipeline-creation failure can't strand a fresh allocation — this only releases scratch from earlier successful calls.

2. **Wave64 subgroup robustness** — declined, with rationale: the new shaders are `local_size_x = 256` and rely on the #318 `requiredSubgroupSize = 32` pin, exactly as every other reduction shader in this codebase (`Attention`, `RmsNorm`, the existing `MatVecBatched*`, etc.) does. The pin forces 32-lane subgroups on Wave64 by construction, so `subgroupMax`/`subgroupAdd` operate over the intended 32-element sub-block. Introducing a one-off butterfly-`subgroupShuffle` reduction here would be inconsistent with the established pattern; a butterfly migration (if desired for pin-independent robustness) is a codebase-wide change, not specific to this PR.

PR #348, opened 2026-06-22
perf(vulkan): int8-DP4A batched Q6_K matvec — whole trunk amortizes (#308 P2)

pekkah:
**#308 P2** — extends the int8-DP4A batched matvec to **Q6_K** weights, so the whole Q4_K_M trunk amortizes weight reads across spec-draft tokens (P1 did Q4_K; ffn_down + token_embd/output are Q6_K and still used the FP path).

## Change
- **`Shaders.MatVecBatchedQ6KInt8`**: Q6_K 6-bit × int8 activation via `dotPacked4x8AccSatEXT`. int8 weight = `(q6−32) ∈ [-32,31]` (`q6 = ql_nibble | (qh_2bits<<4)`), packed 4/int; `acc += (d·scale[sub]·d8)·dp4a(w_int8, act)`. **No min-bias term** (Q6_K has no dmin, unlike Q4_K). Reuses the shared `QuantizeQ8_1` activations + `_q81BatchBuf`.
- **Gate**: `Q6_K && HasShaderIntegerDotProduct` → int8; else FP `MatVecBatchedQ6K` fallback. Q4_K int8 + other dtypes unchanged.

## Correctness
The `(q6−32)` decode is **bit-correct** vs the canonical Q6_K dequant (review verified 0 mismatches / 500 blocks × 256 cols; col↔activation pairing correct; no spurious Σq term). The int8 verify is **argmax-stable** (not bit-exact) — same contract as P1 and CUDA's int8 decode. Since the logit-producing output projection is now int8, the parity test is **margin-gated**: argmax asserted only when the FP top-2 margin exceeds 2×maxAbs (a mathematically-tight argmax-stability gate, + maxAbs<1.0 cap + top-5 overlap — review confirmed it's not a bug-hiding relaxation). BatchVerify vs FP K-loop maxAbs ~0.28-0.33. **E2E greedy byte-identical** to plain decode on Mars/story/code prompts (lossless in practice). 113/113 Vulkan, build clean, both shaders precompiled (Count 69); review clean.

## Win (Qwen3-8B-Q4_K_M, Vulkan, prefill 2048, verify vs K-loop)
| k | P1 (Q4_K int8) | **P2 (+ Q6_K int8)** |
|---|---|---|
| 4 | 1.93× | **2.18×** |
| 6 | 2.13× | **2.47×** |
| 8 | 2.24× | **2.62×** |

The Q6_K output projection (`[vocab×hidden]`) + ffn_down now amortize too. Combined with P1's 1.75× E2E `--draft-lookup` win, this pushes it further. Spec stays opt-in. **Follow-on**: P3 (`--draft-model` rewire, now worth it).


pekkah:

pekkah:
Good catch — fixed (b013241). The int8 fallback now **defers** pipeline disposal: `DisposeInt8PipelineDeferred` adds the pipeline to `_pendingPipelineFrees` when `_recording` (a prior int8 matmul may have recorded a dispatch referencing the SHARED `_quantizeQ8_1Pipeline`, so disposing mid-recording would UAF on submit), else disposes immediately. The deferred list is flushed in `FlushPendingScratchFrees` (after every submit, GPU idle) and in `Dispose` (after `vkDeviceWaitIdle`, since the fields are nulled and the per-field disposals would otherwise miss them). Same UAF-guard pattern as the existing Q8_1-scratch deferred-free. 27/27 spec/matvec tests pass.

Issue #351, opened 2026-06-22
feat(vulkan): full Gemma 4 support — PLE (E2B/E4B), shared-KV, narrowed-KV, partial-offload (beyond #309's PLE-free 12B)


pekkah:
Follow-up to #309 (Gemma-4 on Vulkan, closed) and #252 (Gemma-4 vision on partial-offload + Vulkan, open). #309 brought the gemma4 trunk to the Vulkan full-offload path, but scoped it to **PLE-free, shared-KV-free, fp32-KV, full-offload** models — i.e. the 12B-it-qat (`embedding_length_per_layer_input = 0`, `shared_kv_layers = 0`). This issue tracks closing the remaining gaps so the whole Gemma-4 family runs on Vulkan with parity to CUDA/CPU.

## What works today (#309 / #337 / #252)
`GpuForwardPass.ForwardGemma4` runs the full gemma4 trunk on Vulkan full-offload — per-layer head_dim, SWA + global, dual RoPE (1e6/1e4) + `rope_freqs`, qk-norm, pure V-norm, sandwich norms, `attn_scale=1.0`, final logit softcap, `k_eq_v` globals, raw `EmbedLookupQ6K` (tied embed) — **greedy byte-identical to CUDA**. Vision full-offload (`ForwardEmbedding`) ships too. Measured (RTX 4070 Ti): gemma-4-12B-qat-Q4_0 `-g -1 --backend vulkan -c 2048` = **17.0 prefill / 19.1 decode t/s** (vs CUDA 1714 / 54.1).

## Gaps (each rejected up-front in `GpuForwardPass` ctor, `GpuForwardPass.cs:300-317`, or absent)

- [x] **PLE (per-layer token embeddings)** — `HasPerLayerTokenEmbd` (`embedding_length_per_layer_input > 0`) throws \`Gemma 4 PLE ... is not implemented on the Vulkan backend\`. This blocks **all E2B/E4B** variants on Vulkan (both `gemma-4-E4B_q4_0-it` and `gemma-4-E4B-it-Q8_0` fail immediately). Reference: the CPU/CUDA \`BuildPerLayerProjections\` + PLE-256 injection. The biggest gap — E-series is the common small gemma4.
- [x] **Shared-KV layers** — `shared_kv_layers > 0` (`KvSourceLayer[i] >= 0`) throws. Needed for variants whose tail layers alias an earlier layer's K/V. Reference: CUDA `RunGemma4DeviceRegion` KV-share alias.
[...]

PR #353, opened 2026-06-22
feat(vulkan): Gemma 4 PLE + shared-KV on full-offload (#351 P1)

pekkah:
## What

Phase 1 of #351: closes the two construction-time rejections that blocked the entire **E2B/E4B gemma4 family** on the Vulkan backend — **per-layer token embeddings (PLE)** and **shared-KV tail layers**. #309 brought the gemma4 trunk to Vulkan full-offload but scoped it to PLE-free / shared-KV-free / fp32-KV (the 12B-qat). The real E4B GGUF has **both** `embedding_length_per_layer_input=256` **and** `shared_kv_layers=18`, so PLE alone wouldn't make it run — this wires both together so E4B runs end-to-end on `--backend vulkan`.

## PLE
Mirrors CPU `ForwardPass.BuildPerLayerProjections`/`ApplyPerLayerEmbedding` and CUDA `BuildPerLayerProjectionsGpu`/`ApplyPerLayerEmbeddingGpu`:
- The big `per_layer_token_embd` table (~2–4 GB) stays **CPU-resident** (mmap); one row is gathered+dequant'd per token and uploaded to `_gpuPleRow` **before** `BeginRecord` (`UploadToExisting` owns its own command buffer), made visible via `RecordTransferBarrier`. `model_proj`/`proj_norm` + per-layer `inp_gate`/`proj`/`post_norm` are GPU-resident via `UploadWeight`.
- `BuildPerLayerProjectionsGpu`: scale row by √pleWidth → `model_proj` matmul → 1/√embDim → per-row `RmsNormBatched` + add row + 1/√2. `ApplyPerLayerEmbedding` injects `gelu_tanh(inp_gate@hidden)*slice → proj → post_norm → add`, **after the post-FFN residual, before `layer_output_scale`**.
- **No new shaders** — all primitives already exist on Vulkan. Vulkan has no offset views, so the per-layer proj slice is extracted via `RecordComputeCopyRegion` before `GeluTanhMul`.
- `ForwardEmbedding` (vision seam) builds PLE from the **padding token 0**, mirroring CUDA.

## shared-KV
Mirrors CUDA `kvShared`/`effLayer`: `KvSourceLayer[i] >= 0` layers skip their own K/V projection + K-norm + V-norm + K-RoPE + KvAppend and don't upload `attn_k`/`attn_v`/`attn_k_norm`; they **alias** the source layer's K/V cache handles (tracked in `_kvAliasedLayers` so `Dispose` skips the double-free) and run attention against the source (`effLayer`) cache with their own SWA window. Q + Q-norm always run. For E4B: layers 0–23 own KV, 24–41 alias layer 22 (SWA) / 23 (global).

## Verification
RTX 4070 Ti, `gemma-4-E4B_q4_0-it.gguf`, `-g -1 --backend vulkan -c 2048` vs `--backend cuda`:
- **Byte-identical greedy** on a 96-token prompt (400/400 bytes, 0 diff).
- **Argmax-stable** on another (~40 tokens identical, then a sentence-boundary divergence; coherent, no degeneration) — within the gemma4 cross-backend contract.

The first decoded token alone fully exercises PLE injection at all 42 layers + the shared-KV tail during prefill+decode, so a structural bug is ruled out. Adds a model-gated, silent-skip CPU-oracle parity test (`Gemma4VulkanPleE2ETests`). Build is clean (TreatWarningsAsErrors, 0 warnings).

## Scope / out of scope
Narrowed-KV (bf16/q8_0) and partial-offload (`HybridForwardPass`, the remaining #252 Vulkan item) stay later #351 phases and are still rejected at construction.

### DIFF
--- a/src/SharpInference.Engine/GpuForwardPass.cs
+++ b/src/SharpInference.Engine/GpuForwardPass.cs
@@ -1079,6 +1079,14 @@ public sealed unsafe class GpuForwardPass : IForwardPass
     /// </summary>
     public ReadOnlySpan<float> Forward(int token, int position)
     {
+        // The GPU embedding gather shaders have no bounds check, and the Gemma 4 PLE row gather
+        // (BuildPerLayerRowUpload) reads the mmap'd table at token*bytesPerRow on the HOST — an
+        // out-of-range token id would OOB-read (a device fault or a host access violation). Guard
+        // at the entry, mirroring CudaForwardPass.Forward. Indicates a tokenizer/model mismatch.
+        if ((uint)token >= (uint)_hp.VocabSize)
+            throw new ArgumentOutOfRangeException(nameof(token), token,
+                $"Token id is outside the model vocabulary (0..{_hp.VocabSize - 1}).");
+
         if (_isGemma4)
             return ForwardGemma4(token, position);
 
--- a/src/SharpInference.Engine/GpuForwardPass.cs
+++ b/src/SharpInference.Engine/GpuForwardPass.cs
@@ -272,9 +272,9 @@ public sealed unsafe class GpuForwardPass : IForwardPass
     // The whole gemma4 trunk is dispatched from ForwardGemma4 when _isGemma4. The
     // per-layer head_dim varies (256 SWA / 512 global), so _maxHeadDim sizes the
     // Q/K/V/attnOut scratch for the widest layer and per-layer Tensor VIEWS carve
-    // out the active rows. This gemma4-v2 variant has PLE disabled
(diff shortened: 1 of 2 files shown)


pekkah:

pekkah:
Addressed in 21d0e4f. Added the guard at the `Forward` entry rather than inside `BuildPerLayerRowUpload`, mirroring `CudaForwardPass.Forward` — this covers both the new host-side PLE row gather **and** the pre-existing unchecked GPU embed gather (`EmbedLookupQ6K`), which had the same latent OOB risk for an out-of-range token. `ForwardEmbedding` (vision) takes an embedding and uses the always-valid padding token 0, so it needs no guard.

PR #354, opened 2026-06-22
feat(vulkan): narrowed KV (bf16/q8_0) for Gemma 4 (#351 P2)

pekkah:
## What

Phase 2 of #351: wires `--kv-type bf16|q8_0` for the **gemma4 path** on Vulkan full-offload, giving gemma4 the long-context KV-narrowing it already has on CUDA. **No new shaders** — the Vulkan bf16/q8_0 KV append + attention shaders (#311/#325) already accept per-layer `head_dim`/`numKvHeads`/SWA `window`, so this is pure dispatch wiring.

## Changes (`GpuForwardPass.cs`)
- **ctor**: drop the gemma4 fp32-only KV reject (TurboQuant reject kept).
- **KV alloc**: the uniform bf16/q8_0 branches are now gated `&& !_isGemma4`, so a gemma4 model falls through to the per-layer `_isGemma4` branch — which allocates each KV-owning layer in the active dtype using its **own per-layer kvDim** (bf16 fp16-packed; q8_0 `block_q8_0` raw-word; else fp32). Shared-KV tail layers alias the source handle regardless of dtype (unchanged from P1).
- **`RunGemma4Layers`**: KV append (step f, `!kvShared`) and attention (step h, reads `effLayer` cache) dtype-branch `KvAppendBf16`/`AttentionBf16`, `KvAppendQ8_0`/`AttentionQ8_0`, else fp32 — same per-layer dims + SWA window, no split-KV.

For E4B the per-layer kvDim is 512 (SWA, head_dim 256) / 1024 (global, 512) — both even (bf16) and multiples of 32 (q8_0), satisfied by the existing ctor guards on the model max dims.

## Verification
RTX 4070 Ti, `gemma-4-E4B_q4_0-it.gguf`, `-g -1 --backend vulkan -c 2048`: greedy output with `--kv-type bf16` and `--kv-type q8_0` is **byte-identical to `--kv-type fp32`** over 64 tokens — within (and stronger than) the bf16/q8_0 argmax-stable contract. This exercises the per-layer narrowed alloc + dtype-branched append/attention + shared-KV tail aliasing under a narrowed store. Adds a model-gated narrowed-KV-vs-fp32 parity test. Build clean (TreatWarningsAsErrors, 0 warnings).

## Scope
Builds on #353 (P1 PLE + shared-KV). Remaining #351 phase: partial-offload gemma4 trunk in `HybridForwardPass` (the #252 Vulkan item), still rejected/absent.

### DIFF
--- a/src/SharpInference.Engine/GpuForwardPass.cs
+++ b/src/SharpInference.Engine/GpuForwardPass.cs
@@ -581,8 +581,10 @@ public sealed unsafe class GpuForwardPass : IForwardPass
             // dtype using its OWN per-layer kvDim (bf16 fp16-packed, q8_0 block_q8_0, else fp32).
             // For E4B the per-layer kvDim is 512 (SWA, head_dim 256) or 1024 (global, head_dim
             // 512) — both even (bf16 ✓) and multiples of 32 (q8_0 blocks align to row boundaries
-            // ✓), already enforced on the model max dims by the ctor guards above. Aliased layers
-            // copy the source handle regardless of dtype.
+            // ✓). The ctor guards above validate the scalar model max dims (hp.HeadDim); the
+            // per-layer SWA/global dims are validated explicitly below so a future gemma variant
+            // with an odd SWA head_dim or a per-layer kvDim not divisible by 32 fails loud rather
+            // than mis-storing. Aliased layers copy the source handle regardless of dtype.
             for (int i = 0; i < hp.NumLayers; i++)
             {
                 int kvSrc = hp.KvSourceLayer is { } ksl ? ksl[i] : -1;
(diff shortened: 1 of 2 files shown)


pekkah:

pekkah:
Both addressed in 8efb072:
- **code-reviewer (line 584 guard accuracy):** made the comment true — added explicit per-layer fail-loud guards in the gemma4 narrowed-KV alloc (bf16 even per-layer head_dim; q8_0 per-layer kvDim % 32 == 0), so a future gemma variant with an odd SWA head_dim or non-32-aligned per-layer kvDim fails loud instead of mis-storing. E4B (256/512, 512/1024) passes unchanged.
- **gemini (test distinct-count):** simplified to `new HashSet<int>(decoded).Count`.

No forward-pass numeric change (guards only throw on invalid models); the byte-identical fp32/bf16/q8_0 E4B verification still holds. Build clean.

Issue #356, opened 2026-06-22
feat(vulkan): Gated-DeltaNet (GDN) hybrid forward pass — qwen35moe/qwen36 family on Vulkan


pekkah:
The **Gated-DeltaNet (GDN) hybrid** architecture (`qwen35moe` / `qwen36`: GDN recurrent layers + sparse attention + MoE-or-dense FFN) has **no Vulkan forward pass**, so the entire Qwen3.6 family is rejected up front on `--backend vulkan`:

- `src/SharpInference.Cli/RunCommand.cs:705` — `"Hybrid GDN models (qwen35moe) are not supported on the Vulkan backend yet. Use --backend cuda or -g 0 (CPU)."`
- `src/SharpInference.Server/InferenceEngineLoader.cs:328` — same rejection for the server.

GDN is implemented only on **CPU** (`HybridGdnForwardPass`, `GdnKernels`) and **CUDA** (`CudaHybridGdnForwardPass`). There is no Vulkan GDN path at all (the only Vulkan "GDN" reference is a comment in `Shaders.cs` noting the SiLU formula). This is the umbrella tracking issue for bringing GDN to Vulkan.

## Affected models (all `qwen35moe`/`qwen36`, all blocked on Vulkan today)
These are the README benchmark rows with **no Vulkan column** because of this gap:
- **Qwen3.6-35B-A3B** (GDN + MoE) — `download-model.ps1 -Model qwen36-35b-a3b`
- **Qwen3.6-27B-MTP** (GDN, dense + MTP) — `qwen36-27b-mtp` / `-q5`
- **Qwen3.6-35B-A3B-MTP** (GDN + MoE + MTP) — `qwen36-35b-a3b-mtp`
- **Carnice** (Qwen3.6-35B-A3B-MTP finetune) — `carnice-35b-a3b-mtp`

(Note: these are 16–22 GB models that already run on Vulkan-*hybrid* in principle for the MoE/dense FFN — the blocker is purely the GDN recurrent layers + attention trunk, not VRAM. The MTP **speedup** for the `-MTP` variants is tracked separately — see the follow-up issue.)

## What needs building
1. **Vulkan GDN compute shaders** mirroring `src/SharpInference.Cpu/GdnKernels.cs` (and the CUDA kernels):
[...]

PR #358, opened 2026-06-22
feat(vulkan): GDN support shaders — conv1d/L2-norm/tile-heads (#356 PR1)

pekkah:
## PR1 of the Vulkan Gated-DeltaNet (GDN) port — #356

First of the #356 PR sequence (plan: [#356 comment](https://github.com/pekkah/SharpInference/issues/356#issuecomment-4771073609)). Adds the three GDN **support** compute shaders + `VulkanBackend` methods, mirroring the CUDA kernels and the CPU oracle op-for-op. **No forward pass, no inference yet** — those land in PR4.

### What's here
| Op | Shader | Mirrors |
|---|---|---|
| Depthwise causal conv1d (decode) | `GdnConv1dDecode` | CUDA `llm_gdn_conv1d_decode` / CPU `CausalDepthwiseConv1dDecode` |
| Per-head L2 norm (ggml convention) | `GdnL2NormPerHead` | CUDA `llm_gdn_l2_norm_per_head` / CPU `L2NormPerHead` |
| GQA head broadcast | `GdnTileHeads` | CUDA `llm_gdn_tile_heads` / CPU `TileHeads` |

Notes:
- **`GdnL2NormPerHead`** uses the ggml convention `scale = 1/max(√Σx², eps)` (eps floors the divisor) — distinct from `HeadNormPure`'s `1/√(mean+eps)`, so it's a new shader, not a reuse.
- **`GdnTileHeads`** is *tile* (`dst[h]=src[h % srcHeads]`), **not** torch `repeat_interleave` — matches the GQA K→V pairing `(0,16),(1,17),…`.
- CUDA wrappers use pointer arithmetic for sub-region offsets; Vulkan binds the whole buffer and passes the float-element offset as a **push constant**. C# method signatures match `CudaBackend` exactly so PR4's orchestration stays backend-symmetric.
- Precompiled SPIR-V regenerated (`gen-spirv.ps1`): **72 precompiled + 2 skipped (SgemmBf16/Fp8) = 74** consts; the three GDN shaders compile with bundled glslc → no runtime glslc fallback (AOT intact).

### Verification (GPU, RTX)
`VulkanShaderTests` (GPU-gated) compare each op to `GdnKernels`:
- `GdnConv1dDecodeMatchesCpu` — output **and** mutated state (channels=8192, kernel=4).
- `GdnL2NormPerHeadMatchesCpu` `[Theory]` — offset 0 and a non-zero sub-region; pre-offset bytes untouched.
- `GdnTileHeadsMatchesCpu` `[Theory]` — exact equality at offset (0,0) and (64,256); pre-dstOffset bytes untouched.

**8/8 pass** (incl. the 3 `VulkanPrecompiledShaderTests` drift checks), 612 ms. Clean Release build, 0 warnings under TreatWarningsAsErrors + AOT/trim analyzers.


pekkah:

pekkah:
Thanks @gemini-code-assist — addressed in 1de0875. The OOB you flagged (kernel_size 0 underflowing `retained` to ~4B, or >5 indexing past the fixed `float s_old[4]`) is now prevented at the shader's only entry point: `VulkanBackend.GdnConv1dDecode` throws `ArgumentOutOfRangeException` for kernelSize outside [1,5]. I chose the wrapper guard over in-shader clamping because clamping the read loop to `min(retained,4)` would still leave the final `state_data[(retained-1)*…]` write (and the `w_data[retained*…]` read) out of range for kernel_size>5, silently mis-computing rather than failing — and the shader is internal (never called except through this wrapper). qwen35moe/qwen36 use kernelSize 4. Added a Theory test asserting the throw for 0 and 6. (Same finding the code-reviewer raised.)

PR #359, opened 2026-06-22
feat(vulkan): GDN recurrence decode shader — core delta-rule scan (#356 PR2)

pekkah:
## PR2 of the Vulkan Gated-DeltaNet (GDN) port — #356

The core of the port: the single-token **GDN recurrence delta-rule scan** as a Vulkan compute shader + `VulkanBackend` method + GPU parity test. Mirrors CUDA `llm_gdn_recurrence_decode` and the CPU oracle `GdnKernels.GdnRecurrenceDecode` op-for-op. Builds on PR1's support shaders (#358). **Still no forward pass** — that's PR4.

### Algorithm (per v-head = one workgroup; one thread per output column j, `local_size_x=128`)
```
decay = exp(softplus(alpha+dt_bias) * ssm_a);   b = sigmoid(beta)
pass A:  S *= decay;            p[j] = Σ_i k[i]·S[i,j]
         d[j] = b·(v[j] - p[j])
pass B:  S[i,j] += k[i]·d[j];   o[j] = (1/√d)·Σ_i q[i]·S[i,j]   (fused)
o = RMSNorm(o)·norm_weight;     o *= SiLU(z)
```
State `[numVHeads, headDim, headDim]` row-major (i=key, j=value/out), updated in place. The `1/√d` readout scale is preserved (load-bearing — RMSNorm is scale-invariant except at its eps floor; omitting it causes per-head magnitude drift → gibberish decode).

### Vulkan specifics
- **`local_size_x = 128`, wrapper throws for `headDim != 128`.** Both target models use headDim=128. This makes all 128 invocations active → every `barrier()` reached by every thread (no divergent-barrier UB). A per-thread `j>=d return` before a barrier would be GLSL UB; we avoid it by construction.
- **Full-precision GLSL `exp`/`log`/`inversesqrt`** (vs CUDA's fast `__expf`/`__logf`) — so the Vulkan path tracks the CPU oracle *more* tightly than CUDA does.
- 11 storage-buffer bindings; C# signature matches `CudaBackend.GdnRecurrenceDecode` exactly (PR4 stays backend-symmetric).
- Precompiled SPIR-V regenerated: 73 precompiled + 2 skipped = 75 consts.

### Verification (GPU, RTX)
- `GdnRecurrenceDecodeMatchesCpu` (hv=32, d=128, seeded random, sane gate ranges): mutated **state** within `1e-4`, **output** within `2e-3` of `GdnKernels.GdnRecurrenceDecode`.
- `GdnRecurrenceDecodeRejectsNon128HeadDim`: asserts the guard.

**5/5 pass** (incl. precompiled drift checks), 547 ms. Clean Release build, 0 warnings under TreatWarningsAsErrors + AOT/trim.


pekkah:
Thanks @gemini-code-assist — both addressed in 8bb2fa9.

**Perf (high):** hardcoded `d`→128 in the shader — loop bounds `i < 128u`, indexing `i*128u` and `state_base = h*16384u`, and the `inversesqrt(128.0)` / `sRed[0]/128.0` scales. The shader was already specialized to headDim=128 (local_size_x, 128-wide shared arrays, wrapper guard), so this is consistent and lets the compiler strength-reduce `i*128`→`i<<7` and unroll. Mathematically identical at d=128 — the parity test passes byte-for-byte unchanged (state ≤1e-4, output ≤2e-3, 5/5). (Left `hv`/`d` in the push-constant block for struct stability/dims documentation; GLSL push-constant layout is fixed by declaration so the C# struct still matches.)

**Safety (medium):** added CPU-side size validation for all 11 tensors before dispatch (state ≥ Hv·D·D; q/k/v/z/output ≥ Hv·D; alpha/beta/ssmA/dtBias ≥ Hv; normWeight ≥ D), throwing `ArgumentException` so a too-small buffer fails loud on the CPU instead of an OOB GPU access.

Issue #357, opened 2026-06-22
feat(vulkan): MTP self-speculative decoding (NEXTN heads) — the speedup for the Qwen3.6 -MTP variants


pekkah:
Multi-Token Prediction (MTP / NEXTN self-speculative decoding) is **CUDA/CPU-only** (`MtpDecoder`, `MtpBatchTail`). The Qwen3.6 `-MTP` models get their headline decode speedup from it; on Vulkan they would run at MTP-off baseline (even once GDN lands — see below).

## Dependency
**Blocked by #356** (Gated-DeltaNet on Vulkan). All the MTP models are GDN-arch, so they cannot run on Vulkan *at all* until #356 lands. This issue is the **follow-up speedup**, not the unblocker — once #356 makes them run, this restores MTP self-spec.

## Affected models (the `-MTP` subset of #356's list)
- **Qwen3.6-27B-MTP** (dense GDN + MTP)
- **Qwen3.6-35B-A3B-MTP** (GDN + MoE + MTP)
- **Carnice** (Qwen3.6-35B-A3B-MTP finetune)

## What's reusable / what's needed
- The Vulkan spec-decode **infrastructure already exists**: #308 added `BatchVerify` / batched-attention / batched int8 matvecs to the Vulkan `GpuForwardPass` (used by `--draft-lookup`). MTP verify is the same shape (k-token batched verify), so the verify side is largely there.
- What's missing on Vulkan: the **MTP/NEXTN head forward** (the extra prediction heads that propose the draft tokens) wired into the Vulkan GDN hybrid path, plus `GdnStateCache` MTP rollback (snapshot/restore of the GDN recurrent state across a rejected speculative step — `GdnStateCache` is currently CUDA/CPU).
- Forward-pass + decoder selection so `--mtp` (CLI) / `SpecType` (server) admit the Vulkan GDN hybrid.

## References
- MTP: `MtpDecoder`, `MtpBatchTail`, `GdnStateCache` (rollback)
- Vulkan spec-decode infra to reuse: #308 (`BatchVerify` + batched attention on `GpuForwardPass`)
- CUDA MTP follow-ups for context on the moving parts: #288, #289, #291, #209
- Umbrella / unblocker: #356

## Verification
Greedy byte-identical to the same model's MTP-off Vulkan decode (MTP is lossless self-spec), with measured acceptance ≥ the CUDA figure on a repetitive prompt. Secondary to #356.

pekkah:
[...]

PR #362, opened 2026-06-22
feat(vulkan): batched + fused-scan GDN shaders for prefill (#356 PR5a)

pekkah:
## PR5a of the Vulkan Gated-DeltaNet (GDN) port — #356 — batched + fused-scan shaders

First of the **PR5 (prefill perf)** sequence. Adds the batched + fused-scan GDN shaders that **PR5b** will wire into a one-dispatch-per-stage batched prefill trunk. All mirror the CUDA `llm_gdn_*_batched` / `llm_gdn_recurrence_scan` kernels op-for-op and are **byte-identical to N sequential single-token calls** (same per-row/per-step arithmetic + reduction order; only host launch overhead removed).

| Shader | Purpose |
|---|---|
| `GdnConv1dDecodeBatched` + `GdnConv1dStateUpdateBatched` | depthwise conv1d over a chunk (read-only carried state, (X,nTok) grid) + separate in-place state advance |
| `GdnL2NormPerHeadBatched` | per-(head,token) ggml L2 norm, (numHeads,nTok) grid |
| `GdnTileHeadsBatched` | per-token GQA tile (not repeat-interleave), src/dst strides |
| `GdnRecurrenceScan` | **fused sequential scan** — one workgroup per v-head (local_size_x=128), loops nTok internally carrying per-head state, each position running the exact `GdnRecurrenceDecode` passes with a trailing barrier at the position boundary |

`GdnRecurrenceScan` includes the **#290/#357 ring-capture** (post-Pass-B state mirrored into a ring slot when `nCapture>0`; the always-present ring binding is a no-op placeholder when `ringScan` is null) — so the scan is MTP-verify-ready for #357 with no separate variant.

C# signatures match `CudaBackend` exactly. Precompiled SPIR-V regenerated: 81 + 2 skipped = 83 consts.

### Verification (GPU, RTX — self-parity vs the PR1/PR2 sequential ops, no CPU/model needed)
`GdnConv1dBatchedMatchesSequential` (output + final state), `GdnL2NormPerHeadBatchedMatchesSequential`, `GdnTileHeadsBatchedMatchesSequential`, `GdnRecurrenceScanMatchesSequentialDecode` (output + final state, byte-exact). **9/9 pass** (incl. precompiled drift). Clean Release build, 0 warnings (TreatWarningsAsErrors + AOT/trim).


pekkah:
Thanks @gemini-code-assist — addressed in the latest commit. Added positivity guards (`channels`/`nTok`/`numHeads`/`headDim`/`srcHeads`/`repeat` ≥ 1, and ring-capture params non-negative when `ringScan` is provided) across all five batched methods, throwing `ArgumentOutOfRangeException` so a negative value fails loud on the CPU rather than casting to a huge uint dispatch / OOB. Mirrors the existing `GdnConv1dDecode`/`SplitQG` guard pattern. 6/6 self-parity tests still pass.

PR #363, opened 2026-06-22
feat(vulkan): batched GDN prefill trunk — 1.5x prefill on 27B (#356 PR5b)

pekkah:
## PR5b of the Vulkan Gated-DeltaNet (GDN) port — #356 — batched prefill trunk (1.5× prefill)

Wires a **batched prefill trunk** into `VulkanHybridGdnForwardPass`: one dispatch per stage over a chunk of tokens instead of N sequential `Forward`s, mirroring CUDA's `GdnBlockBatched`/`AttnBlockBatched`. Uses PR5a's batched/fused-scan shaders. **Byte-identical to the existing sequential prefill** by composition.

### New (2 batched gated-attn shaders)
- `SplitQGBatched`, `RoPENeoxPartialBatched` (the existing `RoPEBatched` is full-dim; gated attn needs partial NEOX) + VulkanBackend methods (positivity-guarded), mirroring the CUDA batched kernels.

### Batched trunk
- `GpuGdnBlockBatched`: `MatMulBatched` qkv/z/α/β → `GdnConv1dDecodeBatched` → **(WAR barrier)** → `GdnConv1dStateUpdateBatched` → `SiLU` → `GdnL2NormPerHeadBatched` Q/K → `GdnTileHeadsBatched` → `GdnRecurrenceScan` → `MatMulBatched` ssm_out.
- `GpuAttnBlockBatched`: `MatMulBatched` QKV → `SplitQGBatched` → `HeadNormBatched` → `RoPEPartialBatched` → `KvAppendBatched` → `AttentionBatched` → `SigmoidMulInPlace` → `MatMulBatched` Wo.
- `MatMulBatched` is capped at 8 tokens (the #308 weight-stationary design), so the trunk runs in sub-chunks of ≤8 → ~8× weight-read amortization on the GPU trunk. FFN runs per-row (MoE/CPU-FFN batching out of scope — the trunk is the win). Gated to `startPos+N ≤ 4096` (the `AttentionBatched` range) with a sequential fallback beyond; kill-switch `SHARPI_VULKAN_BATCHED_PREFILL` (default ON).
- Also fixed a latent recording-mode hazard in `MatMulBatched`'s F32 fallback (the GDN α/β matvecs are its first recording-session user): `_recording`-guarded barriers + deferred temp frees. Immediate path byte-unchanged; **#308 spec-decode (Q4_K/Q6_K int8, a different path) unaffected** (regression-checked).

### Verification (GPU, RTX)
- `VulkanHybridGdn_BatchedPrefill_MatchesSequential`: batched and sequential greedy streams **byte-identical** on Qwen3.6-27B-MTP.
- All **#308 `VulkanSpecBatchVerify` tests pass (22/22, no regression)** from the shared `MatMulBatched` change.
- **Measured prefill, 27B/12 GB hybrid, 890-token prompt: 5.3 → 8.0 t/s = 1.51×**, byte-exact. (The GPU-trunk amortization translates to ~1.5× end-to-end even though much of the 27B's FFN is CPU-resident on a 12 GB card and stays per-token; the win is larger on cards where more of the trunk is GPU-resident.)

SPIR-V regenerated (85 consts). Clean Release build, 0 warnings (TreatWarningsAsErrors + AOT/trim).


pekkah:
Thanks @gemini-code-assist — addressed the safety finding; deferring the perf one with measurements.

**Fault latch (high) — DONE** (latest commit): mirrored `CudaHybridGdnForwardPass._faulted` — `PrefillBatched` latches `_faulted=true` for the whole non-transactional batched region and clears it only on consistent completion; `ThrowIfFaulted()` at the start of `Forward`/`Prefill` blocks any retry on corrupted recurrent state. Byte-exact parity test still passes.

**MatMulBatched F32-fallback temp churn (high) — deferring, with rationale.** The fallback only runs for the GDN F32 `ssm_alpha`/`ssm_beta` matvecs (rows=48): ~2 calls/GDN-layer × 2 temps = ~21k buffer alloc/free over an 890-token prefill, ≈ **0.4% of the ~111 s prefill** (measured 1.51× was achieved *with* this churn). The dominant qkv/ssm_out/ffn matmuls take the amortized Q4_K/Q6_K int8 path, not this fallback. A correct fix needs a *recording-safe* buffer pool (the temps can't simply be reused — multiple F32 `MatMulBatched` calls are recorded into one session before submit, so a shared temp would be overwritten before its recorded dispatch executes; that's why the current code allocates per-call + defers the free). Adding a pool to the shared `MatMulBatched` hot path (also used by #308 spec-decode) is higher-risk than the <0.5% it would save, so I'm tracking it as a follow-up rather than landing it here.

PR #364, opened 2026-06-23
feat(vulkan): FlashQLA chunked GDN prefill scan, opt-in (#356 PR5c)

pekkah:
## PR5c of the Vulkan Gated-DeltaNet (GDN) port — #356 — FlashQLA chunked prefill scan (opt-in)

Ports CUDA's `llm_gdn_chunked_prefill` (the chunk-parallel `chunk_gated_delta_rule`) to Vulkan as `GdnChunkedPrefill` — the last item of the agreed "all prefill perf" scope. One workgroup per v-head, `GDN_CHUNK=64`, thread j owns value-column j: per-chunk cumulative log-decay + lower-triangular K·K/K·Q matrices + forward-substitution for `u_t` + output (RMSNorm + SiLU(z)) + state carry, all in ~34 KB shared. Mirrors the CPU oracle `GdnRecurrenceChunkedPrefill`; **numerically equal to the sequential scan up to FP reduction order — argmax-stable, NOT byte-exact**.

- `VulkanBackend.GdnChunkedPrefill` + `MaxComputeSharedMemoryBytes` + `SupportsGdnChunkedPrefill` guard (the ~34 KB tile exceeds the 32 KB some GPUs guarantee → falls back to the byte-exact `GdnRecurrenceScan`). SPIR-V regenerated (86 consts; compiles with bundled glslc).
- **Wiring (opt-in, default OFF):** runs in place of the fused scan in the batched GDN block when `SHARPI_VULKAN_GDN_CHUNKED_PREFILL=1` AND the device fits the tile.

### ⚠️ Known limitation (user-agreed scope)
Meaningful chunked speedup needs 64-token chunks, but the batched trunk is capped at **8 tokens** by `MatMulBatched` (#308 `acc[8]`). This drop-in runs the chunked scan over the ≤8-token sub-chunk, so the **end-to-end win is small** until a "decoupled-64" rewiring (or a larger batched matvec) lands. The validated kernel ships now; the decoupling is a documented follow-up. (Decision: ship the kernel opt-in, defer the rewiring.)

### Verification (GPU, RTX)
- `GdnChunkedPrefillMatchesSequentialScan` (unit, N=64 full + N=20 partial): chunked vs the byte-exact `GdnRecurrenceScan` agree to **~2.5e-6** for L2-normalized-K inputs (the real `GdnL2NormPerHead` regime — un-normed random K is ill-conditioned/out-of-regime, CPU-reproduced + documented).
- `VulkanHybridGdn_ChunkedPrefill_MatchesFusedScanArgmax` (e2e): chunked vs fused greedy streams **argmax-identical** on Qwen3.6-27B-MTP.

**3/3 pass.** Clean Release build, 0 warnings (TreatWarningsAsErrors + AOT/trim). Shared-mem-guarded; silent-skips on <34 KB-shared GPUs.


pekkah:
Thanks @gemini-code-assist — addressed in the latest commit: `SHARPI_VULKAN_GDN_CHUNKED_PREFILL` is now read once into a `chunkedRequested` local and reused for both the flag and the unsupported-device warning. Pure refactor, identical behavior, clean build.

PR #365, opened 2026-06-23
feat(vulkan): GDN conv-state ring-capture for MTP rollback (#357 PR1)

pekkah:
## PR1 of the Vulkan MTP self-spec decode — #357

First of the #357 sequence. Adds the **GDN conv-state ring-capture** — the last piece of the GDN ring-capture op-kit (the scan-state ring already ships in PR5a's `GdnRecurrenceScan`). It's the snapshot mechanism MTP batched-verify (PR2) uses to roll back the GDN recurrent state across a rejected speculative step.

`GdnConv1dStateCaptureRing` mirrors CUDA `llm_gdn_conv1d_state_capture_ring` op-for-op: one launch writes every batched-verify ring slot's conv1d state; slot `i` receives the conv state the sequential decode loop would hold after token `i` — **byte-identical to `GdnConv1dStateUpdateBatched` with nTok=i+1** — reading the PRE-update state. C# signature matches the CUDA wrapper (`ringFloatOffset` carried as a push constant since Vulkan binds whole buffers). Positivity-guarded; 2D dispatch (channel × nCapture). SPIR-V regenerated (87 consts).

### Verification (GPU, RTX — self-parity, no model)
`GdnConv1dStateCaptureRingMatchesSequential`: each ring slot i == `GdnConv1dStateUpdateBatched(nTok=i+1)`, at offset 0 **and** a non-zero `ringFloatOffset` with interleaved multi-layer stride (the packed-ring layout); asserts bytes outside the written region stay untouched. **5/5 pass** (incl. precompiled drift). Clean Release build, 0 warnings (TreatWarningsAsErrors + AOT/trim).


pekkah:
Thanks @gemini-code-assist — addressed the range validation: added `ringSlotStride >= 0` and a `ringFloatOffset ∈ [0, uint.MaxValue]` guard (it's narrowed to a uint push constant) so an out-of-range offset/stride throws instead of silently wrapping. I left the tensor null-checks out to match the codebase convention — `Tensor` is a value type and `GetBuffer` validates the handle; the sibling GDN methods (GdnConv1dDecode/Batched etc.) don't null-check either. Pure validation, valid-input behavior unchanged.

PR #375, opened 2026-06-23
feat(engine): schema/grammar-constrained tool-call argument decoding (#374)

### LINKED ISSUE #374: feat(engine): schema/grammar-constrained decoding for tool-call arguments
## Summary

Add optional **schema/grammar-constrained decoding** so tool-call *arguments* are forced to conform to the supplied JSON Schema (and the model-native call syntax). Today we inject the tool schema into the prompt and parse whatever the model emits, but nothing constrains generation — a weak or prior-biased model can emit malformed/empty arguments and we faithfully pass them through.

## Motivation — measured on Gemma 4 12B QAT

Investigating tool-calling reliability on `gemma-4-12b-it-qat-q4_0` (stock Google QAT, dense `gemma4` path) surfaced a clean failure mode. The *envelope* is always correct (10/10 sampled calls parsed cleanly, `finish_reason=tool_calls`, zero format drift), but **arguments are wrong**, deterministically at `temp 0`:

| Prompt (temp 0) | Model output | |
|---|---|---|
| "Call get_weather with location set to Berlin" | `get_weather{location:"Berlin"}` | ✅ |
| "What is the weather in Berlin?" | `get_weather{}` | ❌ drops required `location` |
| "weather in Reykjavik" | `get_weather{}` | ❌ drops required `location` |
| "17 × 23, use calculate" | `calculate{expression:"17 * 23"}` | ✅ |
| "Search population of France" (schema param = `query: string`) | `web_search{queries:["..."]}` | ❌ invents `queries` **array** |

This was confirmed to be **model behavior, not an engine bug**:

- The rendered prompt contains the full, typed schema (`required:[location]`, `query:{type:STRING}`) — `JinjaChatTemplate` + the GGUF `format_argument`/`format_parameters` macros render correctly.
- The 12B QAT `chat_template` is byte-identical to the unsloth Q4_K_M build and matches E4B; `Gemma4ToolCallAdapter` parses exactly that format. `calculate`/explicit-`location` round-trip end-to-end, so schema injection + parsing are proven correct.
- The model substitutes **memorized tool-signature priors** (a no-arg weather call; a `web_search(queries=[...])` array) over the schema we hand it.

Constrained decoding is the standard mitigation for exactly this class of error: it makes the bad token sequences *unreachable* rather than hoping the model conforms.

## Proposal

When a request is tool-active and constraint is enabled, restrict the sampler to tokens permitted by a grammar derived from the active tool definitions:

1. **Grammar source.** Build a per-call grammar from the tool's JSON Schema (required keys must appear; enums restrict values; types restrict value shape). Arrays/objects/strings follow the schema.
2. **Model-native syntax.** The grammar must target the architecture's *wire format*, not generic JSON — e.g. Gemma 4's bespoke `<|tool_call>call:NAME{key:<|"|>val<|"|>,...}<tool_call|>` vs Qwen/Llama JSON. `IToolCallAdapter` is the natural place to expose a `BuildArgumentGrammar(tools)` hook so each adapter owns its syntax.
3. **Sampler integration.** A grammar/FSM mask applied in `Sampler` (Engine) that zeroes disallowed tokens before sampling; advances state per accepted token. Must stay NativeAOT-friendly (no codegen) and allocation-free on the hot path.
4. **Opt-in.** Gate behind a server option / env (e.g. `SHARPI_TOOL_GRAMMAR=1`, or honor OpenAI `tool_choice`/`response_format`-style hints). Default off → byte-identical to today.

Prior art: llama.cpp GBNF grammars, Outlines/`lm-format-enforcer`, vLLM guided decoding.

## Acceptance criteria

- [ ] With constraint on, `get_weather`-style required args are never omitted and `web_search` uses `query` (schema name), not `queries`, on the same probes above.
- [ ] Enum-typed params only emit declared values.
- [ ] Per-architecture grammar honors each adapter's native call syntax (Gemma `<|"|>` form; Qwen/Llama JSON).
- [ ] Default-off path is byte-identical; constrained path stays NativeAOT-compatible and allocation-free on the hot loop.
- [ ] Tests: grammar derivation from a schema + a decode-time conformance test on at least one real GGUF.

## Notes / open questions

- Scope: start with one architecture (Gemma 4) end-to-end, then generalize via the adapter hook.
- A weak model may still pick the *wrong tool* or wrong values — grammar fixes structure/names, not intent. Pair with model choice (e.g. the `gemma4-12b-agentic` finetune).

_Filed from a tool-calling reliability investigation; companion follow-up benchmarks the `gemma4-12b-agentic` finetune against the same probes._

pekkah:
Closes #374.

## What

Opt-in **schema/grammar-constrained decoding** that forces a tool call's *arguments* to satisfy the supplied JSON Schema, expressed in each architecture's **native call syntax**. Default **off** → byte-identical to today. Enabled via `SHARPI_TOOL_GRAMMAR=1` or `SharpInferenceServerOptions.ToolGrammar`. Scoped to **Gemma 4** end-to-end; the adapter hook generalizes to other families (follow-up for Qwen/Llama JSON).

## Why

On `gemma-4-12b-it-qat-q4_0` (temp 0) the call *envelope* is always correct but *arguments* are wrong deterministically — a **model prior**, not an engine bug:

| Prompt | before | after |
|---|---|---|
| "weather in Berlin" (`location` required) | `{}` | `{"location":"Berlin"}` |
| "population of France" (`query: string`) | `{"queries":["…"]}` | `{"query":"…"}` |
| enum `unit∈{celsius,fahrenheit}` | `{}` | `{"location":"Berlin","unit":"celsius"}` |
| `tags: array<string>` | — | `{"tags":["urgent","bug"]}` |

Constrained decoding makes the bad token sequences *unreachable* instead of hoping the model conforms.

## Approach

**Per-adapter grammar.** `IToolCallAdapter.BuildArgumentConstraint(tools, vocab)` (default `null`; Gemma4 override) so each family owns its wire format.

**The hard part — terminal→token mapping.** A byte-level pushdown automaton over Gemma's `<|tool_call>call:NAME{key:<|"|>val<|"|>,n:3}<tool_call|>` syntax. A candidate token is allowed iff replaying its bytes keeps the FSM alive. Matching is byte-level for the structural skeleton (so multi-token keys and **merged delimiters** like `:[` / `]}` work) and token-level for the `<|"|>` quote special token and free string content. The grammar enforces: only declared keys, every required key exactly once, value shape per declared type (strings quoted with free content; numbers/bool/null bare; arrays `[…]`; nested objects `{…}`), and `enum` restriction. It **engages the instant the tool name completes — before `{`** — so Gemma's merged `{}` empty-args token can't slip required arguments past the mask. Allocation-free hot path (reused mask/scratch buffers + first-byte pruning over the 262k vocab; warm constrained request ~0.33s, overhead negligible).

**Engine.** `SamplingParams.Constraint`; the `InferenceEngine` decode loop masks logits while constraining and advances the constraint per token (disables the device-side argmax fast path + MTP for constrained requests). `ContinuousBatchingEngine` warns once and ignores it (not yet wired — noted as follow-up).

**Server.** `ToolGrammar` option + `SHARPI_TOOL_GRAMMAR` env; OpenAI/Anthropic endpoints build the constraint from `req.Tools` on tool-active requests (OpenAI honors `tool_choice:"none"`).

## Verification (CUDA, gemma-4-12b-it-qat-q4_0, temp 0)

All failure cases above flip to correct; **plain text, already-valid calls, and `tool_choice:"none"` are byte-identical to default-off**; streaming honors the constraint.

## Tests

- `ToolSchemaParseTests` — schema derivation (types/required/enum/array/nested).
- `ToolGrammarMockTests` — decode-time conformance via a fake Gemma tokenizer (**CI-runnable, no GGUF**): required-key, foreign-key rejection, string/enum/number/array/nested-object value shapes, early-engage, reset.
- `ToolGrammarConstraintTests` — model-gated conformance against the real GGUF.

190 Tests.Core + 125 Tests.Server pass; full Release build clean under `TreatWarningsAsErrors` (trim/AOT analyzers + InvariantGlobalization on).

## Follow-ups (filed)

- #376 � Qwen/Llama JSON-syntax argument grammars via the same adapter hook.
- #377 � wire the constraint into `ContinuousBatchingEngine` (today it warns + ignores).
- #378 � constrain partially-typed tools (`Any`-typed value / open object currently left unconstrained).

### DIFF
--- a/README.md
+++ b/README.md
@@ -282,6 +282,20 @@ stream. `--no-thinking` disables it at the template level, `--hide-thinking` kee
 `--temp 0.6 --top-p 0.95 --top-k 20`. The server emits reasoning per protocol (Anthropic `thinking`,
 OpenAI `reasoning_content`).
 
+### Tool-call argument grammar (`SHARPI_TOOL_GRAMMAR`, issue #374)
+
+Opt-in constrained decoding that forces a tool call's **arguments** to satisfy the request's JSON Schema,
+expressed in the model's native call syntax. Gemma 4's instruct line emits the call *envelope* correctly
+but, at `temp 0`, substitutes memorized priors for the schema — `get_weather` drops the required `location`
+(emits `{}`), `web_search` invents a `{queries:[…]}` array instead of the schema's `query` string. With the
+flag on, a byte-level grammar over Gemma's `<|tool_call>call:NAME{…}` form masks the sampler so non-conforming
+tokens are unreachable: only declared keys, every required key once, value shapes per type, enums limited to
+the declared set (free string content stays unconstrained). On the same probes the arguments become
+`{"location":"Berlin"}` and `{"query":"…"}`. Enable on the server with `SHARPI_TOOL_GRAMMAR=1` (or
+`SharpInference:ToolGrammar`); default off is byte-identical to unconstrained decoding. Honored by the
+single-user engine (not yet continuous batching); other families (Qwen/Llama JSON) generate unconstrained
+for now via the same per-adapter hook.
+
 ### CLI examples
 
 ```bash
--- a/src/SharpInference.Core/Grammar/GemmaToolArgumentConstraint.cs
+++ b/src/SharpInference.Core/Grammar/GemmaToolArgumentConstraint.cs
@@ -1,0 +1,905 @@
+using System.Text;
+
+namespace SharpInference.Core.Grammar;
+
+// ── Compiled schema ───────────────────────────────────────────────────────────
+//
+// The parsed ToolSchema is "compiled" once per request into match tables (UTF-8 key/enum bytes,
+// required-key bitmask) so the per-token hot loop never re-encodes strings. Only tools whose schema
+// is FULLY constrainable (every value is a concrete type / typed array / nested typed object — no
+// Any-typed value, no open object) are compiled; a tool that isn't is simply left out of the
+// constraint and generates its arguments unconstrained, so the constraint never blocks generation.
+
+/// <summary>Compiled value-type descriptor (see <see cref="ToolSchemaNode"/>).</summary>
+internal sealed class CompiledNode
+{
+    public required JsonSchemaKind Kind { get; init; }
+    /// <summary>For an enum / boolean / null: the candidate literal byte sequences (≤64).</summary>
+    public byte[][]? Literals { get; init; }
+    /// <summary>True when <see cref="Literals"/> are matched inside the <c>&lt;|"|&gt;</c> quotes (string enum).</summary>
+    public bool QuotedLiterals { get; init; }
+    /// <summary>Element type for an array.</summary>
+    public CompiledNode? Items { get; init; }
+    /// <summary>Nested object shape.</summary>
+    public CompiledObject? Object { get; init; }
+    /// <summary>Integer (no fractional part) vs general number.</summary>
+    public bool IntegerOnly { get; init; }
+}
+
+/// <summary>Compiled object shape: per-property name bytes + value node, plus the required-key mask.</summary>
+internal sealed class CompiledObject
+{
+    public required byte[][] KeyBytes { get; init; }
+    public required CompiledNode[] Values { get; init; }
+    public required ulong RequiredMask { get; init; }
+    public int Count => KeyBytes.Length;
+}
+
+/// <summary>
+/// Constrained-decoding state machine (issue #374) for Gemma 4's native tool-call wire format:
+/// <c>&lt;|tool_call&gt;call:NAME{key:&lt;|"|&gt;val&lt;|"|&gt;,n:3}&lt;tool_call|&gt;</c>.
+///
+/// <para>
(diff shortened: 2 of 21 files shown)


pekkah:
Thanks @gemini-code-assist — addressed in ca28054:

- **TryCompileObject / TryCompileNode depth limit** (both high): added a `MaxDepth` cap on compilation. The schema *parser* was already depth-capped in a prior commit, but the `ToolSchema` records are public, so an in-memory deeply-nested schema would bypass that — good catch on the defense-in-depth.
- **String-open mask** (medium): now `Array.Fill(-inf)` + restore the quote logit, no branchy 262k loop.
- **SimulateToken save/restore** (medium): manual frame-copy loop instead of `Array.Copy` (depth is 1–3), avoiding per-call overhead on the masking hot path.
- **Free-string-content mask** (medium): already EOG-direct as of the prior review commit (e408688) — masks the handful of EOG ids rather than testing all 262k against the set.

22 grammar/schema tests pass; build clean under TreatWarningsAsErrors.

Issue #378, opened 2026-06-23
Tool-arg grammar: constrain partially-typed tools (Any/open-object values) (#374 follow-up)


pekkah:
## Summary

Constrain tool calls whose schema is only **partially** typed. Today (#374, PR #375) a tool is constrained only if its *entire* argument schema is fully constrainable; a single `Any`-typed value (no `type`), an open object (`additionalProperties:true` / no `properties`), or an untyped array drops the **whole tool** to unconstrained — including its required, well-typed keys.

## Current behavior

`GemmaToolArgumentConstraint.TryCompileObject` / `TryCompileNode` return `null` (skip the tool) when any value isn't fully constrainable. Safe (never wedges, never wrong) but coarse: a tool like `{location: string (required), context: object (open)}` gets no constraint at all, so `location` can still be dropped.

## Scope

Add a permissive "free value" frame to the automaton that accepts any single well-formed Gemma value (quoted string via the `<|"|>` token, bare scalar, `{...}`, `[...]`) and balances braces/brackets/quotes until the value ends — so the **structure** (required keys present, declared key names, the typed values) stays enforced while loosely-typed values are left free. Then compile partially-typed tools instead of skipping them.

Edge cases: brace/bracket/quote balancing for nested open objects; ensuring the free-value frame ends at the right delimiter so the enclosing object/array resumes correctly; keep allocation-free + default-off byte-identical.

## Tests

Add mock + model-gated cases: a tool mixing a required typed key with an `Any`-typed / open-object value still forces the typed/required parts.

Follow-up to #374 / #375.

Issue #377, opened 2026-06-23
Tool-arg grammar: wire constraint into ContinuousBatchingEngine (#374 follow-up)


pekkah:
## Summary

Wire grammar-constrained tool-argument decoding (#374, PR #375) into `ContinuousBatchingEngine` so it works under multi-user batching (`SHARPI_MAX_BATCH`).

## Current state

`SamplingParams.Constraint` is honored only by the single-user `InferenceEngine`. `ContinuousBatchingEngine` samples each batched sequence directly from its logits and **ignores** the constraint — it emits a one-time stderr warning ("tool-grammar constraint is ignored under continuous batching") rather than silently dropping it, but a server started with both `SHARPI_TOOL_GRAMMAR=1` and `SHARPI_MAX_BATCH=N` generates tool arguments unconstrained.

## Scope

- Carry a per-sequence `ITokenConstraint` on the batched sequence state (each `ActiveSeq` already owns its `SamplingParams`/`Rng`/decoder state).
- In the batched sample loop (`ContinuousBatchingEngine` ~L343-352 and the admission first-token path ~L687-689): when a sequence's constraint `IsConstraining`, mask that sequence's logits row via `Filter` before sampling, and `Accept` the chosen token for every sequence afterward.
- Each sequence has its own constraint instance (the constraint is stateful/single-request) — confirm `BuildArgumentConstraint` is called per request and the instance flows through admission.
- Remove the one-time warning once wired.

## Notes

The constraint is allocation-free per token and operates on a single logits row, so per-sequence masking composes with the batched layout (one `Filter`/`Accept` per active sequence per step). Keep default-off byte-identical.

Follow-up to #374 / #375.

PR #380, opened 2026-06-23
feat(core): JSON-syntax tool-argument grammar for Qwen/Llama/DeepSeek (#376)

### LINKED ISSUE #376: Tool-arg grammar: Qwen/Llama JSON-syntax constraints (#374 follow-up)
## Summary

Extend the schema/grammar-constrained tool-argument decoding from #374 (PR #375) to the Qwen and Llama families, which emit **JSON** tool-call arguments rather than Gemma's bespoke `<|"|>`-quoted form.

## Context

#374 added `IToolCallAdapter.BuildArgumentConstraint(tools, vocab)` (default `null`) plus a Gemma-specific byte-level pushdown automaton (`GemmaToolArgumentConstraint`). The hook is the generalization seam: each adapter owns its wire format. Gemma is wired end-to-end; everyone else returns `null` (generates unconstrained).

## Scope

- A `JsonToolArgumentConstraint` (reusable across `QwenToolCallAdapter` / `LlamaToolCallAdapter` / `QwenCoderToolCallAdapter` / `DeepSeekToolCallAdapter`) targeting standard JSON: `{"key": "value", ...}` with double-quoted keys/strings, bare numbers/booleans/null, `[...]` arrays, nested `{...}` objects, and `enum` restriction.
- Each adapter's `BuildArgumentConstraint` override returns it for its envelope (Qwen `<tool_call>{json}</tool_call>`, Qwen3-Coder `<function=NAME>` XML wrapper — note: Coder uses `<parameter=…>` XML, not JSON, so it may need its own matcher; Llama `<|python_tag|>{json}`, DeepSeek `name<|tool_sep|>{json}`).
- Reuse the existing `ToolSchema`/`CompiledObject`/`CompiledNode`/`GrammarVocabulary` infra and the same correctness invariants (only declared keys, required-once, value shapes, enum; free string content; default-off byte-identical; allocation-free hot path).
- The JSON form needs the early-engage trick too (the model can emit a merged `{}` token) and double-quote handling (JSON `"` vs Gemma's `<|"|>` special token — the structural quote is a regular byte here, so string content must be matched char-level with escape handling, or kept permissive between the quotes).

## Tests

Mirror `ToolGrammarMockTests` with a fake JSON tokenizer + model-gated conformance on a real Qwen/Llama GGUF (required-key, foreign-key rejection, enum, array, nested).

Follow-up to #374 / #375.

pekkah:
## Summary

Extends schema/grammar-constrained tool-call argument decoding (#374, PR #375) to the families that emit **standard JSON** — Qwen (`<tool_call>{json}</tool_call>`), Llama-3 (`<|python_tag|>{json}`), and DeepSeek (`NAME<|tool_sep|>{json}`) — the JSON sibling of `GemmaToolArgumentConstraint`. Closes #376.

## What changed

- **`JsonToolArgumentConstraint`** — a byte-level pushdown automaton over standard JSON: `{"key":"value",...}` with double-quoted keys/strings (`\`-escape aware), bare numbers/booleans/null, `[...]` arrays, nested `{...}` objects, and `enum` restriction. Matching is **fully byte-level** because a real BPE merges structural bytes with content/whitespace — one token carries `{"`, another `":`, another `"}}`, and the empty object is a single `{}` token (verified by probing the real Qwen vocab).
- **Early-engage at the argument key's colon** (`"arguments":` / `"parameters":`) pushes the object frame in `OExpectOpenBrace` so the *next* token is masked — a merged `{}` token can't drop the required arguments (the #374 failure mode, in JSON form).
- **Two envelopes**, parameterized per family:
  - `NameValueObject` (Qwen/Llama): a preamble parser walks `{"name":"X","argskey":{` to capture the tool name and find the args object; `argskey` is `arguments` (Qwen) or `parameters` (Llama).
  - `NameThenSeparator` (DeepSeek): bare name up to `<|tool_sep|>`, then the args object directly.
  Wired via `QwenToolCallAdapter` / `LlamaToolCallAdapter` / `DeepSeekToolCallAdapter`.`BuildArgumentConstraint`.
- **Shared compile logic extracted** — `CompiledObject`/`CompiledNode` + `TryCompile*` moved out of `GemmaToolArgumentConstraint` into `ToolSchemaCompiler`, used by both constraints (DRY, and sets up #378 to modify one place). The Gemma constraint now calls it; all 72 prior Gemma grammar/adapter tests stay green — behavior-preserving.
- **EOG forbidden inside the region** — empty-byte pruning plus an explicit EOG mask pass, so an EOS that decodes to ordinary text can't slip through a free string and truncate the call.
- **Default-off byte-identical** — inert when no tool is constrainable, when the open marker isn't a vocabulary token, or when the model emits a non-JSON shape (Qwen3.6's XML `<function=>` never engages).

## Tests

- **CI mock (`JsonToolGrammarMockTests`, 14)** — `FakeJsonTokenizer` provides single-byte tokens **plus realistic merged structural pieces** (`{"`, `":`, `{}`, …). Covers: required key can't close early, foreign-key rejection, enum, **string with `\"` escape**, number, nested object, array, the merged-`{}` early-engage, and the Qwen/Llama/DeepSeek envelopes.
- **Model-gated (`JsonToolGrammarConstraintTests`, 4)** — drives the constraint over a **real Qwen3.6 BPE vocabulary** (production merges: `{}`, `{"`, `":`), asserting it engages at the colon, rejects merged `{}`, allows `{"`, enforces declared keys, restricts the enum, and stays inert on the XML form.

`90 grammar/tool tests green`; full solution builds clean under `TreatWarningsAsErrors` + trim/AOT analyzers.

### Note on GPU e2e

A live generation A/B (like #377's) isn't included: the on-disk Qwen3.6 emits the **XML** tool shape (the JSON constraint correctly stays inert on it), and no JSON-emitting Qwen/Llama that fits VRAM was available. The model-gated test instead exercises the byte-level matching against the real production tokenizer.

### Follow-up

Qwen3-Coder's `<parameter=…>` **XML** wrapper isn't JSON and is left for a separate matcher (its adapter has no override → arguments generate unconstrained, never wrong) — the issue flagged this as possibly separate.

Follow-up to #374 / #375.

### DIFF
--- a/src/SharpInference.Core/Grammar/GemmaToolArgumentConstraint.cs
+++ b/src/SharpInference.Core/Grammar/GemmaToolArgumentConstraint.cs
@@ -2,39 +2,6 @@ using System.Text;
 
 namespace SharpInference.Core.Grammar;
 
-// ── Compiled schema ───────────────────────────────────────────────────────────
-//
-// The parsed ToolSchema is "compiled" once per request into match tables (UTF-8 key/enum bytes,
-// required-key bitmask) so the per-token hot loop never re-encodes strings. Only tools whose schema
-// is FULLY constrainable (every value is a concrete type / typed array / nested typed object — no
-// Any-typed value, no open object) are compiled; a tool that isn't is simply left out of the
-// constraint and generates its arguments unconstrained, so the constraint never blocks generation.
-
-/// <summary>Compiled value-type descriptor (see <see cref="ToolSchemaNode"/>).</summary>
-internal sealed class CompiledNode
-{
-    public required JsonSchemaKind Kind { get; init; }
-    /// <summary>For an enum / boolean / null: the candidate literal byte sequences (≤64).</summary>
-    public byte[][]? Literals { get; init; }
-    /// <summary>True when <see cref="Literals"/> are matched inside the <c>&lt;|"|&gt;</c> quotes (string enum).</summary>
-    public bool QuotedLiterals { get; init; }
-    /// <summary>Element type for an array.</summary>
-    public CompiledNode? Items { get; init; }
-    /// <summary>Nested object shape.</summary>
-    public CompiledObject? Object { get; init; }
-    /// <summary>Integer (no fractional part) vs general number.</summary>
-    public bool IntegerOnly { get; init; }
-}
-
-/// <summary>Compiled object shape: per-property name bytes + value node, plus the required-key mask.</summary>
-internal sealed class CompiledObject
-{
-    public required byte[][] KeyBytes { get; init; }
-    public required CompiledNode[] Values { get; init; }
-    public required ulong RequiredMask { get; init; }
-    public int Count => KeyBytes.Length;
-}
-
 /// <summary>
 /// Constrained-decoding state machine (issue #374) for Gemma 4's native tool-call wire format:
 /// <c>&lt;|tool_call&gt;call:NAME{key:&lt;|"|&gt;val&lt;|"|&gt;,n:3}&lt;tool_call|&gt;</c>.
@@ -100,7 +67,7 @@ public sealed class GemmaToolArgumentConstraint : ITokenConstraint
             foreach (var t in tools)
             {
                 if (t.Arguments.Open) continue;             // unconstrained body — skip
-                var compiled = TryCompileObject(t.Arguments);
+                var compiled = ToolSchemaCompiler.TryCompileObject(t.Arguments);
                 if (compiled is not null) _tools[t.Name] = compiled;
             }
         }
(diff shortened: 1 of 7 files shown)


pekkah:
Addressed the `ComputeMask` perf bottleneck in f5bd249: added the fast-path you suggested — inside `SContent` (and when not mid-escape), a token containing neither `"` nor `\` is pure content and is admitted without `SimulateToken`; only tokens that could close/escape get the full replay. This is the byte-level analogue of the Gemma sibling's token-level free-content shortcut. All 18 JSON tests stay green.

PR #382, opened 2026-06-24
feat(cli): --tools / --tool-grammar in sharpi run; canonical server ToolGrammar option

pekkah:
## What

Exposes schema/grammar-constrained tool-call decoding (issue #374) as **first-class CLI/server surface** instead of only the `SHARPI_TOOL_GRAMMAR` environment variable. Branches off master (independent of the #376/#377/#378 follow-ups).

## CLI — `sharpi run`

- **`--tools <file.json>`** — load OpenAI-format tool definitions (a bare array, or a `{ "tools": [ … ] }` wrapper). They're advertised to the model via its chat template (the `tools` variable, previously hardcoded `null`). On a single-prompt (`-p`) run the parsed tool calls are printed after generation.
- **`--tool-grammar`** — constrain the argument bytes to the supplied JSON Schemas via the model family's adapter (Gemma 4 today; Qwen/Llama/DeepSeek arrive with the #376 JSON constraint). Default off → byte-identical to unconstrained decoding.
- The shared `DecodeLoop` resets the constraint once per response, masks logits with `Filter()` while the grammar is restricting, and `Accept()`s every emitted token.
- **Tool-boundary stop** (issue #304): the adapter's markers (Gemma 4 `<|tool_response>`) are resolved against the vocab and added to the stop set, so generation halts right after the call instead of running into a hallucinated trailing turn.
- The speculative-decode path and the MTP fast path are guarded off when a constraint is active (a multi-token verify can't honor a per-token mask).
- JSON parsing and argument rendering use manual `JsonElement` / `Utf8JsonWriter` walks — no reflection-based serialization, **NativeAOT-clean**.

```
sharpi run -m gemma.gguf -g -1 --tools weather.json --tool-grammar \
  -p "What is the weather in Paris? Use the get_weather tool."
```

## Server

`ToolGrammarHelper.Enabled` now returns `opts.ToolGrammar` directly. The runnable host already maps `SHARPI_TOOL_GRAMMAR` onto that option at startup (`Program.cs`), so the env var still works while the library layer stops re-reading the environment — **the option is the single source of truth**.

## Invariants

- Default off → byte-identical (no `--tools`/`--tool-grammar`: `tools` template var stays `null`, `sp.Constraint` null, capture null, `DecodeLoop` unchanged).
- NativeAOT-clean, TreatWarningsAsErrors satisfied (0 warnings), InvariantGlobalization-safe (Ordinal).

## Verification

Build clean. `Tests.Cli` (51) + `Tests.Server` (125) green.

GPU-verified (`gemma-4-12b-it-qat-q4_0`, CUDA `-g -1`):

```
sharpi run --tools weather.json --tool-grammar -p "weather in Paris?"
→ Loaded 1 tool(s) from weather.json.
→ Tool-call arguments are grammar-constrained (issue #374).
→ <|tool_call>call:get_weather{location:<|"|>Paris<|"|>}<tool_call|>
→ Parsed 1 tool call(s): get_weather({"location":"Paris"})
```

The model halts cleanly at the tool-boundary (14 tokens, no trailing-turn noise) — vs. running to the 512-token cap emitting `<|tool_response>` repeats before the stop marker was added.

### DIFF
--- a/src/SharpInference.Cli/RunCommand.cs
+++ b/src/SharpInference.Cli/RunCommand.cs
@@ -823,6 +823,7 @@ public sealed class RunCommand : Command<RunCommand.Settings>
         List<ToolSchema>? toolSchemas = null;
         ITokenConstraint? toolConstraint = null;
         int[] toolBoundaryStops = [];
+        s_tools = null;   // reset: this run advertises tools only if --tools is given (no leak across in-process runs)
         if (settings.ToolsPath is { Length: > 0 } toolsPath)
         {
             if (!File.Exists(toolsPath))
@@ -834,7 +835,9 @@ public sealed class RunCommand : Command<RunCommand.Settings>
             {
                 (s_tools, toolSchemas) = LoadTools(toolsPath);
             }
-            catch (Exception ex)
+            catch (Exception ex) when (ex is IOException or UnauthorizedAccessException
+                                          or System.Security.SecurityException or NotSupportedException
+                                          or JsonException or FormatException)
             {
                 AnsiConsole.MarkupLine($"[red]Error:[/] could not parse --tools file: {Markup.Escape(ex.Message)}");
                 return 1;
--- a/src/SharpInference.Cli/RunCommand.cs
+++ b/src/SharpInference.Cli/RunCommand.cs
@@ -1,8 +1,11 @@
 using System.ComponentModel;
 using System.Diagnostics;
+using System.Text;
+using System.Text.Json;
 using Spectre.Console;
 using Spectre.Console.Cli;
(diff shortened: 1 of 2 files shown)


pekkah:
Addressed both review comments in 6d471e9:
- `s_tools` is now reset to `null` at the start of the tool-calling block so tools never leak across in-process runs.
- The `--tools` parse `catch` is narrowed to `IOException`/`UnauthorizedAccessException`/`SecurityException`/`NotSupportedException`/`JsonException`/`FormatException` so unrelated runtime faults aren't swallowed.

Build clean, Tests.Cli (51) green.

PR #404, opened 2026-06-25
perf(cuda): warp-per-column GDN decode kernel (#404, opt-in, ~6.5x)

pekkah:
## Summary

**Option B** for the unsloth Qwen3.6-35B-A3B decode gap (ours ~28–30 t/s vs llama.cpp tg128 49.9 on this same 60-SM card): a high-occupancy **warp-per-column GDN recurrence decode kernel** mirroring llama.cpp's `gated_delta_net_cuda`, for the **plain per-token decode** path.

The original `llm_gdn_recurrence_decode` launched only `hv`=32 blocks × 128 threads (≈half the SMs idle) and re-read the `[d×d]` state from global memory twice. The new kernel:
- Stores the per-head state **column-major (transposed)** so each warp owns one output column and reads its `d` rows **coalesced + register-resident** (read once).
- Warp-reduces the two key/query dot products; folds decay into the kv dot → single fused state write.
- Computes the per-head gate chain once on lane 0, broadcast across the warp.
- Splits the per-head RMSNorm + SiLU(z) gate into a separate launch (`llm_gdn_decode_norm_gate`) since columns now span blocks.

The state is flipped in place once at decode entry (`llm_gdn_state_transpose`, ~30 launches ≈200µs total, then amortized) and back before any row-major consumer.

## Measured (nsys, Qwen3.6-35B-A3B-MTP `--cpu-moe -g -1`, same card as llama.cpp)

| | recurrence kernel |
|---|---|
| baseline `llm_gdn_recurrence_decode` | **41.6µs** median (11.5% of all GPU kernel time) |
| fast `llm_gdn_recurrence_decode_fast` | **6.4µs** median / **3.5µs** warm-min = **~6.5×** |

The **3.5µs warm-min equals llama.cpp's same-hardware isolated number**. The 6.4µs in-context median is **DRAM-bound** (~4MB read+write of the 2MB/layer state at ~625 GB/s): the concurrent CPU-MoE expert streaming + attention/shared-expert kernels evict the state from L2 between GDN calls. `llama-bench`'s tight isolated loop keeps it L2-resident (1.14 TB/s → 3.5µs); interleaved with MoE, llama.cpp hits the same DRAM floor. We're at the physical in-context floor.

**Decode e2e is unchanged (~28 t/s)** — experts are RAM-bound (~58% of decode), GDN recurrence is only ~5%. The kernel reduction frees GPU time; the decode bottleneck is elsewhere. This is a building block for the broader trunk-coordination work.

## Gating (`SHARPI_GDN_DECODE_FAST=1`, default OFF)

- **Plain per-token decode ONLY** (`GpuGdnBlock` passes `allowFastDecode: true`). The MTP draft/verify 2-token path and the fallback sequential trunk keep the **byte-exact original** kernel.
- Engages only when **MTP is off** (`_gpuGdnRingScan is null`) and the decode CUDA-graph is off, so the transposed layout never coexists with the row-major scan / snapshot / ring consumers. Every state consumer (batched scan, snapshot up/download, ring capture/restore, reset, truncate) guards its layout via `EnsureGdnState{Transposed,RowMajor}`.

## Correctness

- **Argmax-stable, NOT byte-exact** (warp-reduced sums reorder the FP accumulation — same class as the chunked GDN prefill). Greedy decode stays coherent (diverges only at late near-ties).
- Tests: `GdnRecurrenceDecodeFast_ArgmaxStableVsCpuReference` (transposed round-trip vs CPU reference) + `GdnStateTranspose_IsInvolutionAndMatchesCpu` (byte-exact involution). All 11 GDN kernel tests pass.
- **MTP acceptance unaffected** — fast is bypassed under MTP (verified: ring allocates → original kernel path, coherent decode).
- The pre-existing-red Carnice batched-prefill bitwise oracle is **untouched** (fast default-off; the layout guards are `if(!flag) return;` no-ops when the flag is unset).

## Notes

Stacked on #403 (shares `CudaHybridGdnForwardPass.cs`); retarget to `master` once #403 merges.


pekkah:
Both HIGH comments addressed in 31ff6ea.

Both are latent for `head_dim != 128` only — the gated path runs exclusively at the qwen35moe value `head_dim = 128`, where both fixes are **byte-identical no-ops**, so the measured Option B results are unaffected.

- **Speculative OOB load** (`llm_gdn_recurrence_decode_fast`): the load index is now clamped to `0` for inactive register slots (`r >= rpl`), so the address is always in-bounds even if the compiler emits the predicated load. For d=128 every slot is active → `i` unchanged.
- **Non-power-of-two reduction** (`llm_gdn_decode_norm_gate`): the RMSNorm tree reduction now starts at `s=64` and masks the partner with `j + s < d`. For d=128 (power of two) this is byte-identical to the `s = d/2` halving; for d=96 it folds the upper tail instead of dropping it.

Parity tests (fast round-trip, transpose involution, original parity) stay green.

Issue #435, opened 2026-07-11
GPU Lloyd-Max TqAttention writes compressed-region V aggregate in the rotated basis (no inverse WHT)


pekkah:
# GPU Lloyd-Max TqAttention writes compressed-region V aggregate in the rotated basis (no inverse WHT)

## Summary

Found while investigating #432. Both GPU Lloyd-Max TurboQuant attention kernels aggregate the compressed-region V contribution in the **rotated/sign-flipped basis** and never apply the inverse transform:

- CUDA `llm_tq_attention` (`src/SharpInference.Cuda/CudaTextKernels.cs`, Phase 3): `acc += weight * cbook[idx] * norm` per output dim, then `out[out_off + d] = acc` — the TQ V codes are stored rotated (`llm_tq_kv_append` applies WHT + sign flip to both K and V), so this sum is a rotated-domain vector. It is added together with the **unrotated** FP32-window V contribution and written straight to the attention output, which feeds the Wo matmul.
- Vulkan `TqAttention` shader (`src/SharpInference.Vulkan/Shaders.cs`): identical structure, identical omission.

`CudaTurboQuantTests.TqAttention_NeedleInHaystack` documents this as a convention — "The output is in the rotated/sign-flipped basis (the V-side of TqAttention does not un-rotate — same convention as the Vulkan TqAttention shader)" — and compares the kernel output against the *rotated* needle value, so the test suite enshrines the wrong basis rather than catching it.

## Why this is a bug

The rotation `R = D·H` is orthonormal, so K-scores computed in the rotated domain are fine (`llm_tq_rotate_query` rotates q the same way). But the V aggregate must be brought back: the correct output is `R⁻¹·Σ w·norm·centroid + Σ w·v_fp32`. Omitting `R⁻¹` means the compressed-region part of the attention output is scrambled by a fixed rotation before hitting the output projection. This corrupts generation for **any model** once context exceeds the FP32 window (256), on CUDA and Vulkan alike, independently of the quantizer-distortion issue in #432.

The correct pattern already exists in-tree: the CPU path (`TurboQuantKvCache.ComputeVAggregation` Phase 3) accumulates in the rotated domain and applies the deferred per-head sign-flip + inverse WHT once; the KVarN CUDA kernel (`llm_kvarn_attention` Phase 3) does ONE inverse WHT per head in shared memory and then adds the FP32-window contribution in the original domain.

## Fix sketch

[...]

PR #436, opened 2026-07-11
fix(tq): default --tq-mode to auto — KVarN where supported, Lloyd-Max fallback with quality warning (#432)

### LINKED ISSUE #432: TurboQuant Lloyd-Max 3-bit collapses on Qwen3 (PPL 61x, degenerate decode) — pre-existing on master
## Summary

While running the KVarN P0 accuracy gate (#180, branch `feat/180-kvarn`), the existing Lloyd-Max 3-bit TurboQuant path (`--tq`, default mode) collapsed outright on Qwen3 models. Verified **pre-existing on master (f92d999)** via a temp-worktree A/B on the identical prompt — not a regression from the KVarN branch (whose Task-2 refactor of the Lloyd-Max path was also reviewed hunk-by-hunk as behavior-preserving).

## Evidence (Qwen3-0.6B-Q8_0, window 256, wikitext-2-raw, `perplexity` command from #180 Task 3)

| config | PPL (c=3072) | math micro-eval (20 problems, greedy, ~740-token prompts) |
|---|---|---|
| fp32 | 15.47 | 18/20 |
| `--tq` (Lloyd-Max 3-bit) | **945.6 (+6011%)** | **0/20 — 19/20 degenerate ("The The The…" loops to the 256-token cap)** |
| `--tq --tq-mode kvarn` | 15.67 (+1.26%) | 18/20, all coherent |

Position-bucket NLL shows the damage is entirely in the compressed region: [1,256) identical to fp32 (3.3533), [256,1024) 5.57 vs 2.26, [1024,3072) 7.77 vs 2.84.

Master A/B: `--tq` on master with the same P20 math prompt produces equally degenerate output ("1 1 1 1…"). Caveat: single-prompt spot check on 0.6B; the perplexity numbers are from the #180 branch harness.

## Notes

- Lloyd-Max ships hardcoded codebooks for head dims 128/256 (`TurboQuantCodebooks.cs`); Qwen3-0.6B/8B use head dim 128, so this is an in-support configuration.
- Worth bisecting whether this ever worked on Qwen3 (original TQ validation may have been on a different model family), and whether 4-bit (`bits: 4`) survives where 3-bit collapses.
- The `perplexity` subcommand + `scripts/kvarn-gate/` harness from #180 make this reproducible in ~2 min on a 0.6B model.
- If KVarN (#180) lands as the default TQ codec this becomes less urgent, but the current default `--tq` flag actively destroys generation quality on a mainstream model family and at minimum deserves a doc warning or a default change.

Found during #180 Task 3 (P0 accuracy gate), 2026-07-10.

https://claude.ai/code/session_018EWo1niax1g8A5E79RSALJ

pekkah:
## Problem

`--tq` (Lloyd-Max 3-bit, the default TurboQuant codec) collapses on Qwen3: wikitext-2 PPL **945.6** vs 15.47 fp32 at c=3072, and 19/20 math-eval generations degenerate into token loops (#432).

## Root cause

Three probes, all on real Qwen3-0.6B K/V vectors captured from the live forward pass (full analysis in [#432 comment](https://github.com/pekkah/SharpInference/issues/432#issuecomment-4946432935)):

1. **Not the Hadamard sign-flip ordering.** `TurboQuantOps` applies the sign flip after the WHT, which is mathematically a bare un-randomized WHT — but an A/B on real K/V shows identical reconstruction error for both orderings (relErr ≈ 0.18).
2. **Not a read-path bug.** The FastScan tile/LUT machinery tracks the per-block `DequantDot`/`Decompress` reference to ~1% of the quantizer's own error.
3. **Intrinsic 3-bit distortion × Qwen3's logit scale.** ~7% relative K-score error meets QK-norm attention logits with RMS ≈ 1000 → absolute logit errors of tens of units → softmax destroyed. 4-bit Lloyd-Max measures PPL **51.8** — better, still unusable. KVarN measures **15.67**.

## Fix

Make the validated KVarN codec the default wherever it is supported, per the issue's own recommendation:

- **CLI** (`run` + `perplexity`): `--tq-mode` defaults to **`auto`** → KVarN on CPU (any pow-2 head dim in [8, 1024]) and full-CUDA-offload dense (head dim ≤ 256); falls back to Lloyd-Max **with a loud quality warning citing #432** on Vulkan, partial CUDA offload, MoE-on-GPU, SnapKV, or window < 128. Explicit `--tq-mode lloydmax` keeps the old behavior and silences the warning; explicit `kvarn` still errors on unsupported combos.
- **Server**: new `TqMode` option (default `auto`) with the same per-path resolution in `InferenceEngineLoader` (explicit kvarn on an unsupported path fails model load); `SHARPI_TQ` / `SHARPI_TQ_MODE` env vars in the Host.
- **Docs**: flag descriptions, CLI README, and quality warnings on `TqQuantizer.LloydMax` / `ForwardPass.EnableTurboQuant` XML docs.

## Validation

- Default `--tq` on the issue's exact gate now runs `tq-kvarn-k4v2` at **PPL 15.6741** (was 945.56); explicit `--tq-mode lloydmax` reproduces the old numbers.
- `SHARPI_SNAPKV_BUDGET` + `--tq` prints the Lloyd-Max fallback warning and runs.
- `run --tq -g -1` on CUDA resolves to `KVarN K4V2`.
- Solution builds clean (warnings-as-errors); Tests.Cli 93/93, Tests.Server 154/154 (new tests cover the auto-resolution matrix).

## Follow-up

The GPU Lloyd-Max `TqAttention` V-basis defect found during this audit (compressed-region V aggregate never un-rotated on CUDA/Vulkan) is tracked separately as #435.

Closes #432

https://claude.ai/code/session_01EuBdnPNJ2XdaFp9PjYoJyz


pekkah:
## Code review (high effort, 8 finder angles + verify)

Ran an 8-angle multi-agent review (line-by-line, removed-behavior, cross-file, reuse, simplification, efficiency, altitude, conventions). One real regression this PR introduced, fixed in 80c1445; the rest are pre-existing or tech-debt, filed as follow-ups.

### Fixed (80c1445)

- **CONFIRMED regression — partial-offload crash.** A dense model with a pow-2 head dim outside {128,256} (e.g. 64) run with `--tq -g -1` on a GPU too small for full offload resolved auto→KVarN (passes the pow-2 gate), skipped the Lloyd-Max 128/256 gate, then the `wantHybrid` branch downgraded to Lloyd-Max **without re-checking** → `new CudaHybridForwardPass(turboQuant:true)` throws an uncaught `NotSupportedException` (`CudaHybridForwardPass.cs:386`) instead of the clean CLI error explicit `--tq-mode lloydmax` gives. Pre-PR this model errored cleanly at the 128/256 gate. Now guarded: the downgrade rejects non-{128,256} head dims with an actionable "use `-g 0` for CPU KVarN" message.
- **Cleanup**: dropped a dead `KVarN K4V2` message branch in the Vulkan CPU-fallback (KVarN can never reach the Vulkan backend — explicit is rejected, auto falls back to Lloyd-Max). Updated the stale TurboQuant description in `CLAUDE.md`.

### Follow-ups (not blocking — pre-existing or tech-debt)

- **Server head-dim gate** (`InferenceEngineLoader.cs:85`) rejects every head dim ∉{128,256} whenever TQ is on, *before* quantizer resolution — so KVarN's broader head-dim support (32/64/512/1024) is unreachable on the server, and explicit `TqMode=kvarn` on such a model gets a misleading "128 or 256" error. Not a regression (head-dim-64+TQ never worked on the server's Lloyd-Max-only path), but the CLI now runs it. Worth extending the server to match.
- **Triplicated support matrix.** The KVarN-support rules (SnapKV / head dim / backend / offload / MoE) and the #432 warning text are hand-copied across `RunCommand`, `PerplexityCommand.ResolveAutoQuantizer`, and `InferenceEngineLoader.ResolveTq` with duplicated reason strings. The crash above was a direct symptom of one copy drifting. A single Engine-level `TqSupport.ResolveAuto(...)` next to `TqQuantizer` (only the perplexity copy is unit-tested today) would prevent recurrence.
- **SnapKV + Lloyd-Max on CUDA full-offload** prints a "falling back to Lloyd-Max (SnapKV…)" warning before an unavoidable throw — the fallback reason is only valid on CPU (where Lloyd-Max+SnapKV works); on CUDA neither codec composes with SnapKV. Cosmetic; the actionable throw survives.
- **Server silently ignores `TqMode=kvarn` when `TurboQuant=false`** (CLI rejects the same combo). Minor.

I'll open a tracking issue for the matrix centralization + server head-dim extension.

